PR #12, opened 2019-02-07
Reactive Lernstore

smyth64:
If the list should be reactively updated, use get together with a computed property instead of list.

*Problem:*
The 'list' getter itself is not reactive.

Explanation
https://feathers-plus.github.io/v1/feathers-vuex/service-module.html#The-find-action

image

*Why is it like that?*
The reason for that is that the all the requests results are stored in the store so after that they can be separately fetched. 
*Example:*
If I search for "Mathe" the results are stored in the store. If I search for "Deutsch" the results are also stored. But the 'list' getter has now both results, Mathe and Deutsch.
If I search now for 'Mathe' again, we get immediate results because the results are already in the store.

As a solution there is the 'find' getter which makes a local query against the store. Unfortunately it doesn't seem to work with _all[$match].

*Solution*
Until now, this solution is working and doesn't generate more requests because the data is fetched from the store. We get the current results from pagination.ids and map the IDs to the individual results.

```
	computed: {
		...mapGetters("content_search", {
			getContent: 'get',
			fetchContent: 'find',
		}),
		...mapState('content_search', {
			pagination: state => {
				return state.pagination.content_list
			}
		}),	
		searchResults () {
			const { $store, getContent, pagination } = this

			if (pagination) {
				return pagination.ids.map(id => getContent(id))
			}

			return []
		},
       }
```

- Getting the list of results from the content_search getter
- Getting the pagination from the content_search state


pages/content/index.vue
```diff
@@ -61,35 +90,31 @@ export default {
...
-				query["_all[$match]"] = this.searchQuery;
-				query["$skip"] = this.skippedItems;
-			}
-			this.$store
+		async find(searchString) {
+			await this.$store
 				.dispatch("content_search/find", {
-					query: query,
+					query: this.query,
+					qid: 'content_list'
```
adrianjost: What us `qid: 'content_list'` doing?

adrianjost:
Nice work. 👍 I think 90% of this pull could be directly copied over into the documentation ;)

pages/content/index.vue
smyth64: The qid is there so we can make multiple find queries for different lists. As you can see we can access the fetches list with state.pagination.content_list

This is very useful because normally if you make a request without providing a qid it will just overwrite the old list.

PR #45, opened 2019-02-17
WIP - Docs update

adrianjost:
Just a test, if this get's deployed before being merged


adrianjost:
seems to work, I think we should disable this feature and simply build everything as soon as it hits the master branch

PR #55, opened 2019-02-22
Feathers-Vuex Update - Working with models. Make current user globally available

smyth64:
Adding Feathers-Vuex 1.7.0. Now we can use real Models instances and use CRUD operations and also custom operations.

**important**
Run 'yarn' so that feathers-vuex will be updated. Otherwise the models are not working.

Example is provided with the teams pages.

- Create empty team object
--> const team = new Team();

- Create team in the database
--> await team.create() 
--> await team.save() also uses create, if the team has no _id.

- Get list of teams
--> const teams = await Team.find();

- Get single team
--> const team = await Team.get(id);

- Save (Patch) single team
--> await team.save();

- Delete team
--> await team.remove()

In the store/index.js we can now also define default values for models which is very convenient. So we just create an empty model with e.g. "team = new Team()". And from there on we can work directly on the model.

Furthermore the current user is now available as this.$user. The reason is that the user object is needed very often so we don't have to get it from the store in every component.

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -63,6 +63,10 @@ module.exports = {
 			src: "@plugins/authenticate",
 			ssr: false,
 		},
+		{
+			src: "@plugins/auth",
+			ssr: false,
+		},
 		"@plugins/global",
 		"@plugins/directives",
 		"@plugins/theme",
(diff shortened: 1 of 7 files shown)


smyth64:
After discussing in the architecture meeting we're not gonna use the models/object oriented style.

PR #59, opened 2019-02-25
Adding fallback solution for themes

smyth64:
Can be tested with

```
SC_THEME=n21 npm run dev
```

And then open the url localhost:3000/theme/images/logo/logo.png
This image only exists in the default theme, not in n21.


adrianjost:
do we need to change anything on our deployment strategy for this to work? We currently deploy with `nuxt generate` (which equals `nuxt build` because we are running in spa mode) so as far as I understand Nuxt, the server isn't used at all?

adrianjost:
ping @smyth64

smyth64:
can be closed because we don't use express anymore

PR #63, opened 2019-03-04
Base Modal

smyth64:
Included:
- Modal
- UnitTests
- Storybook

image


src/components/ui/BaseButton.vue
```diff
@@ -26,5 +26,11 @@ export default {
 	&:focus {
 		background-color: transparentize($color-text, 0.1);
 	}
+	.is-light {
+		color: #363636;
+		background-color: #fff;
+		border-color: #dbdbdb;
```
adrianjost: @domi should we use variables here instead?

src/components/ui/BaseModal.unit.js
```diff
@@ -0,0 +1,48 @@
...
+			<BaseButton id="button" class="is-light" @click="showModal = false">
+				OK
+			</BaseButton>
+		</div>
+	</BaseModal>				
+	`,
+	components: { BaseModal },
+};
+
+describe("@components/BaseButton", () => {
```
adrianjost: `BaseButton` ?

src/components/ui/BaseModal.vue
```diff
@@ -0,0 +1,143 @@
...
+		<div v-if="showModal" class="modal-mask">
+			<div
+				class="modal-wrapper"
+				@mousedown.self="handleBackgroundClick"
+				@touchstart.self="handleBackgroundClick"
+			>
+				<div class="modal-container">
+					<div class="modal-header">
+						<slot name="header">
+							default header
```
adrianjost: I would use HTML-comments for the default values or remove them entirely so you don't need to specify something empty when you don't want to use this slot at all.

src/components/ui/BaseModal.vue
```diff
@@ -0,0 +1,143 @@
...
+		showModal: {
+			type: Boolean,
+			default: false,
+		},
+	},
+	methods: {
+		handleBackgroundClick() {
+			this.toggle();
+		},
+		toggle() {
```
adrianjost: If toggle is always emitting `close`, wouldn't it be better to name it `close`?

src/components/ui/BaseModal.vue
```diff
@@ -0,0 +1,143 @@
+<template>
+	<transition name="modal">
+		<div v-if="showModal" class="modal-mask">
+			<div
+				class="modal-wrapper"
+				@mousedown.self="handleBackgroundClick"
+				@touchstart.self="handleBackgroundClick"
+			>
+				<div class="modal-container">
+					<div class="modal-header">
```
adrianjost: Idea to make this component even more generic: 
Use a nested slot structure like this:
```
div.modal-container
  slot
    div.modal-header
      slot name=header
    div.modal-body
      slot name=body
    ...
```
Then you can use the component even for modals with completely customized layouts as a (like the name hints us) Base Component

src/components/ui/BaseButton.vue
dominikbruechner: Yes, lightColor, or if not generic buttonLightColor. ansonsten colorBlack für fff

src/components/ui/BaseModal.vue
smyth64: Good idea, gonna implement it.

src/components/ui/BaseModal.vue
smyth64: Gonna change it, also the name for showModal to -> active.

src/components/ui/BaseButton.vue
smyth64: ok i will set this variable into the default theme.

src/components/ui/BaseModal.vue
smyth64: alright

src/components/ui/BaseModal.vue
smyth64: unfortunately slot in slot is not possible, so we have to use just slot instead

adrianjost:
@smyth64 You don't need to import BaseComponents. They are always available. I changed that for you. (https://github.com/schul-cloud/nuxt-client/pull/63/commits/d8e5635e8d9b8616bddd56978a55ec53af67dfc6) 

Do you have any idea, why the BaseModal isn't working in Storybook? 
At `/news/_id/edit` it's working perfectly for me but using storybook I can't open it at all. Any ideas?

src/pages/news/_id/edit.vue
```diff
@@ -44,22 +77,22 @@ export default {
...
-					try {
-						await this.remove(this.news._id);
-						this.$toast.success("Artikel gelöscht");
-						this.$router.push({ name: "news" });
-					} catch (e) {
-						this.$toast.error("Fehler beim Löschen");
-					}
-				},
-			});
+			// this.$dialog.confirm({
```
adrianjost: this comment should be removed

src/pages/news/_id/edit.vue
```diff
@@ -1,38 +1,71 @@
...
+				<h1>{{ news.title }}</h1>
+				<!-- eslint-disable-next-line vue/no-v-html -->
+				<div v-html="news.content" />
+			</section>
+		</div>
+
+		<!-- use the modal component, pass in the prop -->
+		<BaseModal ref="modal" :active.sync="active">
+			<div class="modal-header">
+				<h3>custom header</h3>
```
adrianjost: 👆

src/pages/news/_id/edit.vue
smyth64: ja der kann raus

src/pages/news/_id/edit.vue
smyth64: if you talk about the comment, I removed it ;)

smyth64:
> @smyth64 You don't need to import BaseComponents. They are always available. I changed that for you. ([d8e5635](https://github.com/schul-cloud/nuxt-client/commit/d8e5635e8d9b8616bddd56978a55ec53af67dfc6))
> 
> Do you have any idea, why the BaseModal isn't working in Storybook?
> At `/news/_id/edit` it's working perfectly for me but using storybook I can't open it at all. Any ideas?

Very nice, thanks.
I was wondering before why we don't make these components globally available.

Perfect! 👍

adrianjost:
> @smyth64 You don't need to import BaseComponents. They are always available. I changed that for you. ([d8e5635](https://github.com/schul-cloud/nuxt-client/commit/d8e5635e8d9b8616bddd56978a55ec53af67dfc6))
> 
> Do you have any idea, why the BaseModal isn't working in Storybook?
> At `/news/_id/edit` it's working perfectly for me but using storybook I can't open it at all. Any ideas?

Storybook is now working

src/pages/news/_id/edit.vue
adrianjost: I talked about the title `custom header`. I changed that now

Issue #15, opened 2019-02-11
Base components TODOs


ChrisWeissmann:
Please add your name behind components you are developing and link the branch

# Following should be created as **base components** :
- [x] BaseLink
- [x] BaseInput
  - [x] type="date", ... (with DatePicker Component) [Domi]
  - [x] Switch
  - [x] hidden
  - [x] radio
- [x] BaseSelect [Domi]
  - [x] like normal html select
  - [x] tags/pills design? (Multiselect) [Domi]
- [x] BaseCard
- [x] BaseVideo [Angelika]
- [x] BaseAudio [Angelika]
- [x] BaseIcon
- [x] BaseModal

# Additionaly the following components should be also available:
- [x] Breadcrumb [Angelika]
- [x] Tabs [Chris]
- [x] Dropdown Menüs [Angelika]
- [x] Quote-Tag
- [x] Headline Classes
- [x] Collapsible
- [x] ProfileImage
- [ ] Filter (Feathers Filter Module) [Adrian]
- [ ] FAB (Floating Action Button) [ Adrian]
- [x] Tabelle
- [x] progress bar
- [x] Toasts (sind keine echte Komponente)

## Further todo's:
- [x] Theming (Michael)

dominikbruechner:
BaseInput type=date will be done by me

adrianjost:
Issues found:
- [ ] We should discuss, what components are considered `BaseComponents` and which are just normal Components. For example: I don't think `BaseProgressbar` isn't really a BaseComponent, cause I don't think it will be used that often. **We should create our own definition for what we count as BaseComponents, document this and move/rename our components accordingly**

PR #64, opened 2019-03-04
Component breadcrumb

AngelikaWieck:
# Issue #15 - BreadCrumb

## Description

Implemented the BreadCrumb-Component


src/components/ui/BaseBreadcrumb.vue
```diff
@@ -0,0 +1,63 @@
...
+		removeText({ text, ...input }) {
+			return input;
+		},
+	},
+};
+</script>
+
+<style lang="scss" scoped>
+@import "@variables";
+ul.breadcrumb {
```
adrianjost: For performance reasons you should avoid tag selectors (`ul`) whenever possible and use classes or ID's instead. But as our devices are already reasonable fast, it's not that big of a deal.
https://csswizardry.com/2011/09/writing-efficient-css-selectors/

src/components/ui/BaseBreadcrumb.vue
```diff
@@ -0,0 +1,63 @@
...
+	padding: 10px 16px;
+	list-style: none;
+	background-color: #eee;
+}
+
+/* Display list items side by side */
+ul.breadcrumb li {
```
adrianjost: For readability, I would use sass nesting here, just because it's possible. You would have saved some keystrokes as well ;D

adrianjost:
I will investigate why the tests are broken tomorrow

PR #66, opened 2019-03-05
Component/dropdown menu

AngelikaWieck:
# Issue #15 - Dropdown Menüs 

## Description

implemented the DropdownMenu-Component

Im Storybook zeige ich die Komponente unter den "Base Components" an, weil ich nicht wusste, was ein besserer Ort dafür ist (die DropdownMenu-Komponente ist ja eigtl. keine "Base Component"). Also müsste das noch gefixed werden, wenn es einen besseren Ort gibt :) 

## Screenshot
image


src/components/DropdownMenu.vue
```diff
@@ -0,0 +1,72 @@
+<template>
+	<div class="dropdown">
+		<button class="button">{{ title }}</button>
```
adrianjost: Verwende bitte `BaseButton` statt Button und wrappe diesen in einen `slot`, damit man das Dropdown auch mit anderen Elementen verwenden kann.

adrianjost:
I found some inspiration: “Paying Attention to Details” by Shlok Desai https://medium.com/p/eb0c43d0b02b

src/components/uidMixin.js
```diff
@@ -0,0 +1,9 @@
+let uid = 0;
```
adrianjost: @smyth64 I like the idea of using mixins, but I don't think they belong into the `components` folder. WIth your Nuxt experience, what place do you recommend?

Normally I would create a seperated `src/mixins`-folder for such stuff to keep the components folder clean and consistency filles with vue and unit test files only.

adrianjost: @AngelikaWieck After merging this pull request, could you implement your elegant solution [here](https://github.com/schul-cloud/nuxt-client/blob/master/src/components/ui/BaseInput.vue#L14) as well so we can remove the prop `id` there and be more consistent?

AngelikaWieck: @adrianjost It is even easier:
```
<input id="foo" />
<label for="foo">Bar</label>
```

...can be written as:
```
<label>
  <input />
  Bar
</label>
```

Maybe `Bar` could be wrapped in a span for styling.

So, no `id` is needed  🎉

adrianjost: @AngelikaWieck You may be right, but @dominikbruechner encountered some problems using <label> to wrap the `FlatPickr`-Component ([Link](https://github.com/schul-cloud/nuxt-client/blob/master/src/components/ui/BaseInput.vue#L43)). We should check this again. I will create some Issues for that

dominikbruechner:
Could you please add screenshots in specific UI relevant PRs? cheers :)

stories/misc.stories.js
```diff
@@ -25,9 +25,9 @@ storiesOf("Misc", module)
 		components: { DropdownMenu, MenuLink },
 		template: `
 <DropdownMenu title="Dropdown">
-	<MenuLink href="#" class="test">Link 101</MenuLink>
-	<MenuLink href="#">Link 2</MenuLink>
-	<MenuLink href="#">Link 3</MenuLink>
+	<MenuLink href="/">Link 1</MenuLink>
```
adrianjost: Check your browser console warnings. I think you need to use `to` instead of `href` for internal links.

PR #67, opened 2019-03-06
Update Broken Documentation Links

adrianjost:
quickfix for broken docs links


AngelikaWieck:
According to the [VuePress documentation](https://vuepress.vuejs.org/guide/markdown.html#links) links can also end with `.md`. If all links are changed to `.md` they would also work in GitHub.

adrianjost:
Good point, I will change this :)

PR #71, opened 2019-03-06
#42 deploy cleanup

adrianjost:
# Issue #42 

## Description

- moved deploy related files into `/deploy` files
- Dockerfiles aren't building the projects a second time (huge performance boost!). Instead, they copy the build files generated during the travis run. ~10min -> ~1min for deployment 😎 

**Update**
- the fast dockerfile is now optional, so in case of error or if you wan't to run the build in an enviroment without node.js you can use the normal Dockerfile as before. For Travis, the fast build mode can be enabled by setting the `DOCKERFILE_VERSION` env variable to `Dockerfile.fast` which acts as an prefix for the Dockerfile name used.


adrianjost:
Update: Anmerkungen von @FalkoBenthin sind nun eingeflossen. Die bisherigen Dockerdateien existieren weiterhin, damit man auch auf platformen ohne Node.JS mit Docker einen Build ausführen kann. Der Switch erfolgt über die Umgebungsvariable `DOCKERFILE_VERSION`

PR #80, opened 2019-03-11
#24 add codecov tool

adrianjost:
# Issue #24 

## Description

- Dashboard: https://codecov.io/gh/schul-cloud/nuxt-client


janrenz:
lgtm. we need to discuss how (and if) we want to integrate the results from the integration tests. but this will need some more time.

PR #92, opened 2019-03-14
Adding Dialog for simplified modals

smyth64:
For many modals we only need a text + confirmation/cancel buttons. This plugin makes it easy available with this.$dialog.confirm()


src/components/ui/BaseDialog/BaseDialog.vue
```diff
@@ -0,0 +1,133 @@
+<template>
+	<BaseModal :active.sync="isActive">
+		<div class="modal-header">
+			<h3>{{ title }}</h3>
+		</div>
+
+		<div class="modal-body">
+			<!-- eslint-disable-next-line vue/no-v-html -->
+			<p v-html="message" />
```
adrianjost: Why are you using v-html here? You said the use case of this is a simple confirmation message, so why allow html here and may create a xss security vulnerability?
```suggestion
			<p>{{ message }}</p>
```

smyth64:
> Could you add a unit test for `this.$dialog.confirm(...)` and so on or is this way to complicated?
> 
> Otherwise, I like the idea 👍

EDIT: Found a solution. I use the returned dialog component and test from there! Works nicely :)

Hi Adrian,
I tried this desperately but I failed at this point.

What I did was this:
```
	it("has a property $dialog with the function confirm", () => {
		const localVue = createLocalVue();
		localVue.use(BaseDialogPlugin);
		expect(typeof localVue.prototype.$dialog.confirm).toBe("function");

		const wrapper = mount(BaseDialog, {
			attachToDocument: true,
                        localVue
		});

                wrapper.vm.$dialog.confirm({
					title: "Deleting account",
					message:
						"Sure??",
					confirmText: "Delete Account",
					type: "is-danger",
					onConfirm: () => this.$toast.success("Account deleted!"),
				});
	});
```

The problem is, that the dialog gets attached to the root DOM and I didn't find a way to access the root DOM =/ There seems to be a a way with wrapper.vm.$root.$el but I cannot find it inside there. Another way is to set a ref to the dialog.

src/components/ui/BaseDialog/BaseDialog.vue
smyth64: The reason is, that sometimes we want to style the message a bit for example with some <b>bold words</b> in it. I think there is no xss security issue because the message is passed only inside the code and not from an input field etc. Also it's possible to sanitize the html. there is a package for that: sanitize-html
We could use this for the variables whenever we need to use v-html.

src/components/ui/BaseDialog/BaseDialog.vue
```diff
@@ -103,6 +106,7 @@ export default {
 	},
 	mounted() {
 		this.isActive = true;
+		// this.customRef = 'asdfdsfs'
```
adrianjost: Should be removed

src/components/ui/BaseDialog/BaseDialog.vue
adrianjost: Most of the time, that's correct. But think about something like `do you wan't to delete task "${task.title}"`. This is a common way to introduce user generated content into the dialog which could be mailcious. Ideally this should be filtered on the server side, but I would prefer the safe way. The package you recommend sound's good.

src/components/ui/BaseDialog/BaseDialog.vue
```diff
@@ -0,0 +1,135 @@
...
+		},
+		focusOn: {
+			type: String,
+			default: "confirm",
+		},
+	},
+	data() {
+		return {
+			isActive: false,
+			validationMessage: "",
```
adrianjost: unused variable?

src/components/ui/BaseDialog/index.js
```diff
@@ -0,0 +1,30 @@
...
+		el: document.createElement("div"),
+		propsData,
+	});
+}
+
+const DialogProgrammatic = {
+	confirm(params) {
+		const defaultParam = {};
+		const propsData = Object.assign(defaultParam, params);
+		return open(propsData);
```
rodebert: If `defaultParam` is an empty object you do not need to this and can just return `open(params)`.

src/components/ui/BaseDialog/BaseDialog.vue
```diff
@@ -0,0 +1,135 @@
...
+				case "is-success":
+					return "check-circle";
+				case "is-warning":
+					return "alert";
+				case "is-danger":
+					return "alert-circle";
+				default:
+					return null;
+			}
+		},
```
rodebert: Shorter:

```
return {
    "is-info": "information",
    "is-success": "check-circle",
    "is-warning": "alert",
    "is-danger": "alert-circle"
}[this.type];
```

src/components/ui/BaseDialog/BaseDialog.vue
smyth64: Yep, removed!

src/components/ui/BaseDialog/BaseDialog.vue
smyth64: Thx, this is really cool! 👍👍👍 Implemented

src/components/ui/BaseDialog/index.js
smyth64: This was just a way to ensured the params are not undefined.
Doesn't matter anyway because they can also be undefined. Implemented.

src/components/ui/BaseDialog/BaseDialog.vue
smyth64: Done

src/components/ui/BaseDialog/BaseDialog.vue
smyth64: Hmm yeah that could be possible. Let's not use the v-html then for now.

If the content is too complex, we could pass a pseudo-component into the content section, which can have HTML and also logic. I'll make another pull request for that.

smyth64:
@rodebert 
The idea of you using the Vue Bus for the dialog (And also for the modal then) is very nice and I like to implement it.
What do you think about the idea to merge this dialog into the master now and I make a second pull request including the version with the Vue Bus.

I really need the feature right now because I'm working on the teams page and there are some confirm dialogs.

@adrianjost 
I wrote a comment regarding the v-html issue.

adrianjost:
After resolving conflicts, this can be merged

PR #99, opened 2019-03-19
Component/base video

AngelikaWieck:
# Issue #15 

## Description

- Created BaseVideo Component 

- Hier ein komischer Fehler, der bei mir aufgetreten ist:  

Bildschirmfoto 2019-03-19 um 13 56 47

## Todo
- [ ] Add property-descriptions (currently not possible)


adrianjost:
ich würde vermuten, der fehler kommt daher, dass 
a) der videoplayer auf methoden zugreift, welche es zwar im Browser gibt, im Node.JS Test-Runner nicht gemockt sind
b) er aus irgend einem grund die babel config nicht richtig lädt. Jest kennt "import" nicht, wesshalb wir es mit babel zu require() transformieren. Kann natürlich sein, dass babel nur `import a from "b"` kennt, aber das reine `import "b"` ihm fremd ist.

src/components/ui/BaseVideo.vue
```diff
@@ -0,0 +1,96 @@
...
+
+export default {
+	props: {
+		streams: {
+			type: Array,
+			required: true,
+		},
+		fallbackStream: {
+			type: Object,
+			required: false,
```
adrianjost: If the prop isn't required, you don't need to specify `required: false`. It is false by default. Instead you need to provide a default value (`default: "hi"`).
For `type: Object || Array` this default value must be a function (`default: ()=>({ txt: "hi"})`).

src/components/ui/BaseVideo.vue
```diff
@@ -0,0 +1,96 @@
...
+			type: String,
+		},
+		videoPreload: {
+			type: Boolean,
+		},
+		trimVideo: {
+			type: Object,
+		},
+		chapters: {
+			type: Array,
```
adrianjost: Couldn't you use `v-bind="$attrs"` or a simple `configuration`-prop instead of all these 💯 confusing props? Or have you wrote them done to be able to document them? If so: good idea but I think a simple link to the documentaion would be easier to use.

src/components/ui/video-player-helper.js
```diff
@@ -0,0 +1,3 @@
+import Hls from 'hls.js/dist/hls.light.min';
+
+window.Hls = Hls
```
adrianjost: Why was this moved into a separate file? Couldn't it be done in the component itself?

stories/base.stories.js
```diff
@@ -194,4 +195,13 @@ storiesOf("Base Components", module)
 			</div>
 		`,
 		methods: {},
+	}))
+	.add("Base Video", () => ({
+		components: { BaseVideo },
+		template: `
+		<BaseVideo 
+			:streams="[{hd: 'http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4'}]"
```
adrianjost: @janrenz is this external Google url as an storybook example okay or do you have privacy concerns?

src/components/ui/BaseVideo.vue
adrianjost: That's also the reason why Travis fails

src/components/ui/BaseVideo.vue
AngelikaWieck: I've written them down to document them, but the props description doesn't work.
I also thought about just writing down the link to the documentation but this neither worked, because I wasn't able to figure out how to put normal infotext, or even a link in the storybook (commenting à la jsdoc doesn't work).

src/components/ui/video-player-helper.js
AngelikaWieck: This has to be in a separate file, because Hls has to be exported globally for the video-player to work. If these lines were in the component itself, we can't ensure that first of all Hls is exported globally and secondly the video-player gets imported.

PR #105, opened 2019-03-20
Component/Quote

AngelikaWieck:
# Issue #15 

## Description

- Quote

## Screenshot

Bildschirmfoto 2019-03-21 um 06 41 14


janrenz:
Screenshot please

janrenz:
thx :)

src/components/Quote.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+	<blockquote :cite="cite">
+		<slot />
+	</blockquote>
+</template>
+
+<script>
+export default {
+	name: "Quote",
```
adrianjost: Component Names should always be multi-word. 
https://vuejs.org/v2/style-guide/#Multi-word-component-names-essential

We still don't have a clear definition where to place our component but if we use a Base-Component definition like 

Base Components are Components 
 - without dependencies to other components
- that only react to state changes with events
- have no external requests
- get all there information via props only
- have no specific text like `Courses` included

Then this would count as a Base-Component and the naming problem would be solved.

src/components/Quote.vue
```diff
@@ -0,0 +1,28 @@
...
+	props: {
+		cite: {
+			type: String,
```
rodebert: The `cite` attribute needs to be a valid URL so it should be validated as such and it is not required by the standard.

rodebert:
As the `cite`-Attribute is not provided to the user by browser this component needs to display the source like so:

```
<blockquote cite="http://www.worldwildlife.org/who/index.html">
    For 50 years, WWF has been protecting the future of nature. The world's
    leading conservation organization, WWF works in 100 countries and is
    supported by 1.2 million members in the United States and close to 5 million
    globally.
    <footer>
        — <cite><a href="http://www.worldwildlife.org/who/index.html">Source</a></cite>
    </footer>
</blockquote>
```

Maybe it makes sense to make the Text inside the Link to the citation source configurable.

src/components/ui/BaseBlockquote.unit.js
```diff
@@ -0,0 +1,7 @@
+import BaseBlockquote from "./BaseBlockquote";
+
+describe("@components/ui/BaseBlockquote", () => {
+	it("exports a valid component", () => {
+		expect(BaseBlockquote).toBeAComponent();
+	});
```
adrianjost: We have new helper methods to simplify some common tests. Could you please replace your basic test with those two ones:
```js
it(...isValidComponent(BaseBlockquote));	it(...rendersDefaultSlotContent(BaseBlockquote));
```

adrianjost: You should also check if the srcText gets rendered and if there is a link when the cite attribute is given

PR #111, opened 2019-03-22
unified ports

adrianjost:
## Description
- unified port numbers to make it easier to remember
  - **nuxt:** 4000
  - **storybook:** 4001
  - **docs:** 4002


adrianjost:
@ChrisWeissmann I like your idea. I will create an issue for that, as it can be tricky to implement this output cause all those services have a fallback implemented if the given Port is already in use. Therefore I can't hardcode those port numbers in a log message cause they may change during startup. I think it's also more an parallel output problem rather than something we should add. So it's more a bug than a feature.

Issue #129, opened 2019-03-27
add postCss for autoprefix plugin


adrianjost:
👆

adrianjost:
I wasn't able to figur out how to do that, so I need some help here @rodebert / @ChrisWeissmann .

ChrisWeissmann:
I looked into it a bit. What would the perfect test case be to check if it works?

adrianjost:
I would say add something that needs to be prefixed, like `@keyframe`, to the css anywhere. Then run `yarn build:nuxt` and search in the `dist/nuxt` folder (vscode can do that) for the prefixed version (like `@-webkit-keyframe` for keyframes)

ChrisWeissmann:
Followed instructions on here: [nuxtjs.org](https://nuxtjs.org/faq/postcss-plugins/)
I think we can delete the .postcssrs.js file or does it serve a purpose?

PR #143, opened 2019-04-04
added postcss autoprefixer

ChrisWeissmann:
# Issue #129 

## Description

- added autoprefixer in nuxt config for cross browser compatibility


adrianjost:
And what is the status of the `.postcssrs.js` file? Is it still needed? I personally don't know.

Issue #109, opened 2019-03-22
Remove Google Fonts from Storybook


adrianjost:
Storybook is currently loading fonts from fonts.google.com. This should be replaced with a local copy to remove all unnecessary external calls.

ChrisWeissmann:
Not sure what best practice is for this. 
Using the package https://www.npmjs.com/package/google-fonts-webpack-plugin 
Or import them to the project as a ttf file?

adrianjost:
I would go with downloaded, local versions. We do this in all other repositories as well.
There shouldn't be any need for updates of the font in the furture so version management like npm is not an huge improvement.

adrianjost:
done

PR #144, opened 2019-04-04
Added fonts locally

ChrisWeissmann:
# Issue #109 

## Description

Removed google fonts import and added them in the assets folder. Not sure this is the corret way.
Appriate any comment on if this is correct.


src/themes/default/styles/variables/_fonts.scss
```diff
@@ -1,19 +1,155 @@
-@import url("https://fonts.googleapis.com/css?family=PT-Sans|PT+Sans+Caption|PT+Sans+Narrow");
+/* latin-ext */
+@font-face {
+	font-family: "PT Sans";
+	font-style: normal;
+	font-weight: 400;
```
adrianjost: We could define the font-swap property to improve the rendering and give us more control: e.g. `font-display: swap;`

adrianjost: Da wir hier keine präferenz haben, ist auto auch okay.

adrianjost:
@ChrisWeissmann you can merge, if you agree with my opinion 😅

ChrisWeissmann:
I had contact with the maintainers in Discord and from what I understand the problem is, is that the preview-head.html (what is loaded before the stories) is noth parsed through webpack and theirfore the files aren't loaded. I don't know a solution for this :/ I would also merge so that the nuxt client fonts are at least local

Issue #86, opened 2019-03-11
Add tests to all BaseComponents


dominikbruechner:

adrianjost:
Almost all BaseComponents are now tested.
Only `BaseIcon.vue` is missing, cause testing was pretty hard (impossible for us so far) and we will redo the implementation with #130 anyway.

PR #152, opened 2019-04-12
#86 add tests to all base components

adrianjost:
# Issue #86 

## Description

- Just added some tests and fixed a small modal bug


src/components/ui/BaseModal.unit.js
```diff
@@ -57,7 +57,21 @@ describe("@components/BaseModal", () => {
 		const wrapper = mount(modal);
 		wrapper.vm.active = true;
 		expect(wrapper.find("#button").exists()).toBe(true);
-		wrapper.find(".modal-wrapper").trigger("mousedown");
+		wrapper.find(".modal-wrapper").trigger("click");
 		expect(wrapper.find("#button").exists()).toBe(false);
 	});
+
+	it("closed modal can be reopened after clicking outside", (done) => {
+		// TODO this test may can be removed, the problematic code is removed now.
```
ChrisWeissmann: There is a todo that seems relevant here. Can this part of the test be removed?

src/components/ui/BaseModal.unit.js
```diff
@@ -57,7 +57,21 @@ describe("@components/BaseModal", () => {
...
+		const wrapper = mount(modal);
+		wrapper.vm.active = true;
+		expect(wrapper.find("#button").exists()).toBe(true);
+		wrapper.find(".modal-wrapper").trigger("click");
+		expect(wrapper.find("#button").exists()).toBe(false);
+		setTimeout(() => {
```
ChrisWeissmann: Does this need to be done?
[Docs](https://vue-test-utils.vuejs.org/guides/#what-about-nexttick) say that all updates are applied synchronously

src/components/ui/BaseModal.unit.js
adrianjost: The test covered this specific code: https://github.com/schul-cloud/nuxt-client/pull/152/files#diff-6279f0ec653ce805d7c1c07287b7c7f1L32
I removed the code, but I am unsure wheather we should keep the test or not to prevent the error in the future.

After rethink, I removed the test entirely, cause the bug was way to specific and woudn't be likely to be introduced again.

(@smyth64 just to keep you informed that I removed your code there and why)

src/components/ui/BaseModal.unit.js
adrianjost: This was needed, because the original code used an timeout and I wanted to reproduce the bug. I now removed the test entirely.

PR #159, opened 2019-04-14
#104 refactor style variables

adrianjost:
# Issue #104

based on #156 

## Description

- add a universal spacing unit


adrianjost:
Will continue later on this

PR #168, opened 2019-04-26
Add tests for all regular components

ChrisWeissmann:
## Description

- Added tests for all regular components to start a good habit of always having a test
- Test coverage is now at around 90%


src/components/CourseCard.unit.js
```diff
@@ -0,0 +1,76 @@
+import CourseCard from "./CourseCard";
+
+describe("@components/CourseCard", () => {
+	it(...isValidComponent(CourseCard));
+
+	it("check for footer and tab", () => {
```
adrianjost: Check for what? Check correct rendering, functionality? ... 
Please update the test description.

src/components/Courses.unit.js
```diff
@@ -0,0 +1,11 @@
+import Courses from "./Courses";
+
+describe("@components/Courses", () => {
+	it(...isValidComponent(Courses));
+
+	it("Test default/empty", () => {
```
adrianjost: Shouldn't we remove thid dummy test until we have a real one?

src/components/DropdownMenu.unit.js
```diff
@@ -1,7 +1,35 @@
...
+	it("Check if everything is rendered", () => {
+		const titleString = "Test Dropdown";
+		const wrapper = shallowMount(DropdownMenu, {
+			propsData: {
+				title: titleString,
+			},
+		});
+		expect(wrapper.find(".dropdown").isVisible()).toBe(true);
+		expect(wrapper.find(".button").isVisible()).toBe(true);
+		expect(wrapper.find(".button").text()).toBe(titleString);
```
adrianjost: This test is pretty much useless I think. It just tests the exact implementation of the DropdownMenu. To make it an proper test, it should also check, that everything that shoudn't be rendered isn't.

src/components/LandingCTA.unit.js
```diff
@@ -0,0 +1,27 @@
+import LandingCTA from "./LandingCTA";
+
+describe("@components/LandingCTA", () => {
+	it(...isValidComponent(LandingCTA));
+
+	it("Render with defaults", () => {
```
adrianjost: Whats the intention of this test? I don't even understand, why a LandingCTA should even have a default text. I mean, what#s the usecase for that?

src/components/LandingCTA.unit.js
```diff
@@ -0,0 +1,27 @@
...
+		const wrapper = shallowMount(LandingCTA);
+		expect(wrapper.find("img").exists()).toBe(true);
+		expect(wrapper.find("base-button-stub").exists()).toBe(true);
+	});
+
+	it("Render with some text", () => {
```
adrianjost:
```suggestion
	it("Renders Text Props", () => {
```

src/components/LandingCTA.vue
```diff
@@ -22,9 +22,10 @@ export default {
 		return {};
 	},
 	methods: {
-		onClick() {
-			this.$emit("click");
-		},
+		//TODO can this be removed? I don't think it does anything
```
adrianjost: :+1:

src/components/Pagination.unit.js
```diff
@@ -0,0 +1,55 @@
+import Pagination from "./Pagination";
+
+describe("@components/Pagination", () => {
+	it(...isValidComponent(Pagination));
+
+	it("Check that no previous page link is being rendered on first page", () => {
```
adrianjost: These are some good unit tests, that actually test some functionality! ❤️

src/components/Pagination.vue
```diff
@@ -50,14 +50,16 @@ export default {
 		},
 	},
 	computed: {
+		// Shouldn't this be rounded
```
adrianjost: 👍

src/components/Pagination.vue
```diff
@@ -50,14 +50,16 @@ export default {
...
 		currentPage() {
-			return this.state.skip / this.state.limit + 1;
+			return Math.floor(this.state.skip / this.state.limit + 1);
 		},
 		lastPage() {
 			return Math.ceil(this.state.total / this.state.limit);
 		},
 	},
 	methods: {
+		//TODO is gotoPage needed? not called anywhere
```
adrianjost: Can be removed I guess. @dominikbruechner

src/components/ProfilePicture.unit.js
```diff
@@ -0,0 +1,24 @@
...
+	});
+
+	it("Render with a src and specified size", () => {
+		const wrapper = shallowMount(ProfilePicture, {
+			propsData: {
+				image: "test",
+				size: "large",
+			},
+		});
+		expect(wrapper.find("img").classes("large")).toBe(true);
```
adrianjost: What if I remove all the styles from the component? The tests would still run fine and nobody would notice. => The test is pretty much useless in this form. A better test would be, to compare multiple renderings and compare the rendered sizes (e.g. using [this method](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect)).

src/components/PulsatingDot.unit.js
```diff
@@ -0,0 +1,11 @@
+import PulsatingDot from "./PulsatingDot";
+
+describe("@components/PulsatingDot", () => {
+	it(...isValidComponent(PulsatingDot));
+
+	it("render default", () => {
+		const wrapper = shallowMount(PulsatingDot);
+		expect(wrapper.is("div")).toBe(true);
+		expect(wrapper.find(".ring-container").exists()).toBe(true);
```
adrianjost: This test is also pretty much testing nothing. Checking if a div exists is 100% checking implementation details what should be avoided. A better test would be, to test if an animation is running on the ring-container.
(maybe by watching some events like `animationstart`?)

src/components/Searchbar.vue
```diff
@@ -15,6 +15,7 @@
 				@blur="isActive = false"
 			/>
 			<transition name="fade">
+				<!-- shouldn't this be baseButton -->
```
adrianjost: I'm unsure about this. Yes it is a button, but 100% of the styles are different. If we wan't to use the baseButton here, we need to add a "nostyle" option to it, so we can apply such custom styles like here, without needing multiple css overrides.

src/components/TheTopBar.unit.js
```diff
@@ -0,0 +1,44 @@
...
+	};
+	it("Render defaults", () => {
+		const wrapper = shallowMount(TheTopBar, {
+			mocks: {
+				$theme,
+			},
+		});
+		expect(wrapper.find("img").exists()).toBe(true);
+		expect(wrapper.find(".page-title").text()).toBe("HPI Schul-Cloud");
+		expect(wrapper.find(".action").exists()).toBe(false);
```
adrianjost: What is this test testing? Looks like it's testing implementation rather than behaviour.
A better test would be, to check if there is a logo and the theme title rendered.

src/components/TheTopBar.unit.js
```diff
@@ -0,0 +1,44 @@
...
+	});
+
+	it("Render with links and buttons", () => {
+		//Issue with the ripple effect => Failed to resolve directive: ripple
+		const mockActions = [
+			{ icon: "house", title: "test home", to: "home" },
+			{ title: "test away", href: "https://www.google.com" },
```
adrianjost:
```suggestion
			{ title: "test away", href: "https://schul-cloud.org" },
```

src/components/TheTopBar.vue
```diff
@@ -54,6 +54,7 @@ export default {
 	},
 	methods: {
 		sendEvent(eventName) {
+			// Is this if needed? Currently I don't think the sendEvent can be triggered without an eventName
```
adrianjost: 👍

src/components/Courses.unit.js
ChrisWeissmann: Removed it.

src/components/CourseCard.unit.js
ChrisWeissmann: It was basically checking the implementation. I removed it and added one that checks if what is not supposed to be rendered is really not showing.

src/components/LandingCTA.unit.js
ChrisWeissmann: Then we should change the implementation to make the strings required instead of having a defualt.

src/components/LandingCTA.unit.js
ChrisWeissmann: 👍

src/components/TheTopBar.vue
ChrisWeissmann: Removed

src/components/DropdownMenu.unit.js
ChrisWeissmann: That's already done in the test above this one.

src/components/ProfilePicture.unit.js
ChrisWeissmann: Should there also maybe be a validator for the size name?

src/components/Searchbar.vue
ChrisWeissmann: hm yea I see. I don't know what the best practice would be. But if we are still considering implementing a linter rule to not allow the button html tag then I guess a no style option is the way to go.

src/components/TheTopBar.unit.js
ChrisWeissmann: it also tests that no actions are being rendered as per default their aren't any.

src/components/PulsatingDot.unit.js
ChrisWeissmann: I do agree that the test basically tests implementation. Not sure I agree that we should test the animation.

adrianjost: The question is: What do you expect when using the `PulsatingDot`-Component? Do you expect to get a `div`-tag with the class `ring-container` or do you expect that something is round and animated? The test should cover what we expect when using it

src/components/Searchbar.vue
adrianjost: I totally agree with you

src/components/TheTopBar.unit.js
adrianjost: Yeah but the test fails as soon as we change the img tag to a picture tag or change the theme. The value of the image test is also pretty low, cause it's not specififed what img we expect there. Maybe we could read the theme file and compare, that the title is showing, what is specified in the theme file, including the correct image url?

src/components/Searchbar.vue
ChrisWeissmann: I will remove my comment and add a ticket in Jira for this.

src/components/Pagination.vue
ChrisWeissmann: I talked with Dominik. He doesn't need it.

src/components/TheTopBar.unit.js
ChrisWeissmann: From what I understand we should mock the theme variable. So I am not sure if mocking an image path and then checking if its the correct image is that good of a test. 
I would rather remove the image line and leave the rest which checks the defaults are rendering.

PR #170, opened 2019-05-02
reorganized courses grid, added tests

dominikbruechner:
code refactoring + tst


src/components/CoursesGrid.unit.js
```diff
@@ -0,0 +1,29 @@
+import { shallowMount } from "@vue/test-utils";
```
adrianjost: not needed, shallowMount is imported globaly
```suggestion
```

src/components/CoursesGrid.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+	<div class="app">
```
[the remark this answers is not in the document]
adrianjost: Everything else looks good.

src/components/CoursesGrid.vue
```diff
@@ -0,0 +1,46 @@
...
+			</div>
+		</div>
+	</div>
+</template>
+
+<script>
+import CourseCard from "./CourseCard.vue";
+
+export default {
+	name: "CoursesGrid",
```
adrianjost: We don't specify the name anymore. It's autogenerated from the filename and is only needed when defining recursive components.

src/components/CoursesGrid.vue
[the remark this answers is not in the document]
dominikbruechner: oh I was just copying code, and missed it....

src/components/CoursesGrid.vue
dominikbruechner: kk

PR #179, opened 2019-05-10
Codecov block on coverage drop

adrianjost:
based on #178 because we had some broken tests.

## Description

To prevent coverage dropping in the future, codecov will block any pr where the coverage will be decreased by more than 1%.


dominikbruechner:
Gibts einen Story-book UI screenshot oder hat sich nix geändert?

PR #183, opened 2019-05-12
#160 Update Nuxt

adrianjost:
# Issue #160 

~Again, it is based on #178 because `yarn install` was not working with the current master of the video player.~

## Description

- The issue was, that we had overwritten the nuxt aliases. I changed our `~` to `@@` to make it conform with the existing nuxt aliases and switched from overwriting to extending the nuxt config.


aliases.config.js
```diff
@@ -4,8 +4,8 @@ const prettier = require("prettier");
 const themeName = process.env.SC_THEME || "default";
 
 const aliases = {
-	"~": ".",
```
adrianjost: nuxt defines:
`~` : `src`
`~~` : `.`
`@`: `src`
`@@`: `.`

I kept our `@` aliases, even if they are defined redundant to the existing nuxt aliases, so we can look up faster what is the alias `@` is resolving to.

nuxt.config.js
```diff
@@ -96,7 +96,8 @@ module.exports = {
 		 ** You can extend webpack config here
 		 */
 		extend(config, ctx) {
-			config.resolve.alias = require("./aliases.config").webpack;
+			Object.assign(config.resolve.alias, require("./aliases.config").webpack);
```
adrianjost: That's the real change.

Issue #173, opened 2019-05-04
use DayJS instead of Moment.JS


adrianjost:
It's **a lot** smaller!
https://github.com/iamkun/dayjs

ChrisWeissmann:
Looks good!
Supports I18n!
Here is the doc about how to add the support: https://github.com/iamkun/dayjs/blob/dev/docs/en/I18n.md

ChrisWeissmann:
@adrianjost  can this be closed then? seems like you mentioned this in the PR but I can't find the reference.

adrianjost:
No, it isn't implemented yet. We are still using mobile. This ticket was created as a future todo

PR #187, opened 2019-05-17
use dayjs instead of moment

adrianjost:
# Issue #173 

## Description

- replaced moment.js with day.js. It wasn't 100% compatible but close enough. The on missing function was pretty easy to code in plain js.
- fixed some browser console errors & removed unused code
- added some tests to pass the codecov threshold


src/components/ui/BaseSelect.vue
```diff
@@ -60,7 +60,7 @@ export default {
 		},
 		optionLabel: {
 			type: String,
-			default: "",
+			default: "label",
```
adrianjost: add default label that we forgot in #139

src/pages/courses/create.vue
```diff
@@ -86,10 +83,10 @@ export default {
...
 
 			course.times = this.course.times.map((time) => {
-				time.startTime = moment
-					.duration(time.startTime, "HH:mm")
-					.asMilliseconds()
-					.toString();
+				let [startHours, startMinutes] = time.startTime.split(":");
+				startMinutes = startMinutes * 60 * 1000;
+				startHours = startHours * 60 * 60 * 1000;
+				time.startTime = startHours + startMinutes;
```
adrianjost: `.duration` and `.asMilliseconds` is not supported by dayjs 😢 But I think the plain JS version should be fine here.

src/pages/teams/_id/members.vue
```diff
@@ -397,7 +396,7 @@ export default {
 	computed: {
 		invitedMembers() {
 			return this.team.invitedUserIds.map((invite) => {
-				invite.createdAt = moment(invite.createdAt).format("DD.MM.YYYY");
+				invite.createdAt = dayjs(invite.createdAt).format("DD.MM.YYYY");
```
adrianjost: yeah this is compatible 👍

tests/unit/setup.js
```diff
@@ -192,6 +192,13 @@ global.createModuleStore = (vuexModule, options = {}) => {
 };
 */
 
+// ===
+// Let tests fail on console.error
+// ===
+console.error = jest.fn((error) => {
+	throw `💔 ${error}`;
+});
```
adrianjost: We had some tests that we logging errors to the console, but are not failing. This trick converts all error logs into real errors and let the test fail.

src/components/ui/BaseSelect.vue
```diff
@@ -14,13 +14,12 @@
 			{{ label }}
 		</span>
 		<multi-select
+			v-model="vmodelProxy"
```
adrianjost: use v-model so we are not relying on the value prop implementation of multi-select. multi-select could now use whatever prop they wan't as a v-model without breaking our code

tests/unit/setup.js
ChrisWeissmann: Nice I was wondering about that while looking at travis earlier :)

Issue #165, opened 2019-04-25
Improve Docs Search


adrianjost:
Vuepress offers algolia integration. It's free to use and easy to integrate.

1. We just need to apply here:
https://community.algolia.com/docsearch/
1. And follow this vuepress docs entry:
https://vuepress.vuejs.org/default-theme-config/#algolia-search

ChrisWeissmann:
Applied and waiting for API key.

ChrisWeissmann:
Contacted them via discord. Apperently we fell out of the que. But are back in now :) Should be only a matter of time

ChrisWeissmann:
Apperently HPI thought the email with the key was spam. Found the key in the spam folder today and gave it a try successfully.

PR #193, opened 2019-05-24
Cw/improve docs search

ChrisWeissmann:
# Issue #165 

## Description

- Added the aligola integration to the project. 

## Screenshot
Bildschirmfoto 2019-05-24 um 13 03 43


adrianjost:
Is it okay to have this apikey publicly exposed or should we use an env variable instead?

ChrisWeissmann:
> Is it okay to have this apikey publicly exposed or should we use an env variable instead?

Very good point! I will change it

adrianjost:
Please also add the new variables to the docs: https://github.com/schul-cloud/nuxt-client/blob/cw/Improve-Docs-Search/docs/0-GettingStarted/1-Build.md
I would prefer to create a second table for each subproject to make it clear, which variable is required for which build.

adrianjost:
@ChrisWeissmann FIY: I removed your commits from history by force pushing over it to remove the key from the git history. It's still available with some tricks, but should be fine for now.

PR #198, opened 2019-05-31
Add real backend to deployment

adrianjost:
## Description

- moved api config from travis config to website and changed it to `https://api.test.schul-cloud.org`
- updated documentation
- hopefully fixed a docs deployment bug (variables not loaded)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,6 @@ env:
     - GIT_SHA=$( git rev-parse HEAD )
     - DOCKERTAG=$( echo $TRAVIS_BRANCH | tr -s "[:punct:]" "-" )
     - SC_THEME=default
-    - API_HOST=http://localhost:3030
 before_install:
   - openssl aes-256-cbc -K $encrypted_b7461320c5f4_key -iv $encrypted_b7461320c5f4_iv -in deploy/travis_rsa.enc -out deploy/travis_rsa -d
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,9 @@ install:
   - yarn global add codecov
   - yarn install --production=false
 script:
-  #  - yarn test:ci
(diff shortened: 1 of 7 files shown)


docs/2-Styles/0-Variables.md
```diff
@@ -16,30 +16,37 @@ p {
 
 ## Colors
 
-<<< @/src/themes/default/styles/variables/\_colors.scss
+<!-- prettier-ignore -->
+<<< @/src/themes/default/styles/variables/_colors.scss
```
adrianjost: prettier escapes the _ which makes the path invalid on unix systems. I haven't discovered it before, because it works fine on windows.

.travis.yml
```diff
@@ -10,26 +10,25 @@ env:
...
     - DOCKERTAG=$( echo $TRAVIS_BRANCH | tr -s "[:punct:]" "-" )
     - SC_THEME=default
-    - API_HOST=http://localhost:3030
 before_install:
   - openssl aes-256-cbc -K $encrypted_b7461320c5f4_key -iv $encrypted_b7461320c5f4_iv -in deploy/travis_rsa.enc -out deploy/travis_rsa -d
 install:
   - yarn global add codecov
   - yarn install --production=false
 script:
-  - yarn ci
```
ChrisWeissmann: Why remove this command andf adding the following?

docs/2-Styles/0-Variables.md
ChrisWeissmann: Nice find.

.travis.yml
adrianjost: Just for readability. If you wan't to know anything about travis (like what command are getting executed) you can simply look it up in one place. It's also better to debug, because you can disable commands faster. 
It's also a specific group of commands you will never use outside of travis.

PR #220, opened 2019-06-20
VUE-51 - Styling der Base Buttons

josafatburmeister:
# [VUE-51](https://ticketsystem.schul-cloud.org/browse/VUE-51)

## Description

- adapted tertiary button color
- added hero-cta / hero-cta fancy icon buttons
- replaced icon outline design by icon only design
- adapted colours of disabled buttons
- added disabled icon buttons 

## Screenshot

VUE-51-1
VUE-51-2
VUE-51-3


josafatburmeister:
> you can use simply `text` instead of `only`. This way you don't even need new css for that option. Your CSS implementation for `is-only` is identical to the implementation for `is-text`. This would also make the interface more consistent.
> 
> After doing so, you also don't need `<!-- eslint-disable max-lines -->` anymore.

Done 👍

src/components/ui/BaseButton.vue
```diff
@@ -139,9 +141,34 @@ export default {
...
+			width: 56px;
+			height: 56px;
+		}
+	}
+
 	&:disabled {
-		--button-background: var(--color-gray);
-		--button-text-color: var(--color-white);
+		--button-background: color-mod(var(--color-black) lightness(85%));
+		--button-text-color: color-mod(var(--color-black) lightness(62%));
```
adrianjost: Maybe we should extact those values into variables like:
`--color-disabled`, `--color-disabled-dark`, `--color-disabled-light`.

I am personally not 100% sure about this. For one, those values are looking pretty random (62 and 85) on the other hand we agreed that we wan't as few global variables as possible.

What do you think @dominikbruechner @ChrisWeissmann ?

Maybe it is possible to just simply reuse our `gray` variables here @josafatburmeister ? Even if it doesn't 100% match the colors on the design template given to you.

adrianjost:
Ignoring the one discussion point I found your pull looks good now and can be merged after we finished discussing the one controversial thing. You therefore can consider this pull request as done @josafatburmeister 😊.

adrianjost:
@josafatburmeister Sorry, I found one more issue. Your new tertiary button doesn't have a dedicated hover/focus style. Could you please add this?

src/components/ui/BaseButton.vue
dominikbruechner: If we extract them they might come in handy for any kind of disabled stuff... 

If we don't I would not change them, as we should not limit the UI designers work by implementing harsh technological constraints...

my vote goes for extract them into color-disabled, dark and light

josafatburmeister: I've extracted the values into variables in my last commit, this should be the cleanest solution. Because Mona has explicitly chosen these colors, I would not simply reuse the existing gray variables.

josafatburmeister:
> @josafatburmeister Sorry, I found one more issue. Your new tertiary button doesn't have a dedicated hover/focus style. Could you please add this?

Done 👍

dominikbruechner:
Please fix

PR #234, opened 2019-07-15
Full Fallback Master -> Stable

dominikbruechner:
merging full fallback


adrianjost:
I renamed the branches, because there were to many merge conflicts:
master => develop
stable => master

We now have the same git branches as in the legacy client and server.

PR #238, opened 2019-07-22
updated travis config

dominikbruechner:
Update travis config to run deploy script always, deploy script decides


dominikbruechner:
> Why?
> Wouldn't it be better to name the branches we wan't to be deployed explicitly?

the deploy script decides.. we can reorganize it the other way around too

PR #244, opened 2019-07-25
Easier Setup + Bugfix

smyth64:
# Description
- no need of global nuxt or gulp, using npx. (npx needs npm > 5.2)
- adding "yarn setup" for easy setup instead of doing everything manual.
- adding additional static files 
- updated getting started docs for easy setup
- fix possible crash in server/index.js
Crash happened, when I commented out the admin route in the src/server/routes.js and opened the url /administration/teams. The error was cannot read "messages" of undefined. It's because err.error was not defined.


package.json
```diff
@@ -5,15 +5,17 @@
...
-		"dev:nuxt": "nodemon --inspect --watch src/ --ignore src/legacy-client src/server",
+		"dev:nuxt": "npx nodemon --inspect --watch src/ --ignore src/legacy-client src/server",
 		"dev:storybook": "start-storybook -c stories/.storybook -p 4001",
 		"dev:docs": "vuepress dev docs",
 		"dev": "concurrently \"yarn dev:nuxt\" \"yarn dev:storybook\" \"yarn dev:docs\" -n \"nuxt,story,docs\" -c \"green,magenta,cyan\" -i false --kill-others-on-fail",
 		"build:nuxt": "cross-env NODE_ENV=production nuxt build && cross-env sh ./copy_static.sh",
 		"build:docs": "vuepress build docs -d dist/docs",
 		"build:storybook": "build-storybook -c stories/.storybook -o dist/storybook",
 		"build": "run-s build:*",
+		"gulp": "cd src/legacy-client && npx gulp",
```
adrianjost: Maybe:
```suggestion
		"build:nuxt:gulp": "cd src/legacy-client && npx gulp",
```
to make clear, that this build process is used for nuxt and not storybook or vuepress?

src/server/index.js
```diff
@@ -143,7 +144,11 @@ async function start() {
 	// this middleware is never called in this case
 	// eslint-disable-next-line no-unused-vars
 	app.use((err, req, res, next) => {
-		if (err.error.message.startsWith("Cast to ObjectId failed for value")) {
+		if (
+			err &&
+			err.error &&
+			err.error.message.startsWith("Cast to ObjectId failed for value")
```
adrianjost: Maybe we can use some new syntax for that and keep it short:
```suggestion
			err?.error?.message.startsWith("Cast to ObjectId failed for value")
```
This syntax was developed for exactly this use-case.

I think we then need this bable plugin:
https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining

package.json
```diff
@@ -5,15 +5,17 @@
 	"author": "HPI Schul-Cloud",
 	"private": true,
 	"scripts": {
-		"dev:nuxt": "nodemon --inspect --watch src/ --ignore src/legacy-client src/server",
+		"dev:nuxt": "npx nodemon --inspect --watch src/ --ignore src/legacy-client src/server",
```
adrianjost: Do I still have to install anything to be able to use npx?

src/server/index.js
smyth64: Wow awesome!! I was waiting for this feature since angularjs had it. Of course let’s use it. We can also use it in the whole project!!! Awesomely:)

package.json
smyth64: No, npx automatically comes with npm > 5.2 :)

adrianjost: nice

src/server/index.js
smyth64: I tried to add it but it's not working out of the box.

image


Lets better add it in another PR

smyth64:
@adrianjost can we merge this?

The ? syntax cannot be added so easy.

src/server/index.js
```diff
@@ -74,6 +74,7 @@ viewDirs.unshift(path.join(legacyClientRoot, `./theme/${themeName}/views/`));
 app.set("views", viewDirs);
 app.engine("hbs", wax.engine);
 app.set("view engine", "hbs");
+app.use(express.static(path.join(legacyClientRoot, "./static")));
```
adrianjost: Shouldn't we still send the build files?

adrianjost: When this is cleared, we can merge.

ChrisWeissmann:
Should this merge wait for the switch from yarn to npm and that way be adjusted?

adrianjost:
@ChrisWeissmann doesn't really matter. The conflicts shouldn't be huge.

src/server/index.js
smyth64: The files from there are needed. The login page used the css from this folder that’s why I added it.

adrianjost: Shouldn't those files be copied to the build folder when running gulp? I've never seen that we serve files from outside the build directory.

adrianjost: It's working totally fine without it. I removed it therefore.

PR #195, opened 2019-05-27
Internationalization integration.

ChrisWeissmann:
# Issue: [VUE-38](https://ticketsystem.schul-cloud.org/browse/VUE-38)

## Description

- Added internationalization with the nuxt-i18n plugin

PR #247, opened 2019-07-26
Vue 63/permission component

adrianjost:
# [Vue-63](https://ticketsystem.schul-cloud.org/browse/VUE-63)

## Description

- we have two new components to restrict access for users by permission/role. Please check the storybook for implementation details
  - http://stories.247.nuxt.schul-cloud.surge.sh/?path=/story/userhas--userhaspermission
  - http://docs.247.nuxt.schul-cloud.surge.sh/storybook/UserHas.html
- this could also act as an example how to mock the store in tests.
- and it adds support for vuex to storybook. (copied from #195)


adrianjost:
Tests are missing

PR #265, opened 2019-09-03
SC-1998 - legacy fallback proxy

adrianjost:
# Issue: [SC-1998](https://ticketsystem.schul-cloud.org/browse/SC-1998)

## Description

- new fallback logic for the legacy client
- all requests are now proxied by default to the legacy client. You need to whitelist new urls with a regex in the routes.js just like before.
- due to the new fallback logic you need to have the legacy client running in parallel to use this project. This enabled us to independently deploy the new and the old client.

Everything else is hopefully well enough documented in our doucmentation.

@FalkoBenthin new deployment infos:
http://docs.265.nuxt.schul-cloud.surge.sh/Changelog.html#_1-1-0


adrianjost:
code is ready, need some new tests

PR #268, opened 2019-09-03
Feature/vue 68 adapt spacings

dominikbruechner:
adapted spacing to suit designers' needs


src/themes/default/styles/utility/_spacing.scss
```diff
@@ -33,9 +33,9 @@
 }
 
 .mb--xl {
-	margin-bottom: var(--space-xl);
+	margin-bottom: var(--space-xxxl);
 }
 
 .mb--xxl {
-	margin-bottom: var(--space-xxl);
+	margin-bottom: var(--space-xxxxl);
```
adrianjost: this does not look logical
class xxl creates a spacing of xxxxl?

src/themes/default/styles/variables/_spacing.scss
```diff
@@ -6,7 +6,9 @@
...
 	--space-md: 1rem;
 	--space-lg: 1.5rem;
-	--space-xl: 3.25rem;
-	--space-xxl: 5.25rem;
-	--space-xxxl: 8.5rem;
+	--space-xl: 2rem;
+	--space-xxl: 2.5rem;
+	--space-xxxl: 3.25rem;
+	--space-xxxxl: 5.25rem;
+	--space-xxxxxl: 8.5rem;
```
adrianjost: These are a lot of x''es. I don't think is is any usefull. We should find a better naming scheme for spacing. It's already hard to count the number of x'es

src/themes/default/styles/utility/_spacing.scss
dominikbruechner: oh right, it was just a plain find and replace, good call

src/themes/default/styles/variables/_spacing.scss
dominikbruechner: we should talk to Mona if this makes sense for her, and what naming would suit her, as we need to talk the same language :)

dominikbruechner: and there are definitely too many x's, I got blurry vision just now from it

dominikbruechner: changed and resolved

PR #281, opened 2019-09-24
'feature/vue 80/breadcrump

luizzaly:
# Issue: New styled Breadcrump with a view tests

## Description

- last link should be always inactive

## Screenshot
Screenshot 2019-09-24 at 11 30 13


src/components/ui/BaseBreadcrumb.unit.js
```diff
@@ -6,19 +6,22 @@ describe("@components/BaseBreadcrumb", () => {
...
 			},
 		});
 
-		links.forEach((link) => {
-			const linkElement = link.to
-				? element.querySelector(`[to="${link.to}"]`)
-				: element.querySelector(`[href="${link.href}"]`);
-			expect(linkElement.textContent).toContain(link.text);
+		links.forEach((link, index) => {
+			if (links.length - 1 !== index) {
```
adrianjost: I think you just need this, because you have the url `/` two times specified. If you change `const emptyLink = { to: "/", text: "any" };` above to `const emptyLink = { to: "/empty", text: "any" };` or something else different than the other links, you can remove this if and keep everything else as it is

PR #284, opened 2019-10-05
Atomic Design Component Structure - Cleanup

adrianjost:
Some refacotring to organize our components by the atomic design specification.
The idea is from:
https://dev.to/maciekchmura/how-i-structure-a-react-project-3c2i
[Atomic Design - How to categorize](https://dev.to/maciekchmura/how-i-structure-a-react-project-3c2i)

But with the small extensions `legacy` and `base`. ``legacy` contains components that are needed for a consistent design when switching between the new and the legacy client (like the sidebar, topbar, ...).
`base` contains our base components that are always available and don't have to be imported manually.

The Component directory therefore now has 5 subdirectories
- atoms
- molecules
- organisms
- templates (currently only used for the error page, because for some reasons we can't store it in the pages directory)
- legacy
- base

Please share your opinions with the thumbs below or a comment.


dominikbruechner:
We should reorganize our storybook according to the new structure

adrianjost:
> We should reorganize our storybook according to the new structure (not in this PR any more)

Totally agree! This could be done with the storybook refactoring to the new syntax. With the new syntax the story components are supposed to be next to the components so the new structure would be easy to follow.

PR #309, opened 2019-10-17
Offlinesupport

janrenz:
Note that this add already add some nuxt routes for testing

Open TODOs:

- [ ] Prefetch Content
- [ ] Error State
- [x] Offline indicator 
- [ ] Disable Links to stuff that is not available offline
- [ ] Clear local Storage on Logout
- [ ] MultiUserCache


adrianjost:
Closing because of low priority and outdated PR. The changes are also easy to Google so we can do this again if we have time to do so. For now, I will close it to keep the PR list clean and filled with stuff we are actually working on.

PR #311, opened 2019-10-17
Impressum page

luizzaly:
# Issue: ...

## Description
Impressum Seite
-

## Screenshot
Screenshot 2019-10-17 at 15 29 02


src/pages/impressum.vue
```diff
@@ -1,5 +1,108 @@
...
+			HPI Schul-Cloud.</p
+		>
+	</div>
 </template>
+
+<script>
+import BaseIcon from "@basecomponents/BaseIcon";
+export default {
+	components: {
+		BaseIcon,
```
adrianjost: Are you sure, that you need to import the BaseIcon component? I think it is already mounted globally

src/serverMiddleware/routes.js
```diff
@@ -6,6 +6,7 @@
 // const mongoId = "[a-z0-9]{24}";
 
 module.exports = [
+	`^/impressum`,
```
adrianjost:
```suggestion
	`^/impressum/?`,
```
So if there gets any subpage added to the legacy client, we don't loose them.

src/pages/impressum.vue
```diff
@@ -1,5 +1,95 @@
...
+<script>
+export default {};
+</script>
+
+<style lang="scss" scoped>
+@import "@styles";
+
+p {
+	font-size: var(--text-lg);
+	line-height: var(--line-height-lg);
```
adrianjost: I think simple text-paragraphs should be of size --text-md (default) and not be scaled up. If we need scaling, we should active our base font size scaling.
1rem in our instance is currently 16px,
in the legacy client it is 18px. 
I think that's were the issue with small fonts is coming from.

src/themes/default/styles/utility/_typography.scss
```diff
@@ -13,8 +13,8 @@ body {
 @for $i from 1 through 6 {
 	h#{$i},
 	.h#{$i} {
-		margin-top: var(--space-sm);
-		margin-bottom: var(--space-xs-2);
+		margin-top: var(--space-xl);
+		margin-bottom: var(--space-sm);
```
adrianjost: is this communicated with mona and @dominikbruechner ?

stories/Impressum.stories.js
```diff
@@ -0,0 +1,17 @@
+import { storiesOf } from "@storybook/vue";
+import Impressum from "@pages/impressum";
+import notes from "@docs/storybook/base.md";
+
+storiesOf("Pages", module)
+	.addParameters({
+		notes,
+	})
+	.add("Impressum", () => ({
```
adrianjost: I don't think that pages should be documented with stories.
They are specific and should never get reused.
if we start documenting all our pages, we could start building the complete app in storybook ;D

=> Delete this file

PR #321, opened 2019-10-28
add sentry error logging

adrianjost:
# Issue: [SC-2032](https://ticketsystem.schul-cloud.org/browse/SC-2032)

## Description

- Add Sentry Error Reporting

### DIFF
--- a/deploy/compose-client.dummy
+++ b/deploy/compose-client.dummy
@@ -2,7 +2,6 @@
 version: \"3.7\"
 
 services:
-
   nuxtclient:
     image: schulcloud/schulcloud-nuxt-client:$DOCKERTAG
     deploy:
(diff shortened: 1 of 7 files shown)


deploy/compose-client.dummy
```diff
@@ -18,6 +17,7 @@ services:
       - TZ=Europe/Berlin
       - LEGACY_CLIENT_URL=http://client:3100
       - API_URL=https://api.test.schul-cloud.org
+      - SENTRY_DSN=https://18a7cbf1a7ef427ebcfc38429d18ba40@sentry.schul-cloud.dev/7
```
tofixx: would not use that locally, please update the token asap and remove it here :)

src/plugins/sentry.js
```diff
@@ -0,0 +1,18 @@
+import Theme from "@theme/config";
+
+export default ({ app, store }) => {
+	const { user } = store.state.auth;
+	app.$sentry.configureScope((scope) => {
+		scope.setLevel("warn");
+		scope.setTag("frontend", true);
+		scope.setTag("theme", Theme.name);
+		scope.setTag("domain", window.location.host);
```
tofixx: use SC_DOMAIN env instead?
is there no server part?
set git sha and release version

deploy/compose-client.dummy
adrianjost: thats not locally. That is our test system docker-compose file.

src/plugins/sentry.js
adrianjost: This is only the client part where we have no server

PR #334, opened 2019-11-04
feature/nuxtversion page

adrianjost:
## Description

- add the page /nuxtversion to show the git infos of the current deployment
- refactored the auth middleware to allow public pages
- added an integrationtest to check if the versionpage is accessible https://github.com/schul-cloud/integration-tests/pull/37

## Screenshot

image


nuxt.config.js
```diff
@@ -18,6 +31,7 @@ module.exports = {
 		FALLBACK_DISABLED: process.env.FALLBACK_DISABLED || false,
 		FEATURE_EXTENSIONS_ENABLED: process.env.FEATURE_EXTENSIONS_ENABLED || false,
 		FEATURE_TEAMS_ENABLED: process.env.FEATURE_TEAMS_ENABLED || false,
+		GIT_INFO,
```
adrianjost: I must get all the data in the nuxt.config because that's the only place I have access to node methods which are required to get git infos

dominikbruechner: I guess that's fine, its somewhat a config issue ;)
but maybe put the data object to another .js file?
const GIT_INFO = require("./git.info.js"); or sth

src/pages/error/proxy.vue
```diff
@@ -9,6 +9,9 @@ export default {
 		ErrorPage,
 	},
 	layout: "loggedout",
+	meta: {
+		isPublic: true,
```
dominikbruechner: adding info how to make pages public and how it works to Vuepress would be awesome

adrianjost:
Whoops sorry. I accidentaly merged this into master with my git gui. Didn't knew that this was even possible.
I will open another pull with the requested changes.

PR #346, opened 2019-11-14
Simple FAB button

dominikbruechner:
# Issue: [Vue-140](https://ticketsystem.schul-cloud.org/browse/VUE-140)

## Description

- Adding a simple fab button
- icon changeable
- position changeable 
- storybook cleanup

## Screenshot

Screen Shot 2019-11-14 at 11 30 55

### DIFF
--- a/docs/storybook/molecules/FloatingFab.md
+++ b/docs/storybook/molecules/FloatingFab.md
@@ -1,0 +1,6 @@
+# FloatingFab
+
+Floating Fab - Action Button
+
+Parameters:
+position: top-right, top-left, bottom-left, bottom-right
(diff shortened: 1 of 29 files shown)


janrenz:
A responsive Version would be nice (like the little version here https://material.io/components/buttons-floating-action-button/)

dominikbruechner:
> A responsive Version would be nice (like the little version here https://material.io/components/buttons-floating-action-button/)

what exactly do you mean?

janrenz:
the material one has a smaller version, that could automatically be used on smaller devices ..

dominikbruechner:
> the material one has a smaller version, that could automatically be used on smaller devices ..

material uses the mini one for different kinds of action:
Screen Shot 2019-11-15 at 15 54 52

But we need a second size, true

adrianjost:
@dominikbruechner as it's already merged, could we still discuss my open issues and maybe change them in a follow up PR?

PR #363, opened 2019-11-18
added the fab to news index

dominikbruechner:
adding fab to news index


src/pages/news/index.vue
```diff
@@ -1,8 +1,6 @@
 <template>
 	<div>
-		<base-button design="primary" @click="$router.push({ name: 'news-new' })">
-			Artikel anlegen
-		</base-button>
+		<floating-fab icon="add" @click="$router.push({ name: 'news-new' })" />
```
adrianjost: Is there any option to use a link instead of the button? 
Semantic wise should be everything that is linking to another page should be a link and everything that is triggering a js action should be a button.

dominikbruechner: do you suggest to rework the BaseButton? To be displayed as a style linked once it is linking to another page?
Otherwise we can plainly wrap it with a <a> tag... 
maybe i do not get your point
in vue every page switch is a ja action, right? if we are not proxying to the legacy client

adrianjost: in theory yes, every page switch is a js action in vue. But in Nuxt has ServerSide Rendering and may prerender pages. In this case it would be cool to have clean, working html so you can use most pages even without JS. 

Yes, this is totally focused on the implementation of the FAB. It just stood out to me here, because we are using it this ("wrong") way here.

I will approve it here and will create a new issue to refactor the FAB to accept a route prop and then switches to a link element.

janrenz:
Please remember to add screenshots for lazy mes.

PR #364, opened 2019-11-19
fixing fullscreen mode

dominikbruechner:
fullscreen mode now works properly

### DIFF
--- a/src/layouts/legacyLoggedIn.vue
+++ b/src/layouts/legacyLoggedIn.vue
@@ -1,7 +1,7 @@
 <template>
 	<!-- default template = loggedin view -->
 	<div>
-		<div class="page">
+		<div class="page" :style="style">
 			<div class="topbar">
 				<user-has-role :role="isDemoRole">
 					<demo-banner v-if="!fullscreenMode"></demo-banner>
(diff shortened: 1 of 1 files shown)


src/layouts/legacyLoggedIn.vue
```diff
@@ -153,6 +153,10 @@ export default {
 				return item;
 			});
 		},
+		style() {
+			if (this.fullscreenMode) return "display: inherit;";
+			else return "";
```
adrianjost:
```suggestion
			return this.fullscreenMode ? "display: inherit;" : "";
```

src/layouts/legacyLoggedIn.vue
```diff
@@ -153,6 +153,10 @@ export default {
 				return item;
 			});
 		},
+		style() {
+			if (this.fullscreenMode) return "display: inherit;";
+			return this.fullscreenMode ? "display: inherit;" : "";
```
adrianjost: @dominikbruechner you forgot to remove the first line!
Please fix this :)

dominikbruechner: yikes...

PR #366, opened 2019-11-19
Getting store data in Course View

dominikbruechner:
## Description
Preparations for new course view

adding service templates for lessons, groups, homeworks
retrieving data in course

-

## Screenshot
Screen Shot 2019-11-19 at 16 07 07

### DIFF
--- a/src/pages/courses/_id/index.vue
+++ b/src/pages/courses/_id/index.vue
@@ -11,18 +11,28 @@
 import { mapGetters } from "vuex";
 
 export default {
+	async asyncData({ store, params }) {
+		return {
+			lessons: await store.dispatch("lessons/find", {
+				query: {
+					courseId: params.id,
+				},
+			}),
+			homeworks: await store.dispatch("homeworks/find", {
+				query: {
+					courseId: params.id,
+				},
+			}),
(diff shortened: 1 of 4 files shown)


dominikbruechner:
@adrianjost please review and mention if you have a better store access pattern in mind. I reckon it's ok like this...

adrianjost:
We have two options:
1. Use nuxt asyncData and fetch the data driectly with the dispatch method
2. Like you implemented it

Solution 1 is cleaner and shorter but has the disadvantage that we don't get any updates from the store. The user must reload the page (or navigate away and back) to get new data.
As of now I don't see that we support that previously described use case, because we are not planning to use websockets or something similar to get push updates from the api. So I would use async data and refactor if needed at any time

adrianjost:
Example for async data from the store:
https://github.com/schul-cloud/nuxt-client/pull/353/files#diff-7b7e089fcdfd365e9cb09f39e57a4acdR63

dominikbruechner:
> Example for async data from the store:
> 
> https://github.com/schul-cloud/nuxt-client/blob/3647a8d18e7f4777be903b17287822adc4158674/src/pages/news/_id/edit.vue#L63-L67

moved to async...

PR #373, opened 2019-11-26
show context in news

adrianjost:
# Issue: No News Context available

## Description

- Some new Language Keys with variables. They now include the target Info as well

### legacy client

image

### old nuxt version

image

### new nuxt version

image


adrianjost:
- [ ] tests are missing

PR #379, opened 2019-12-02
add margin bottom to  inputs

adrianjost:
# Issue: Adjust Base Input Margins

## Description

- add a bit bottom margin to add space between mulitple inputs
- added some example stories for these cases

### Before
image 
### After
image

### DIFF
--- a/src/components/base/BaseInput/BaseInput.vue
+++ b/src/components/base/BaseInput/BaseInput.vue
@@ -3,6 +3,7 @@
 		:is="component"
 		:vmodel="vmodel"
 		v-bind="{ ...$attrs, ...$props }"
+		class="input"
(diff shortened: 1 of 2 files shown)


src/components/base/BaseInput/BaseInput.vue
```diff
@@ -82,4 +83,8 @@ export default {
 
 <style lang="scss" scoped>
 @import "@styles";
+
+.input {
+	margin-bottom: var(--space-sm);
```
dominikbruechner: I would go direction space-md or even space-lg.... otherwise all good

adrianjost: md is a good compromise. I have changed it.

PR #382, opened 2019-12-02
feature/multi os tests

adrianjost:
# Issue: Move Tests from Travis to GitHub Actions

## Description

- testing on multiple os's with build stages is a pain in travis so I moved over to github actions


adrianjost:
Codecov secret is missing

PR #394, opened 2019-12-09
Feature/sc 2844 create news production ready

dominikbruechner:
# Issue: SC-2844

## Description

-

## Screenshot
Screen Shot 2019-12-09 at 15 45 45

Screen Shot 2019-12-09 at 15 45 56


janrenz:
Kein Placeholder für den Inhalt?

dominikbruechner:
> Kein Placeholder für den Inhalt?

das wäre schick

dominikbruechner:
> > Kein Placeholder für den Inhalt?
> 
> das wäre schick

done

src/components/base/BaseInput/BaseInputDefault.vue
```diff
@@ -124,6 +128,9 @@ export default {
 		showLabel() {
 			return !!this.vmodel || !this.$attrs.placeholder;
 		},
+		addedClasses() {
+			return this.inputTeaser ? "h1" : "";
```
adrianjost: I would prefer having a prop "inputClass" instead of such a specific use case. That way we could more easily extend this.
We could also think about making this h1 input an own component. The BaseInput component already provides a slot for passing in your own input. But it's not part of this ticket and the current implementation is good enough.

src/plugins/directives.js
```diff
@@ -3,6 +3,12 @@ import Vue from "vue";
 import Ripple from "vue-ripple-directive";
 Vue.directive("ripple", Ripple);
 
+Vue.directive("focus", {
+	inserted: function(el, binding) {
```
adrianjost: 👍 
My only concerns are that the name may be somewhat unspecific. Maybe "focus-on-mount" or similar is better here?

dominikbruechner: yuppp

PR #395, opened 2019-12-09
remove padding from BaseCard and add it to other components which nee…

j-jmnz:
# Issue:  [SC-2759](https://ticketsystem.schul-cloud.org/browse/SC-2759)

## Description

- Removed padding from BaseCard
- Added padding to cards that needed BaseCard's padding


adrianjost:
@j-jmnz great work! The size of this pull request is perfect. Thank you very much for using the feedback from last week. Keep it up! :)

PR #413, opened 2019-12-20
Title Input Component refactoring for news

dominikbruechner:
Refactoring of the title input for create news.

- Added a TitleInput component, wrapping and customizing the BaseInput, allowing for a general design implementation for this kind of input fields
- enabled the BaseInput with a classes prop, to be customized if necessary


src/components/molecules/TitleInput.vue
```diff
@@ -0,0 +1,27 @@
...
+<script>
+export default {
+	inheritAttrs: false,
+	model: {
+		prop: "vmodel",
+		event: "update:vmodel",
+	},
+	props: {
+		vmodel: {
+			type: [Array, String, Number, Boolean],
```
adrianjost:
```suggestion
			type: [String, Number],
```
I think Array and Boolean are not really usefull for a TitleInput because they can only appear for checkboxes and radio buttons

dominikbruechner: yepp, was just copying

src/components/base/BaseInput/BaseInputDefault.vue
```diff
@@ -107,13 +107,13 @@ export default {
 				return supportedTypes.includes(type) || !type;
 			},
 		},
-		label: { type: String, required: true },
+		label: { type: String, default: "" },
```
adrianjost: I think the label should still be required because of accessibility issues. But there should be an option to hide the label and use an `aria-label` instead

src/components/base/BaseInput/BaseInputDefault.unit.js
```diff
@@ -118,4 +137,19 @@ describe("@components/BaseInputDefault", () => {
...
+				vmodel: "",
+				type: "password",
+				label: "test",
+				success: true,
+			},
+		});
+		const baseInputDefault = wrapper.find(BaseInputDefault);
+		expect(baseInputDefault.vm.passwordVisible).toBe(false);
+		baseInputDefault.vm.togglePasswordVisibility();
+		expect(baseInputDefault.vm.passwordVisible).toBe(true);
```
adrianjost: I think it would be better to toggle this by actually clicking on the button and expect that the input type changes according to these changes

PR #414, opened 2019-12-20
hotfix/mint ec error page

adrianjost:
## Description

- if a page can't get loaded from ghost, show a 404 error
- add error handler to view
- removed error hiding catches from store

## Screenshot

image


src/pages/mint-ec/_article.vue
```diff
@@ -33,30 +34,35 @@
...
+	async asyncData({ store, params, error, app: { i18n } }) {
+		try {
+			const page = await store.dispatch("ghost/getSinglePage", params.article);
+			return { page: page };
+		} catch (e) {
+			error({
+				statusCode: 404,
+				message: i18n.t("error.404"),
+			});
+		}
```
adrianjost: The new catch

src/store/ghost.js
```diff
@@ -10,40 +10,24 @@ const api = new GhostContentAPI({
 const module = {
 	actions: {
 		async getSinglePost(postSlug) {
-			return await api.posts
-				.read({
-					slug: postSlug,
-				})
-				.catch((err) => {
-					console.error(err);
-				});
```
adrianjost: removed all these error handler here because they have hidden the error from the views.

PR #419, opened 2020-01-06
Feature/VUE-BaseGrid-ColumnGap

j-jmnz:
# Issue: Feature/VUE-BaseGrid-ColumnGap

## Description

- Add columnGap dynamic property


src/components/base/BaseGrid.vue
```diff
@@ -10,14 +10,16 @@ export default {
 			type: String,
 			default: "15rem",
 		},
+		columnGap: {
```
adrianjost: Shouldn't this be named gridGap if it is actually just setting the grid gap?

j-jmnz: @dominikbruechner recommended to name this property as such, please let me know if it should be changed.

dominikbruechner: I am unpassionate about it, feel free to rename it to gridGap.

PR #443, opened 2020-01-20
enable `jest/no-test-return-statement` rule

adrianjost:
## Description

- depends on #442
- enables `jest/no-test-return-statement` rule. 

**Why?**
https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-test-return-statement.md


adrianjost:
merged with #444

PR #470, opened 2020-01-22
Clear Auth Store after Logout

dominikbruechner:
Till now just the cookie was deleted, but we want to clear the token from the store as well (+ user data etc.)

Let me know if I just haven't seen it. @adrianjost


adrianjost:
great catch @dominikbruechner 

after doing a `npm i` please restart your VSCode to use the new prettier version and fix your lint issues

PR #500, opened 2020-01-27
Added datasources to sidebar

adrianjost:
# Issue: [SC-3240](https://ticketsystem.schul-cloud.org/browse/SC-3240)

## Description

- Client Pull: https://github.com/schul-cloud/schulcloud-client/pull/1581

## Screenshot

image


src/utils/sidebarBaseItems.js
```diff
@@ -144,6 +144,12 @@ export default [
 				icon: "building",
 				href: "/administration/school/",
 			},
+			{
+				title: "Datenquellen",
+				icon: "database",
+				permission: "DATASOURCES_VIEW",
```
dominikbruechner: don't forget to add the permission to the user in auth.js ....

adrianjost: This should be provided by the database and we don't need to add it manually. Not all users should have this.

dominikbruechner: y

PR #522, opened 2020-01-28
added aria-label to FAB/BaseButton

dominikbruechner:
# Issue: ...

## Description

-


dominikbruechner:
> Maybe we should even make the aria-label required for the FloatingFAB Component. Currently there is never a text in the FAB so a aria-label is a good idea.
> Everything else looks good 👍

done

dominikbruechner:
> I think it isn't working anymore on the FAB because ariaLabel is now in $props and not in $attrs. I think you need to pass it explicitly to the button in the FAB Component
> image

oh didn't realise it, yes this fixed it

PR #524, opened 2020-01-28
Feature/screenshot test crop screenshots to content

adrianjost:
# Issue: Screenshot Tests should fail earlier

Sorry for the gigantic PR! Most of the changed files are only the new cropped screenshots.

## Description

- take full screenshots using puppeteer so long stories will be captured completly
- mocked all external images with base64 or repo internal Images
- auto crop screenshots to its actual content using sharp
- adjusted some stories to work with this setup (puppeteer screenshots actually have some bugs)
    - puppeteer only takes a really small screenshot if only `position: fixed` or `absolute` elements are on the page. The solution is to add a grey rectangle anywhere on the page (was required for FloatingFAB component)
    - added some rectangles around some stories for the same reason but I couldn't find the reason for that.
- reduced the diff threshold to fail more often (should not happen that often anymore, because all screenshot updates are on the same system when using the sc-bot update command)
- added an documentation hint to the screenshot action, how to update screenshots.


adrianjost:

PR #549, opened 2020-01-30
SC-3305 | add pagination to datasource overview page

BlackGhostHD:
# Issue: https://ticketsystem.schul-cloud.org/browse/SC-3305

## Description

-

## Screenshot
Anmerkung 2020-02-04 141410


src/components/base/BaseContentContainer.unit.js
```diff
@@ -3,4 +3,18 @@ import BaseContentContainer from "./BaseContentContainer";
...
+	});
+
+	it("renders prop size small", () => {
+		const wrapper = mount(BaseContentContainer, {
+			propsData: {
+				size: "small",
+			},
+		});
+		expect(wrapper.find(".small").exists()).toBe(true);
+	});
```
adrianjost: Why do these two tests which test similar things use differnt approaches to test it? Please unify this

src/pages/administration/datasources/add.vue
```diff
@@ -1,40 +1,49 @@
...
-			class="mb--md"
-		>
-			<template v-slot:actions>
-				<BaseButton design="primary text">
-					<BaseIcon source="custom" icon="datasource-add" />
-					{{ $t("pages.administration.datasources.new.add") }}
-				</BaseButton>
-			</template>
-		</datasource-card>
+		<user-has-permission permission="DATASOURCES_VIEW">
```
adrianjost: seems like this pull uses a pretty outdates develop base. Instead of a wrapping user-has-permission you can add the hasPermission or something like this to the script section

PR #569, opened 2020-02-03
Feature/storybook icons

dominikbruechner:
- Reordered Storybook
- Added Custom Icon Overview
Screen Shot 2020-02-03 at 12 07 33


adrianjost:
I like that. But as a follow up ticket it would be nice to add a small searchbar ;D Shouln't be that complicated, just a simple `this.iconNames.filter(name => name.includes(this.searchString))`

PR #570, opened 2020-02-03
Title limit in datasource card

luizzaly:
## Description

-Title output in Datasource Card is now limited to 15 characters

## Screenshot

Screenshot 2020-02-03 at 12 25 40


adrianjost:
@dominikbruechner what do you think about this?
I would prefer to limit it using css to the actual available width and add an option to show the full name. I think it's not a good idea to cut off the title that strictly. But because the ticket system is offline, I can't check what the requirements were.

dominikbruechner:
> @dominikbruechner what do you think about this?
> I would prefer to limit it using css to the actual available width and add an option to show the full name. I think it's not a good idea to cut off the title that strictly. But because the ticket system is offline, I can't check what the requirements were.

yes, same, I have added a codepen example here

adrianjost:
@luizzaly please resolve conflicts

adrianjost:
@dominikbruechner @luizzaly shouldn't the subtitle be also get the dots at the end?
image

PR #574, opened 2020-02-04
Feature/datasource login component

CordlessWool:
# Issue: ...
https://ticketsystem.schul-cloud.org/browse/SC-3281

## Description
Added a base Login form for datasources and the specific form for webunits

## Screenshot
image

image

image


src/pages/administration/datasources/webuntis/new.vue
```diff
@@ -45,7 +45,7 @@
 						v-model="config.url"
 						type="text"
 						label="URL"
-						:placeholder="'erato.webuntis.com'"
+						:placeholder="'URL der Datenquelle'"
```
adrianjost: This should rather be a hint and the old Placeholder `erato.webuntis.com` should stay.
Placeholders should hint the user what he could enter in the input. This descriptive one you now have chosen is more a description and the user still does not know if he have to use https:// in front of it and such stuff which an actual example would hint him.

https://www.nngroup.com/articles/form-design-placeholders/

PR #602, opened 2020-02-10
feature/VUE-173, VUE-176 DataTable refactoring

adrianjost:
# Issue: [VUE-173](https://ticketsystem.schul-cloud.org/browse/VUE-173), [VUE-176](https://ticketsystem.schul-cloud.org/browse/VUE-176)

## Description

- simplify API of DataTable
- implement sorting in DataTable
- add a lot of tests for DataTable (not BackendDataTable)

Unfortunately the PR is pretty huge but here is a list of features that are tested so everything required should work:

image

image


.vscode/launch.json
```diff
@@ -19,7 +19,7 @@
 			"request": "launch",
 			"name": "Test (unit) Current File",
 			"program": "${workspaceFolder}/node_modules/.bin/jest",
-			"args": ["${fileBasename}", "--config", "jest.config.js"],
+			"args": ["/${fileBasename}", "--config", "jest.config.js"],
```
adrianjost: required for testfiles that have the same name like the directory they are in. In this case to debug `DataTable/DataTable.unit.js`

PR #653, opened 2020-02-20
SC-3442 datasources: added attribute new-pwd, avoids pre-fill username in ff, chrome, safari

dominikbruechner:
# Issue: [SC-3442](https://ticketsystem.schul-cloud.org/browse/SC-3442)

## Description

-


src/pages/administration/datasources/webuntis/new.vue
```diff
@@ -51,6 +51,7 @@
 						label="Passwort"
 						:placeholder="'WebUntis Passwort'"
 						class="mt--md"
+						autocomplete="new-password"
```
adrianjost: I think 
```suggestion
						autocomplete="off"
```
is semanticly better suited here, because new-password seems to indicate that submitting this form updates the currently saved login credentials.

adrianjost: I think 
```suggestion
						autocomplete="off"
```
is semanticly better suited here, because new-password seems to indicate that submitting this form updates the currently saved login credentials.

dominikbruechner: autocomplete off does not work for all browsers unfortunately... it's already set for the form
autocomplete="new-password" seems the only option to make it work for all browsers. 
I did some research...

adrianjost: agree. Just didn't knew `new-password` before and haven't found that little piece of information.

PR #663, opened 2020-02-24
feature/SC-3437 checkbox refactoring

adrianjost:
# Issue: [SC-3437](https://ticketsystem.schul-cloud.org/browse/SC-3437), [VUE-170](https://ticketsystem.schul-cloud.org/browse/VUE-170)

## Description

- refactor checkbox and switch input types to use BaseIcon Component with svg files
- implemented new indetermined mode for `type=checkbox`

## Screenshot

image

image


src/components/base/BaseInput/BaseInputCheckbox.vue
```diff
@@ -51,22 +54,63 @@ export default {
...
 		},
 	},
+	created() {
+		this.validateProps();
+	},
 	methods: {
+		// Perform more complex prop validations than is possible
+		// inside individual validator functions for each prop.
+		validateProps() {
+			if (process.env.NODE_ENV === "production") return;
```
CeEv: How process.env. is mapped, after build?

adrianjost: no, just during build time. Nuxt does not support runtime NODE_ENV checks. Actually it compiles static values in during build.

PR #664, opened 2020-02-24
Feature/sc-3358-content-searchbar

j-jmnz:
# Issue: [SC-2760](https://ticketsystem.schul-cloud.org/browse/SC-2760)

## Description

- Searchbar implementation

-

## Screenshot

Screenshot from 2020-02-24 14-35-39

Screenshot from 2020-02-24 14-36-06


src/components/molecules/ContentSearchbar.vue
```diff
@@ -0,0 +1,140 @@
+<template>
+	<div class="search">
+		<div :class="[isActive ? 'input-active' : '', 'search__container']">
+			<input
+				ref="searchInput"
+				v-autowidth="{ maxWidth: '960px', minWidth: '0px', comfortZone: 40 }"
```
adrianjost:
```suggestion
				v-autowidth="{
					minWidth: 'var(--size-content-width-min)',
					maxWidth: 'var(--size-content-width-max)',
					comfortZone: 40,
				}"
```

40 seems to be a magic number. Can you explain me why you have chosen `40`? ([Why magic numbers are bad](https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad))

src/components/molecules/ContentSearchbar.vue
```diff
@@ -0,0 +1,140 @@
...
+		updateSearchString(newValue) {
+			this.inputValue = newValue;
+			this.$emit("input", this.inputValue);
+		},
+		enterKeyHandler() {
+			if (this.isActive && this.inputValue.length > 0) {
+				this.isActive = false;
+				this.inputValue = "";
+			}
+			this.$emit("enter-key-event");
```
adrianjost:
```suggestion
		enterKeyHandler(...args) {
			if (this.isActive && this.inputValue.length > 0) {
				this.isActive = false;
				this.inputValue = "";
			}
			this.$emit("keyup:enter", ...args);
```
the event name is not really declerative. From the current name it is not clear when this event will be triggered.
I also suggest passing the event itself through just in case the parent needs it.

src/pages/content/index.vue
```diff
@@ -112,6 +118,14 @@ export default {
...
 				return;
 			}
+			/**
+			FOR ALTERNATIVE IMPLEMENTATION:
+			Activate transition after 3 key presses
+
+			if (this.searchQuery.length >= 3) {
+				this.transitionHandler();
+			}
+			**/
```
adrianjost:
```suggestion
```
Please remove unused code.

src/components/molecules/ContentSearchbar.vue
```diff
@@ -0,0 +1,140 @@
...
+					icon="search"
+					@click="enterKeyHandler"
+				/>
+				<base-icon
+					v-else
+					class="search-icon"
+					source="custom"
+					icon="clear"
+					@click="clearBtnHandler"
+				/>
```
adrianjost: These should be wrapper by a BaseButton to be accessible. Otherwise you can not focus them with the keyboard. Everything that you can click should always also be reachable by keyboard.

You can read our accessibility guidelines [here](https://docs.schul-cloud.org/pages/viewpage.action?pageId=113641342) (2.1)

package.json
```diff
@@ -63,6 +63,8 @@
 		"vue": "^2.6.11",
 		"vue-echarts": "^4.0.4",
 		"vue-flatpickr-component": "^8.1.5",
+		"vue-i18n": "^8.15.3",
+		"vue-input-autowidth": "^1.0.9",
 		"vue-i18n": "^8.15.4",
```
adrianjost: seems like `vue-i18n` is now defined twice 🤔
```suggestion
		"vue-i18n": "^8.15.4",
		"vue-input-autowidth": "^1.0.9",
```

src/components/molecules/ContentSearchbar.stories.js
```diff
@@ -0,0 +1,34 @@
...
+
+import ContentSearchbar from "@components/molecules/ContentSearchbar";
+
+const template = `
+	<content-searchbar
+		v-model.lazy="searchQuery"
+		:class="
+			!activateTransition ? 'first-search__searchbar' : 'content__searchbar'
+		"
+		placeholder="Lernstore durchsuchen"
```
adrianjost:
```suggestion
		:placeholder="placeholder"
```
It's a good idea to actually use the knobs you defined

src/components/molecules/ContentSearchbar.stories.js
```diff
@@ -0,0 +1,34 @@
+import { storiesOf } from "@storybook/vue";
+import { text } from "@storybook/addon-knobs";
+import { select } from "@storybook/addon-knobs";
```
adrianjost:
```suggestion
import { text, select } from "@storybook/addon-knobs";
```
you can combine these imports

src/components/molecules/ContentSearchbar.stories.js
```diff
@@ -0,0 +1,34 @@
...
+
+const template = `
+	<content-searchbar
+		v-model.lazy="searchQuery"
+		:class="
+			!activateTransition ? 'first-search__searchbar' : 'content__searchbar'
+		"
+		placeholder="Lernstore durchsuchen"
+		:loading="loading"
+		@keyup:enter="transitionHandler"
```
adrianjost: The Attribute values are not defined :/ (`loading`, `transitionHandler`)

- [ ] For `loading` you could define a boolean knob I guess ([example](https://github.com/schul-cloud/nuxt-client/blob/1f097a4fc49602991b49dd0a53e8578a0a9fed23/src/components/base/BaseInput/BaseInputCheckbox.stories.js#L18))

- [ ] And for `transitionHandler` you could define an action to log all the events in storybook.
```
import { action } from "@storybook/addon-actions";

storiesOf("5 Molecules/ContentSearchbar", module).add("ContentSearchbar", () => ({
		// ...
		methods: {
			transitionHandler: action("transitionHandler"),
		}
}));
```

src/components/molecules/ContentSearchbar.stories.js
```diff
@@ -0,0 +1,34 @@
...
+		data: () => ({
+			searchQuery: text("searchQuery", ""),
+			placeholder: text("placeholder", "Lernstore durchsuchen"),
+			activateTransition: select(
+				"activateTransition",
+				{ true: true, false: false },
+				false
+			),
+		}),
+		template,
```
adrianjost: I would recommend moving the template directly into the story instead of having it in a seperate variable because the template is highly coupled to the data defined explicitly for a specific story. If you ever need to add another example/story for this component you would otherwise have problems finding a suitable name and everything gets messy because you would have to scroll a lot to find the template for a story.

src/components/molecules/ContentSearchbar.unit.js
```diff
@@ -0,0 +1,71 @@
...
+		const testString = "teststring";
+		const wrapper = mount(contentSearchbarComponent);
+		const inputField = wrapper.find("input");
+		expect(wrapper.vm.value).toBe("");
+		inputField.trigger("focus");
+		inputField.setValue(testString);
+		expect(wrapper.vm.value).toBe(testString);
+	});
+
+	it("Emits input event on ANY keypress in input", async () => {
```
adrianjost:
```suggestion
	it("Emits input event on any input value change", async () => {
```
I would suggest this name, because you are not testing any keypress events with this test.

src/pages/content/index.vue
```diff
@@ -132,6 +146,12 @@ export default {
 		async searchContent() {
 			await this.$store.dispatch("content/getResources", this.query);
 		},
+		transitionHandler() {
+			this.activateTransition = true;
+			setTimeout(() => {
+				this.firstSearch = false;
+			}, 500);
```
adrianjost: Is it necessary to set `this.activateTransition = false;` after a defined amount of time?

src/pages/content/index.vue
```diff
@@ -165,4 +187,22 @@ export default {
...
+		width: 100%;
+		padding: var(--space-md) 0;
+		margin: var(--space-md) 0;
+		transform: scale(1.3) translateY(250%);
+	}
+}
+
+.fade-enter-active,
+.fade-leave-active {
+	transition: opacity 0.5s;
```
adrianjost: Please use a variable for the transition time:
[`var(--duration-transition-slow)`](https://github.com/schul-cloud/nuxt-client/blob/bade10e0a51014f2a1ba2198653a12055fa46a5f/src/themes/default/styles/variables/_durations.scss#L4)
```suggestion
	transition: opacity var(--duration-transition-slow);
```

src/pages/content/index.vue
adrianjost: Instead of using a timeout to set the prop you could use an event handler that hooks to the transition states.
Check out these cool hooks provided by the transition component: https://vuejs.org/v2/guide/transitions.html#JavaScript-Hooks
You propably need `v-on:after-enter="afterEnter"`

PR #691, opened 2020-02-27
added permission check for single tab

BlackGhostHD:
## Description
added permission check for single tab

-

### DIFF
--- a/src/components/atoms/Tab.vue
+++ b/src/components/atoms/Tab.vue
@@ -1,19 +1,18 @@
 <template>
-		<div v-show="isActive && hasPermission" class="tab-content">
-			<slot></slot>
-		</div>
+	<div v-show="isActive && hasPermission" class="tab-content">
+		<slot></slot>
+	</div>
 </template>
 
 <script>
+import UserHasPermission from "@/mixins/UserHasPermission";
+
 export default {
+	mixins: [UserHasPermission],
 	props: {
 		name: {
 			type: String,
 			required: true,
-		},
-		permission: {
-			type: String,
-			default: "",
 		},
 		selected: {
 			type: Boolean,
@@ -27,14 +26,7 @@ export default {
 	},
 	created() {
 		this.isActive = this.selected;
-		this.hasPermission = this.checkPermission();
-	},
-	methods: {
-		checkPermission() {
-			return (
-				!this.permission || this.$user.permissions.includes(this.permission)
-			);
-		},
+		this.hasPermission = this.$_hasPermission;
 	},
 };
 </script>
--- a/src/components/atoms/Tab.vue
+++ b/src/components/atoms/Tab.vue
@@ -1,6 +1,8 @@
 <template>
-	<div v-show="isActive" class="tab-content">
-		<slot></slot>
+	<div v-if="hasPermission">
+		<div v-show="isActive" class="tab-content">
+			<slot></slot>
+		</div>
 	</div>
 </template>
 
@@ -11,6 +13,10 @@ export default {
 			type: String,
 			required: true,
 		},
+		permission: {
+			type: String,
+			default: "",
+		},
 		selected: {
 			type: Boolean,
 		},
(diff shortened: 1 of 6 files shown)


src/components/organisms/Tabs/Tabs.vue
```diff
@@ -6,9 +6,11 @@
 				:key="tab.name"
 				:class="{ 'is-active': tab.isActive }"
 			>
-				<button class="tab-button" @click="selectTab(tab)">
-					<span>{{ tab.name }}</span>
-				</button>
+				<div v-if="tab.hasPermission">
```
adrianjost: We should not introduce another div here. You could add the v-if to the button itself or use a template tag which will not be rendered and therefore does not pollute the compiled HTML output with unnecessary elements.

src/components/atoms/Tab.vue
```diff
@@ -11,17 +13,30 @@ export default {
...
 		};
 	},
 	created() {
 		this.isActive = this.selected;
+		this.hasPermission = this.checkPermission();
+	},
+	methods: {
+		checkPermission() {
+			return (
+				!this.permission || this.$user.permissions.includes(this.permission)
```
adrianjost: Instead of implementing the permission check again, isn't it possible to reuse the UserHasPermission component?
https://github.com/schul-cloud/nuxt-client/blob/develop/src/components/helpers/UserHasPermission.vue

I'm not 100% sure how to do this but maybe you can import the component as a mixing or extract the permission check logic into a mixing and only implement the permission check in a single place. I would prefer to have only a single place to maintain this critical logic.

src/components/atoms/Tab.vue
```diff
@@ -1,6 +1,8 @@
 <template>
-	<div v-show="isActive" class="tab-content">
-		<slot></slot>
+	<div v-if="hasPermission">
```
adrianjost: Again, I think you can get rid of this wrapping div by righter using a template tag, moving the v-if to the child or using a && hasPermission in the v-show.

src/components/atoms/Tab.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-	<div v-show="isActive" class="tab-content">
-		<slot></slot>
-	</div>
+		<div v-show="isActive && hasPermission" class="tab-content">
+			<slot></slot>
+		</div>
```
adrianjost:
```suggestion
	<div v-show="isActive && hasPermission" class="tab-content">
		<slot></slot>
	</div>
```

src/mixins/UserHasPermission.js
```diff
@@ -0,0 +1,34 @@
...
+		...mapState({
+			userPermissions: (state) =>
+				state && state.auth && state.auth.user && state.auth.user.permissions
+					? state.auth.user.permissions.map((p) => p.toLowerCase())
+					: [],
+		}),
+		$_hasPermission() {
+			// empty or undefined permission will return true
+			// console.log(this.permission);
+			// console.log(
```
janrenz: this should be removed

src/mixins/UserHasPermission.js
```diff
@@ -0,0 +1,26 @@
...
+		$_hasPermission() {
+			return typeof this.permission === "string"
+				? !this.permission ||
+						this.userPermissions.includes(this.permission.toLowerCase())
+				: !this.permission() || this.permission(this.userPermissions);
+		},
+	},
+};
+
+export default UserHasPermission;
```
adrianjost:
```suggestion
export default {
	props: {
		permission: {
			type: [String, Function],
			default: () => () => false,
		},
	},
	computed: {
		...mapState({
			userPermissions: (state) =>
				state && state.auth && state.auth.user && state.auth.user.permissions
					? state.auth.user.permissions.map((p) => p.toLowerCase())
					: [],
		}),
		$_hasPermission() {
			return typeof this.permission === "string"
				? !this.permission ||
						this.userPermissions.includes(this.permission.toLowerCase())
				: !this.permission() || this.permission(this.userPermissions);
		},
	},
};
```

to keep the codebase consistent you could directly export the mixin instead of storing it in a variable before.

adrianjost:
@BlackGhostHD will rebase this to develop today

BlackGhostHD:
This PR can now be found here: https://github.com/schul-cloud/nuxt-client/pull/751

PR #697, opened 2020-03-03
Vue 58 integrate universal filter ui into filter menu

josafatburmeister:
# Issue: [Vue-58](https://ticketsystem.schul-cloud.org/browse/VUE-58)

## Description

- removed FilterMenu from DataTable and moved it to standalone component FilterMenuStandalone
- started to integrate [vue-filter-ui](https://github.com/adrianjost/vue-filter-ui) into FilterMenuStandalone


package.json
```diff
@@ -61,6 +61,7 @@
 		"tiptap": "^1.26.6",
 		"tiptap-extensions": "^1.28.6",
 		"vue": "^2.6.11",
+		"vue-clickout": "^1.0.0",
```
adrianjost: We already have a solution for that with the `v-click-outside` directive.
You can find the implementation [here](https://github.com/schul-cloud/nuxt-client/blob/develop/src/plugins/directives.js#L12).

I think this is doing exactly the same.

src/mixins/defaultFilters.unit.js
```diff
@@ -0,0 +1,135 @@
...
+		expect(wrapper.vm.filterTextContains("test", "te")).toBe(true);
+		expect(wrapper.vm.filterTextContains("test", "a")).toBe(false);
+		expect(wrapper.vm.filterTextContains("Test", "te")).toBe(true);
+		expect(wrapper.vm.filterTextContains(11, 1)).toBe(true);
+		expect(wrapper.vm.filterTextContains(11, 2)).toBe(false);
+
+		expect(wrapper.vm.filterTextEquals("test", "test")).toBe(true);
+		expect(wrapper.vm.filterTextEquals("test", "te")).toBe(false);
+		expect(wrapper.vm.filterTextEquals("Test", "test")).toBe(true);
+	});
```
adrianjost: Maybe you can use `test.each`/`it.each` to auto generate a lot of tests and group them in a describe called `can filter strings`. Then all the different test cases are independent.

I find this feature of jest really helpfull. You can read more here: https://jestjs.io/docs/en/api#testeachtablename-fn-timeout

src/components/organisms/FilterUI/Filter.vue
```diff
@@ -0,0 +1,379 @@
...
+	props: {
+		labelAdd: { type: String, default: "add filter" },
+		labelApply: { type: String, default: "apply" },
+		labelCancle: { type: String, default: "cancle" },
+		labelRemove: { type: String, default: "remove" },
+		/*
+    "handleUrl": { type: Boolean },
+    "saveState": { type: Boolean },
+		"consistentOrder": {type: Boolean, default: true},
+		*/
```
adrianjost:
```suggestion
```

src/components/organisms/DataTable/DataTable.vue
```diff
@@ -230,7 +231,9 @@ export default {
 				// sort strings
 				return a.localeCompare(b);
 			});
-			return sortOrder !== "desc" ? sortedData : sortedData.reverse();
+			return sortOrder !== "desc"
+				? sortedData
+				: sortedData.reverse();
```
adrianjost: I think you can safely revert the changes in this file. Especially the `eslint-disable max-lines` does not seem to be necessary.

josafatburmeister:
>     * [ ]  1. Is there a reason why you don't have [npm installed vue-filter-ui](http://docs.vue-filter-ui.surge.sh/1-GettingStarted.html) and instead copied over a bunch of files?
> 
> 
> Where there bugs in the implementation? If so, please let me know and we could fix them directly the original implementation. I can give you write access to the repo as well.
> 
> I propably missed reviewing a lot of code changes but let's start with this and I will continue reviewing.

The main reason for my decision was that the parser most likely needs to be moved from the vue-filter-ui to the FilterMenuStandalone component and therefore the query api of the vue-filter-ui has to be changed. This could imply some bigger changes, but if you agree we can do the changes on the original repo of cause.

src/mixins/defaultFilters.unit.js
josafatburmeister: Done :+1:

package.json
josafatburmeister: If we import vue-filter-ui as npm package we just can remove this again.

adrianjost: Ahh I see. It was just required because you copied over the vue-filter-ui

josafatburmeister:
> > ```
> > * [ ]  1. Is there a reason why you don't have [npm installed vue-filter-ui](http://docs.vue-filter-ui.surge.sh/1-GettingStarted.html) and instead copied over a bunch of files?
> > ```
> > 
> > 
> > Where there bugs in the implementation? If so, please let me know and we could fix them directly the original implementation. I can give you write access to the repo as well.
> > I propably missed reviewing a lot of code changes but let's start with this and I will continue reviewing.
> 
> 
> The main reason for my decision was that the parser most likely needs to be moved from the vue-filter-ui to the FilterMenuStandalone component and therefore the query api of the vue-filter-ui has to be changed. This could imply some bigger changes, but if you agree we can do the changes on the original repo of cause.

The vue-filter-ui package is imported now instead 👍

src/components/organisms/FilterUI/Filter.vue
josafatburmeister: Not necessary anymore because the due-filter-ui package is imported now.

adrianjost:

adrianjost:
funktioniert zwar nur so semi gut, aber das fixen wir später in kleineren Tickets :)

adrianjost:
braucht noch paar tests, bevor automerge funktioniert :/ Entweder wir schreiben welche oder mergen manuell

PR #700, opened 2020-03-05
feature/GitHub integration tests

adrianjost:
# Issue

The deployment process takes really long, because the integration tests have to pass before the build starts. This is not necessary because the IT must have already passed before anything gets merged (straight forward) into the branch to deploy.

## Description

To solve this problem I tried to move the integration tests out of travis to github actions so they can run in parallel.


adrianjost:
#788

PR #734, opened 2020-03-16
Students Admin Page with Backend-datatable

dominikbruechner:
https://ticketsystem.schul-cloud.org/browse/SC-3628
https://ticketsystem.schul-cloud.org/browse/SC-3634
https://ticketsystem.schul-cloud.org/browse/SC-3652

## Description

Birthday willl be extra ticket
Legend will be extra ticket

-added students admin page
-added store action
-added backend data table
-added consent icons
-added date mapping
-added i18n
-added pagination

## Screenshot

Screen Shot 2020-03-17 at 09 10 30


src/serverMiddleware/routes.js
```diff
@@ -7,6 +7,7 @@
 
 module.exports = [
 	`^/administration/datasources?`,
+	`^/administration/students?`,
```
adrianjost: Let's not commit this at the current situation where we release often. Otherwise we may break the system with our unfinished page.
```suggestion
	// `^/administration/students?`,
```

PR #735, opened 2020-03-16
Feature/SC-3624 expendable FAB

adrianjost:
# Issue: [SC-3624](https://ticketsystem.schul-cloud.org/browse/SC-3624)

## Description

- implemented expendable icon fab
- use new FAB in FabFloating
- renamed FloatingFab to FabFloating to follow our naming schema (least specific part first)
- FAB with text content will not be part of this PR because it is currently not required!

## Screenshot

image


adrianjost:
> Nice!
> Just double the spacing between the icon buttons
> Maybe add backdrop?

Mäh, more spacing looks weird and unrelated
image

I thought about an backdrop as well but decided to skip it because of time constrains. We can add this later on :)

dominikbruechner:
> > Nice!
> > Just double the spacing between the icon buttons
> > Maybe add backdrop?
> 
> @dominikbruechner
> Mäh, more spacing looks weird and unrelated
> image
> 
> I thought about an backdrop as well but decided to skip it because of time constrains. We can add this later on :)

not so much more, just a little bit more

dominikbruechner:
but storybook looks fine now

dominikbruechner:
screenshot tests need fix

PR #739, opened 2020-03-17
add add_student fab

dominikbruechner:
Screen Shot 2020-03-17 at 11 43 15

### DIFF
--- a/src/pages/administration/students.vue
+++ b/src/pages/administration/students.vue
@@ -48,12 +48,34 @@
 				</span>
 			</template>
 		</backend-data-table>
+		<fab-floating
+			position="bottom-right"
+			:show-label="true"
+			:actions="[
+				{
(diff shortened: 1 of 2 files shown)


src/pages/administration/students.vue
```diff
@@ -48,12 +48,34 @@
...
+					event: 'fabActionClick',
+					arguments: 'new',
+				},
+				{
+					label: $t('pages.administration.students.import'),
+					icon: 'arrow_downward',
+					'icon-source': 'material',
+					event: 'fabActionClick',
+					arguments: 'import',
+				},
```
adrianjost: you could give each action it's own event and add an event listener for each event to have the code more precise

then you could write:
```
@addStudent="addStudent"
@importStudents="importStudents"
```

In this case we want to link to other pages so instead of adding an event with arguments, add an href or to with the according link

PR #746, opened 2020-03-18
SC-3040 | Auto-Logout Warning

BlackGhostHD:
# Issue: [SC-3040](https://ticketsystem.schul-cloud.org/browse/SC-3040)

## Description

-

## Screenshot
Anmerkung 2020-03-18 150841
Anmerkung 2020-03-18 150906
Anmerkung 2020-03-18 151108
Anmerkung 2020-03-18 151052


src/components/organisms/AutoLogoutWarning.vue
```diff
@@ -0,0 +1,193 @@
...
+
+.container {
+	width: 100%;
+	height: 100%;
+	text-align: center;
+
+	.sloth {
+		display: inline-block;
+		width: 35%;
+		min-width: 150px;
```
dominikbruechner: you should always use variables, z.b. size-content-width-min
https://github.com/schul-cloud/styles/blob/master/src/css/default/variables/_breakpoints.scss

src/components/organisms/AutoLogoutWarning.vue
```diff
@@ -0,0 +1,193 @@
...
+
+	.sloth-text {
+		display: inline-block;
+		width: 100%;
+		font-size: var(--text-lg);
+
+		@include breakpoint(tablet) {
+			width: 60%;
+			/* stylelint-disable-next-line sh-waqar/declaration-use-variable */
+			margin-top: 4em;
```
dominikbruechner: please use our mt--xl-3 or mt--xl-4 css classes for spacing
or the regarding css variables https://github.com/schul-cloud/styles/blob/master/src/css/default/variables/_spacing.scss

src/components/base/BaseModal.vue
```diff
@@ -58,6 +58,10 @@ export default {
 			type: String,
 			default: "medium",
 		},
+		onBackDrop: {
```
adrianjost:
```suggestion
		onBackdropClick: {
```
I don't like the naming of this prop, because it does not tell when this callback will be called, only that it has something to do with the backdrop.
- [ ] rename prop

adrianjost: - [ ] Please add a test for this behaviour

src/components/organisms/AutoLogoutWarning.unit.js
```diff
@@ -0,0 +1,118 @@
...
+	it("changing the active property should open and close the modal", async () => {
+		const wrapper = mount(AutoLogoutWarning, {
+			...getMocks(),
+		});
+
+		expect(wrapper.find(".sloth-text").exists()).toBe(false);
+		wrapper.vm.active = true;
+		await wrapper.vm.$nextTick();
+		expect(wrapper.find(".sloth-text").exists()).toBe(true);
+	});
```
adrianjost:
```suggestion
```
This is only testing behaviour of BaseModal component, so it does not belong here and is potentially a duplicate of another test

src/components/organisms/AutoLogoutWarning.unit.js
```diff
@@ -0,0 +1,118 @@
...
+					},
+				},
+			});
+			const wrapper = mount(AutoLogoutWarning, {
+				...mock,
+			});
+
+			wrapper.vm.active = true;
+			wrapper.vm.remainingTimeInSeconds = remainingTime;
+			wrapper.vm.retry = 4; //skip retries for testing
```
adrianjost: this will break as soon as you change the value in the component itself. Maybe you can import it frm the component directly and increment it by 1?
I am not sure what the best way is, I just want to discuss this potential issue.

src/components/organisms/AutoLogoutWarning.unit.js
```diff
@@ -0,0 +1,118 @@
...
+
+			wrapper.find("button").trigger("click");
+			await wrapper.vm.$nextTick();
+			expect(toastStubs.success.mock.calls).toHaveLength(0);
+			expect(toastStubs.error.mock.calls).toHaveLength(0);
+			expect(wrapper.vm.remainingTimeInSeconds).toBe(remainingTime);
+			expect(wrapper.vm.error).toBe(true);
+			expect(wrapper.vm.active).toBe(true);
+
+			wrapper.vm.retry = 4; //skip retries for testing
```
adrianjost: here as well. As you are using this magic constant multiple times, maybe you can at least move it into a variable at the top level of this file?

src/components/organisms/AutoLogoutWarning.vue
```diff
@@ -0,0 +1,193 @@
+<template>
+	<base-modal :active.sync="active" :on-back-drop="extendSession">
+		<template v-slot:header></template>
```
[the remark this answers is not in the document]
adrianjost:
```suggestion
```
I think so too

src/components/organisms/AutoLogoutWarning.unit.js
BlackGhostHD: correctly, the moment the maximum retries are changed in the component, this test would fail. Maybe an additional variable "maxRetries"?

adrianjost: Yes a variable would make it a bit better

src/components/organisms/AutoLogoutWarning.vue
```diff
@@ -0,0 +1,194 @@
+<template>
+	<base-modal :active.sync="active" :on-back-drop="extendSession">
+		<template v-slot:body>
+			<div class="container">
+				<img
+					:src="getImage"
+					class="sloth"
+					:alt="$t('components.organisms.AutoLogoutWarning.image.alt')"
+				/>
```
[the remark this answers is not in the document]
adrianjost: - [ ] add `role="presentation"`

src/components/base/BaseModal.vue
```diff
@@ -70,8 +70,12 @@ export default {
 	},
 	methods: {
 		handleBackgroundClick() {
+			this.onBackdropClick();
```
adrianjost:
```suggestion
			this.$emit("onBackdropClick");
```

src/components/base/BaseModal.vue
```diff
@@ -70,8 +70,12 @@ export default {
...
 			this.close();
 		},
+		onBackdropClick() {
+			this.$emit("onBackdropClick");
+		},
```
adrianjost:
```suggestion
```

src/components/organisms/AutoLogoutWarning.vue
```diff
@@ -0,0 +1,143 @@
...
+					this.$toast.error(
+						this.$t("components.organisms.AutoLogoutWarning.error.401")
+					);
+					break;
+
+				default:
+					break;
+			}
+		},
+	},
```
adrianjost: For the future, an event bus might be a better idea but this is out of scope of this ticket and we need a clear concept before we introduce it, otherwise it may get messy. => so this hack is okay here

src/components/organisms/AutoLogoutWarning.vue
adrianjost: agree

src/components/organisms/AutoLogoutWarning.vue
```diff
@@ -0,0 +1,143 @@
...
+		}
+	}
+
+	.sloth-text {
+		display: inline-block;
+		width: 100%;
+		font-size: var(--text-lg);
+
+		@include breakpoint(tablet) {
+			width: 60%;
```
adrianjost:
```suggestion
			width: 60%;
```
also use a varuable here

src/components/organisms/AutoLogoutWarning.vue
```diff
@@ -0,0 +1,144 @@
...
+				default:
+					break;
+			}
+		},
+	},
+	created(ctx) {
+		this.$store.dispatch("autoLogout/init", null, { root: true });
+	},
+	beforeDestroy() {
+		//this.$store.dispatch("autoLogout/reset");
```
dominikbruechner: oh does this need to come in again?

BlackGhostHD: This is only necessary in a single page application

adrianjost: then we should have removed it or at least add this as a comment. 
Could you do this quickly @BlackGhostHD

BlackGhostHD: Will do this together with https://ticketsystem.schul-cloud.org/browse/SC-3823

PR #747, opened 2020-03-18
feature/SC-3631 student table actions

adrianjost:
# Issue: [SC-3631](https://ticketsystem.schul-cloud.org/browse/SC-3631)

## Description

- added actions to students datatable

## ToDos

- [ ] implement actions
  - [ ] Einverständnis
  - [ ] E-Mail
  - [x] QR (backend fehlt)
  - [x] Löschen (backend fehlt)
- [ ] use language files

## Screenshot

image
image


dominikbruechner:
Maybe we should merge the Delete + i18n as is, and then we can further split up working on the other actions... + we keep the PR smaller

src/serverMiddleware/routes.js
```diff
@@ -7,7 +7,7 @@
 
 module.exports = [
 	`^/administration/datasources?`,
-	// `^/administration/students?`,
+	`^/administration/students/?$`,
```
dominikbruechner:
```suggestion
	// `^/administration/students/?$`,
```

PR #748, opened 2020-03-18
Service Template Reset Action

dominikbruechner:
For resetting the state of an outdated store


src/utils/service-template.js
```diff
@@ -66,6 +73,9 @@ export default function(endpoint) {
 				commit("remove", id);
 				return res;
 			},
+			resetState({ commit }) {
+				commit("reset");
+			},
```
adrianjost: why using an action for that? This does not have any dependencies and we are only mutating the store, so it should be only a mutation.
```suggestion
```

dominikbruechner:
you are right, this was unnecessary

PR #759, opened 2020-03-19
add unit tests to print methods

adrianjost:
# Issue: [SC-3629](https://ticketsystem.schul-cloud.org/browse/SC-3629)

## Description

- add snapshot test setup
- add tests for print mixin methods


jest.config.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
 		"<rootDir>/tests/unit/matchers",
 		"<rootDir>/tests/unit/requireAssertions",
 	],
+	snapshotResolver: "<rootDir>/tests/unit/snapshotResolver",
```
adrianjost: this puts all snapshots into the test directory

tests/unit/__snapshots__/mixins/print.unit.js.snap
```diff
@@ -0,0 +1,298 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
adrianjost: autogenerated snapshot that test runs get compared to

tests/unit/snapshotResolver.js
```diff
@@ -0,0 +1,27 @@
+const path = require("path");
```
adrianjost: resolver for jest snapshots so they are located cleanly inside the test directory and do not clutter our src-code directories. (it is very unlikely that anyone will wan't to read these snapshots anyway)

src/mixins/print.js
```diff
@@ -11,79 +11,80 @@ const printStyles = (optionalStyles) => `<style>
...
+	w.document.write(getPageHtml(printStyles(styles), content));
+	w.document.close();
+	/* eventListener is needed to give the browser some rendering time for the image */
+	w.addEventListener("load", () => {
+		w.focus();
+		w.print();
+		// why is this timeout required?
+		setTimeout(() => {
+			w.close();
+		}, 500);
```
dominikbruechner: there are some browsers which won't start the print dialogue if the timeout is not set. I think I implemented it, more details in the matching commit description (if it's possible to find it)

PR #822, opened 2020-04-01
[SC-3823] Add event bus; refactoring autoLogout

BlackGhostHD:
# Issue: [SC-3823](https://ticketsystem.schul-cloud.org/browse/SC-3823)

## Description
- added event bus plugin
- refactoring autoLogout tzo use event bus


docs/1-Tutorials/HowToUseEventBus.md
```diff
@@ -0,0 +1,58 @@
...
+		doSomethingOnEvent(payload) {
+			...
+		},
+	},
+	eventBus: {
+		"eventName": function (payload) {
+			this.doSomethingOnEvent(payload);
+		},
+	},
+```
```
adrianjost, line 56: this is genious. I like that. but we should decide us for a single syntax and only document one, because otherwise we will have inconsistencies. Let's remove the upper syntax

docs/1-Tutorials/HowToUseEventBus.md
```diff
@@ -0,0 +1,58 @@
...
+
+The above code can be shortened and written as follows:
+
+```js
+	methods: {
+		doSomethingOnEvent(payload) {
+			...
+		},
+	},
+	eventBus: {
```
adrianjost, line 51:
```suggestion
	onEventBus: {
```
Can we rename this to show, that this will only add listeners?
Otherwise I think it might be confusing

stories/.storybook/mockComponents.js
```diff
@@ -2,6 +2,7 @@ import Vue from "vue";
 import Vuex from "vuex";
 
 import "@plugins/global";
+import "@plugins/eventBus";
```
adrianjost, line 5: is it possible to import this directly into the @plugins/global file? Then we can save us an additional import everywhere.
Naming wise, the eventBus is per definition global, so it would fit great into the global file.

docs/1-Tutorials/HowToUseEventBus.md
BlackGhostHD: To better understand this syntax and for the sake of completeness we should not remove the syntax above. I will swap the two syntaxes and declare this syntax as recommended for most cases.

docs/1-Tutorials/HowToUseEventBus.md
BlackGhostHD: Maybe: `eventBusListeners`

docs/1-Tutorials/HowToUseEventBus.md
```diff
@@ -0,0 +1,58 @@
...
+
+In this document you will find how to use the EventBus. But first make sure that the EventBus is the right or better solution for your problem. In most cases, using vuex is a better approach, especially when it comes to state management.
+
+[[toc]]
+
+## How to name your events
+
+Events emitted over the eventBus are global, and can therefore be subscribe to by any component. We need to make sure, that we do not have event identifiers of the same name. This could otherwise cause unpredictable behavior or unexpected bugs.
+
+To ensure this, we use the following naming scheme: <br /> `eventName@component` whereby 'component' is the component that emits the event. If more than one component emits the same event, use: `eventName@global`
```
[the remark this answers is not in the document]
BlackGhostHD: will be commented out for later use

docs/1-Tutorials/HowToUseEventBus.md
```diff
@@ -0,0 +1,76 @@
...
+
+In this document you will find how to use the EventBus. But first make sure that the EventBus is the right or better solution for your problem. In most cases, using vuex is a better approach, especially when it comes to state management. The most common use case for the EventBus is to trigger a function in another component.
+
+[[toc]]
+
+<!-- ## How to name your events
+
+Events emitted over the eventBus are global, and can therefore be subscribe to by any component. We need to make sure, that we do not have event identifiers of the same name. This could otherwise cause unpredictable behavior or unexpected bugs.
+
+To ensure this, we use the following naming scheme: <br /> -->
```
adrianjost, line 11:
```suggestion
```
I have copied this text to the follow up ticket, so we can remove it here.

docs/1-Tutorials/HowToUseEventBus.md
```diff
@@ -0,0 +1,71 @@
...
+		...
+	},
+},
+```
+
+In most cases you probably want to call the listener in the created hook. Additionally, don't forget to unsubscribe unused listeners.
+
+## How to unsubscribe from events
+
+Don't forget to unsubscribe unused listeners ;)
```
adrianjost, line 67:
```suggestion
Don't forget to unsubscribe unused listeners, if attached yourself manually using `$on` ;)
```

src/components/organisms/AutoLogoutWarning.vue
```diff
@@ -94,15 +99,9 @@ export default {
...
-	},
-	beforeDestroy() {
-		//this.$store.dispatch("autoLogout/reset");
-	},
-	methods: {
-		extendSession() {
-			this.$store.dispatch("autoLogout/extendSession");
+	eventBus: {
+		"showToast@autologout": function (value) {
+			this.showToast(value);
```
adrianjost, line 104:
```suggestion
		"showToast@autologout": (value) => {
			this.showToast(value);
```
geht das nicht auch als arrow function? dann ist klarer, dass der this kontext von der Komponente weiterhin genutzt wird.

BlackGhostHD: This is not possible, otherwise `this` is undefined

adrianjost: but it is an arrow function 🤔  But ok, I trust you :)

PR #847, opened 2020-04-06
SC-3910 | show floating action button only if user has permission to create/import student/teacher

BlackGhostHD:
# Issue: [SC-3910](https://ticketsystem.schul-cloud.org/browse/SC-3910)

## Description

- show floating action button only if user has permission to create/import student/teacher


src/pages/administration/students/index.vue
```diff
@@ -65,6 +65,7 @@
 			v-if="schoolInternallyManaged"
 			position="bottom-right"
 			:show-label="true"
+			permission="STUDENT_CREATE"
```
adrianjost, line 68: Please import the UserHasPermission mixin here, and then adjust the v-if

PR #889, opened 2020-04-16
feature/dependency updates 2

adrianjost:
# Issue

Outdated puppeteer and jest versions used

## Description

- updated jest
- updated puppeteer


adrianjost:
machen leider doch beide probleme

PR #901, opened 2020-04-21
SC-438 - persist filters, pagination and more for session

BlackGhostHD:
# Issue: [SC-438](https://ticketsystem.schul-cloud.org/browse/SC-438)

## Description

(https://github.com/schul-cloud/schulcloud-client/pull/1773)


src/store/uiState.js
```diff
@@ -0,0 +1,74 @@
...
+					return state[key][identifier];
+				} else {
+					return state[key];
+				}
+			} else {
+				throw new SyntaxError("Key is missing!");
+			}
+		} else {
+			throw new SyntaxError("uiState not initialize");
+		}
```
adrianjost, line 29: You could escape the condition hell here, by using early error handling.
```suggestion
		if (!initialize) {
			throw new SyntaxError("uiState not initialize");
		}
		if (!key) {
			throw new SyntaxError("Key is missing!");
		}
		return identifier
			? state[key][identifier]
			: state[key]
```

src/store/uiState.js
```diff
@@ -0,0 +1,74 @@
...
+					localStorage.setItem(localStorageKey, JSON.stringify(state));
+				} else {
+					throw new Error("Overwriting the default state is not permitted!");
+				}
+			} else {
+				throw new SyntaxError("Key or/and object is missing!");
+			}
+		} else {
+			throw new SyntaxError("uiState not initialize");
+		}
```
adrianjost, line 62: same thing with early error handling here.
In my experience, it is most often cleaner to handle errors first and have the actual logic afterwards, cleanly in a safe environment where all your conditions are already met.

BlackGhostHD:
@adrianjost First of all, thank you for taking the time for this comprehensive answer. Maybe I have overengineered the topic a bit, but here's my thought why. 
Sure, you could use mutations like 
```
this.$store.commit("uiState/alterState", {
		key: "pagination",
		identifier,
		object: { page, limit },
	});
```
to set the state without any plugin or extras. Nevertheless I wanted to provide an easy to use function set to make it fast and clear what is set or returned. Similar to a facade pattern.

This is also the reason why I decided to separate filter and pagination.
Also it should be relatively easy to store new stuff. To do this you just have to add it to the defaultState and add the getters and setters in the plugin. 

Moving the initialize method to the getDefaultState method is a good idea.

BlackGhostHD:
Yes,`this.$uiState.set("filter", "pages.administration.students.index", { query });` is a better syntax and probably more understandable

src/pages/administration/datasources/index.vue
```diff
@@ -390,11 +391,10 @@ export default {
...
 			this.limit = limit;
-			// save user settings in localStorage
-			localStorage.setItem(
-				"pages.administration.datasources.index.itemsPerPage",
-				limit
-			);
+			// save user settings in uiState
+			this.$uiState.setPagination("pages.administration.datasources.index", {
+				itemsPerPage: limit,
+			});
```
adrianjost, line 397: you forgot to update this one

src/plugins/uiState.js
```diff
@@ -0,0 +1,27 @@
+let store;
+
+const set = (key, identifier, value) => {
+	store?.commit("uiState/set", {
+		key,
+		identifier,
+		object: value,
```
adrianjost, line 7: vllt, statt `value` (impliziert einfache daten) und `object` (verwirrend zu Object) einfach `data` nutzen?
```suggestion
const set = (key, identifier, data) => {
	store?.commit("uiState/set", {
		key,
		identifier,
		data,
```

src/store/uiState.js
```diff
@@ -0,0 +1,114 @@
...
+		if (uiState.version == version) {
+			Object.assign(defaultState, uiState);
+		} else {
+			uiState.version = version;
+		}
+	}
+	initialize = true;
+
+	return defaultState;
+};
```
adrianjost, line 25:
```suggestion
const getDefaultState = () => {
	const defaultState = {
		pagination: {},
		filter: {},
		version,
	};
	if (localStorage.getItem(localStorageKey)) {
		const uiState = JSON.parse(localStorage.getItem(localStorageKey));
		if (uiState.version == version) {
			Object.assign(defaultState, uiState);
		} else {
			uiState.version = version;
		}
	}
	initialize = true;

	return defaultState;
};
```
move the defaultState definition into the method to create and use a new object for defaultState each time. Otherwise there may be conflicts when resetting the store, because you had modified your defaultState.

src/store/uiState.js
```diff
@@ -0,0 +1,114 @@
...
+		case "pagination":
+			return handlePagination(methode, value);
+
+		case "filter":
+			return handleFilter(methode, value);
+
+		default:
+			return value;
+	}
+};
```
adrianjost, line 44: I still think you can build something wothout explicitly describing all the avaialble types explicitly. I will post you an example snipped what I mean here soon.

src/store/uiState.js
```diff
@@ -0,0 +1,114 @@
...
+		if (!key) throw new SyntaxError("Key is missing!");
+
+		const value = identifier ? state[key][identifier] : state[key];
+		return createPattern(key, "get", value);
+	},
+};
+
+export const mutations = {
+	set(state, { key, identifier, object }) {
+		if (!initialize) throw new Error("uiState not initialize");
```
adrianjost, line 83: this will never happen, because all the code for initializing the store is synchronous and will be executed first.

src/store/uiState.js
```diff
@@ -0,0 +1,114 @@
...
+		if (!key || !object) throw new SyntaxError("Key or/and object is missing!");
```
adrianjost, line 84: I like that you use SyntaxError 👌

src/store/uiState.js
BlackGhostHD: I wanted to eliminate the possibility of any future bugs. That's why I chose this strict definition and structure of localstorage.I wanted to avoid that you have to write  e.g.: `this.$uiState.get("filter", "pages.administration.students.index")?.query` instead of `this.$uiState.get("filter", "pages.administration.students.index")`.  If you forget the optional chaining operator, bugs will occur if the localstorge is empty. This can, but does not have to occur during programming. The same applies to the query of limit or page in pagination. Here the key should always be defined, even if the value is undefined.
I'm always happy to learn new things, so I would be very grateful for a code sample from you.

adrianjost: Ahhh I get it know. With this in mind your solution seems fine. 👍

src/store/uiState.js
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return value;
+};
+
+const handleFilter = (methode, value) => {
+	if (methode === "get") {
+		return value?.query || {};
+	}
+	return value;
+};
```
adrianjost, line 67: I would like to see those definitions above the `createPattern`method. It's best practice to define functions before usage because it is mandatory in most languages (js is obviously not one of those) and it keeps code more readable if you follow this convention everywhere. It also prevents you from accidentally building cycling function dependencies.

src/store/uiState.js
```diff
@@ -0,0 +1,110 @@
...
+				state[key][identifier] = mergeDeep(state[key][identifier], value);
+			}
+		} else {
+			if (key in getDefaultState())
+				throw new Error("Overwriting the default state is not permitted!");
+			if (!state[key]) {
+				state[key] = value;
+			} else {
+				state[key] = mergeDeep(state[key], value);
+			}
```
adrianjost, line 97:
```suggestion
			state[key] = mergeDeep(state[key] || {}, value);
```
This should work as well and may be a fit shorter

src/store/uiState.js
BlackGhostHD: done

PR #912, opened 2020-04-23
SC-3664 - Feedback CSV Import

luizzaly:
# Issue: [SC-](https://ticketsystem.schul-cloud.org/browse/SC-3664)

## Description 

- New mixin in order to render toasts from the legacy client in Nuxt

-  needs to be merged with `https://github.com/schul-cloud/schulcloud-client/pull/1777`


src/serverMiddleware/proxy.js
```diff
@@ -30,8 +30,9 @@ console.log(
 console.log("PROXY TARGET:", proxyOptions.target);
 
 const vueRoutes = require("./routes.js");
-const isNuxtRoute = (url) =>
-	vueRoutes.some((regexString) => !!new RegExp(regexString).exec(url));
+const isNuxtRoute = (url) => {
+	return vueRoutes.some((regexString) => !!new RegExp(regexString).exec(url));
+};
```
adrianjost, line 35:
```suggestion
const isNuxtRoute = (url) => 
	vueRoutes.some((regexString) => !!new RegExp(regexString).exec(url));
```
hier hat sich inhaltlich nichts geändert, dann kann auch der code so bleiben wie er ist :D

src/mixins/toasts.js
```diff
@@ -0,0 +1,22 @@
+export default {
```
adrianjost, line 1: I would like to rename this mixin to be more specific what it does.
I could imagine something like `toastsFromQueryString.js` as a filename.
Please also adjust the testfile-name and the import accordingly

PR #915, opened 2020-04-24
SC-4085 - Admin-Table-SideScrolling

ABialoskorski:
# Issue: [SC-4085](https://ticketsystem.schul-cloud.org/browse/SC-4085)
## Description

I had to add overflow on content tag, to fit all content when the browser window is decreasing

## Screenshot

image


ABialoskorski:

src/layouts/loggedInFull.vue
```diff
@@ -24,6 +24,7 @@ export default {
 	width: inherit;
 	min-height: 100vh;
 	padding: 0 var(--space-lg);
+	overflow-x: auto;
```
adrianjost, line 27: I don't think this should be in the layout. If we add it to the layout and the user scrolls horizontally, he loses context  because he scrolls away all normal content. I think it is better to add this to the DataTable

PR #925, opened 2020-04-28
SC-3888 - Sticky-Header-Data-Tables

ABialoskorski:
# Issue: [SC-3888](https://ticketsystem.schul-cloud.org/browse/SC-3888)
## Description

Added position sticky on thead, but it works only when the list of items is too big to fit in table
 ## Screenshot when items would not fit in table
image


src/serverMiddleware/routes.js
```diff
@@ -7,10 +7,10 @@
...
 module.exports = [
 	`^/administration/datasources?`,
-	// `^/administration/students/?$`,
-	// `^/administration/students/new/?$`,
-	// `^/administration/teachers/?$`,
-	// `^/administration/teachers/new/?$`,
+	`^/administration/students/?$`,
+	`^/administration/students/new/?$`,
+	`^/administration/teachers/?$`,
+	`^/administration/teachers/new/?$`,
```
adrianjost, line 13: I think they should still be disabled. @CordlessWool commented them out just two days ago.
```suggestion
	// `^/administration/students/?$`,
	// `^/administration/students/new/?$`,
	// `^/administration/teachers/?$`,
	// `^/administration/teachers/new/?$`,
```

ABialoskorski: pushed solution to branch about that, I don't know why it's still showing in PR

PR #976, opened 2020-05-13
remove filter from deepMerge

pauludobeneke:
# Description

Base links to copy
- https://ticketsystem.schul-cloud.org/browse/SC-4294

## Approval for review

- [ ] All points were discussed with the ticket creator, support-team or product owner. The code upholds all quality guidelines from the PR-template.

### Link to Definiton of Done

### DIFF
--- a/src/store/uiState.js
+++ b/src/store/uiState.js
@@ -84,7 +84,11 @@ export const mutations = {
 		const value = createPattern(key, "set", object);
 
 		if (identifier && state[key]) {
-			state[key][identifier] = mergeDeep(state[key][identifier] || {}, value);
+			if (key === "filter" && state[key][identifier]) {
+				state[key][identifier]["query"] = object["query"];
+			} else {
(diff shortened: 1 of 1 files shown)


src/store/uiState.js
```diff
@@ -84,7 +88,11 @@ export const mutations = {
 		const value = createPattern(key, "set", object);
 
 		if (identifier && state[key]) {
-			state[key][identifier] = mergeDeep(state[key][identifier] || {}, value);
+			if ("overWrite" in value && state[key][identifier]) {
+				state[key][identifier]["query"] = object["query"];
```
BlackGhostHD, line 92: Good approach, but then you can only use this for querys. Maybe we can make it even more reusable.
Maybe something like: 
```
value = {
    overwrite: true/false
    data: {...}  <-- what was previously value
}

...

if (value.overwrite && state[key][identifier]) {
    state[key][identifier] = value.data;
}else{
    ... deepMerge
}
```

PR #1115, opened 2020-07-07
SC-5265 - lernstore-arrow-back-window

ABialoskorski:
# Description
changed reuirements, right now it should be full size window with arrow to close it
please check also client PR https://github.com/schul-cloud/schulcloud-client/pull/1988

- https://github.com/schul-cloud/schulcloud-client/pull/1988
- https://ticketsystem.schul-cloud.org/browse/SC-5265

## Screenshots of UI changes
image


src/pages/content/index.vue
```diff
@@ -101,6 +110,9 @@ export default {
 			}
 			return query;
 		},
+		isInline() {
+			return window.location.search.includes("inline=1");
```
BlackGhostHD, line 114: use $route object instead of window;  this.$route.query.inline

src/pages/content/index.vue
```diff
@@ -63,6 +70,8 @@ import infiniteScrolling from "@mixins/infiniteScrolling";
 import BaseGrid from "@components/base/BaseGrid";
 import ContentEduSharingFooter from "@components/molecules/ContentEduSharingFooter";
 
+import Theme from "@theme/config";
```
BlackGhostHD, line 73: You do not need to do that anymore, it is already imported into the global.js. You can directly use $theme to get for example: the short_name

BlackGhostHD, line 73: You do not need to do that anymore, it is already imported into the global.js. You can directly use $theme to get for example: the short_name

src/pages/content/index.vue
```diff
@@ -169,11 +181,23 @@ export default {
...
 	head() {
-		return {
-			title: "LernStore",
-		};
+		if (this.isInline) {
+			return {
+				title:
+					this.$t("pages.content.page.window.title") +
+					Theme.name +
+					this.$t("pages.content.page.window.title_2"),
```
BlackGhostHD, line 194: https://kazupon.github.io/vue-i18n/guide/formatting.html#named-formatting

src/pages/content/index.vue
BlackGhostHD, line 114: use $route object instead of window;  this.$route.query.inline

src/pages/content/index.vue
```diff
@@ -1,5 +1,12 @@
 <template>
 	<section>
+		<base-icon
```
BlackGhostHD, line 3: I think, but I'm not sure, for the sake of Accessibility a button would be better here

src/pages/content/index.vue
```diff
@@ -169,11 +181,23 @@ export default {
...
 		},
+		goBack() {
+			window.close();
+		},
 	},
 	head() {
-		return {
-			title: "LernStore",
-		};
+		if (this.isInline) {
```
BlackGhostHD, line 189: You could use a ternary operator here

src/pages/content/index.vue
ABialoskorski: done ;)

src/pages/content/index.vue
ABialoskorski: done ;)

src/pages/content/index.vue
ABialoskorski: nice ;)

src/pages/content/index.vue
ABialoskorski: done ;)

src/pages/content/index.vue
ABialoskorski: added ;)

src/pages/content/index.vue
```diff
@@ -1,5 +1,8 @@
 <template>
 	<section>
+		<base-button v-if="isInline" design="none" type="button" @click="goBack">
+			<base-icon class="arrow__back" source="material" icon="arrow_back" />
+		</base-button>
```
BlackGhostHD, line 5:
```suggestion
		<base-button
			v-if="isInline"
			design="text icon"
			type="button"
			class="arrow__back"
			@click="goBack"
		>
			<base-icon source="material" icon="arrow_back" />
		</base-button>
```
design="none" will not have any effects, e.g: thier will be not cursor: pointer

ABialoskorski: done ;)

ABialoskorski:
> Unfortunately doesn't work with multiple screens.

fixed on client PR

PR #1120, opened 2020-07-08
SC-5492 - detailview-from-course

ABialoskorski:
# Description
Add inline=1 to lernstore link always when we are in 'new window mode', when You are normally in lern-store inline=0

- https://ticketsystem.schul-cloud.org/browse/SC-5492

### DIFF
--- a/src/pages/content/index.vue
+++ b/src/pages/content/index.vue
@@ -125,10 +125,15 @@ export default {
 				return;
 			}
 			this.$options.debounce = setInterval(() => {
+				let inline = 0;
 				clearInterval(this.$options.debounce);
+				if (window.location.search.includes("inline=1")){
+					inline = 1;
+				}
 				this.$router.push({
 					query: {
 						q: this.searchQuery,
+						inline: inline
 					},
 				});
 			}, 500);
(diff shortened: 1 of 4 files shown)


src/pages/content/index.vue
```diff
@@ -125,10 +125,15 @@ export default {
...
 			this.$options.debounce = setInterval(() => {
+				let inline = 0;
 				clearInterval(this.$options.debounce);
+				if (window.location.search.includes("inline=1")){
+					inline = 1;
+				}
 				this.$router.push({
 					query: {
 						q: this.searchQuery,
+						inline: inline
```
BlackGhostHD: Good approach 👍, but I would suggest that you just take all old query elements and then overwrite them if necessary. e.g:
```
query: {
	...this.$route.query,
	q: this.searchQuery,
},
```

BlackGhostHD: This must also happen in line 122. Otherwise, when deleting the current search term, all other query parameters are also deleted

ABialoskorski: nice ;)

PR #1128, opened 2020-07-09
SC-5517 - hover-bug

ABialoskorski:
# Description
need to fix color on hover, there is always green on hover, regardless Success or not

- https://ticketsystem.schul-cloud.org/browse/SC-5517

### DIFF
--- a/src/components/molecules/ModalFooterConfirm.vue
+++ b/src/components/molecules/ModalFooterConfirm.vue
@@ -1,5 +1,5 @@
 <template>
-	<div :style="{ backgroundColor: backgroundcolor }" class="footer-button">
+	<div :style="{ backgroundColor: backgroundcolor }" class="footer-button" :class="isSuccess ? '' : 'error'">
 		<button class="btn-confirm" v-on="$listeners">
 			Ok
 		</button>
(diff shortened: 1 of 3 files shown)


src/components/molecules/ModalFooterConfirm.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-	<div :style="{ backgroundColor: backgroundcolor }" class="footer-button">
+	<div
+		:style="{ backgroundColor: backgroundcolor }"
+		class="footer-button"
+		:class="isSuccess ? '' : 'error'"
```
mwolowyk-cap, line 5: Since we are trying to do this modal universaly usable. Probably it makes sence to set the type of the modal like: 'success', 'error', 'warning', 'info' .... 
Some enumeration?

ABialoskorski: umm, but we don't have right now anything like warning or info. It will be added in future right?

ABialoskorski: for me, here I'm jusr fixing bug, making NotifiactionModal even more universal is diffrent story which will show up when needed I think

mwolowyk-cap: ok.

ABialoskorski:
I will probably rebuild it with base button

PR #1145, opened 2020-07-15
SC-4275 - SC International Nuxt Theming

ABialoskorski:
# Description
added new theme and adjusted colors for it

- https://github.com/schul-cloud/schulcloud-client/pull/2004
- https://ticketsystem.schul-cloud.org/browse/SC-4275

## Screenshots of UI changes
image


src/themes/int/config.js
```diff
@@ -0,0 +1,4 @@
+export default {
+	name: "HPI Schul-Cloud International",
+	short_name: "Internationale Schul-Cloud",
```
dominikbruechner, line 3: somehow possible to use i18n keys here?

ABialoskorski: these are brand names, so I think the names could stay there, but I will ask Mona

src/themes/int/styles/variables/_colors.scss
```diff
@@ -0,0 +1,14 @@
...
+	/* main colors */
+	--color-primary: #064963;
+	--color-primary-light: #417491;
+	--color-primary-dark: #002239;
+	--color-secondary: #d9174c;
+	--color-secondary-light: #f15574;
+	--color-secondary-dark: #a31139;
+	--color-accent: #74c599;
+	--color-gray-dark: #17212a;
+	--color-gray-light: rgba(23, 33, 42, 0.1);
```
dominikbruechner, line 12: one more thing, here you might want to use
	--color-primary-light: color-mod(var(--color-primary) tint(#{$brighten}));

with $brighten: 10%;

ABialoskorski: shouldn't I use color instead of color-mod? webstorm complains about color-mod

ABialoskorski: I did commit https://github.com/schul-cloud/nuxt-client/pull/1145/commits/90af50a888a3f4e3483e9aeb4a75667a452d1e67, please check ;)

dominikbruechner: just changed real quick the reference color (primary in my example was wrong) otherwise fine

src/themes/int/config.js
dominikbruechner: I think the short_name is prone to problems (out of a grammatical perspective), where was it defined?

ABialoskorski: nowhere, tbh, I will check with Mona

dominikbruechner:
fixed the focus issue

ABialoskorski:
> Screen Shot 2020-07-16 at 14 34 41
> 
> Still an issue with focused menu item

fixed, webpack problem, there were no issue ;)

ABialoskorski:

dominikbruechner:
I have just realized when you posted the screenshots: that were all legacy client screenshots.
the Nuxt client is e.g. active for /news/new, or /imprint, or some other things

please double check

 I have updated the screenshot as well

ABialoskorski:
> I have just realized when you posted the screenshots: that were all legacy client screenshots.
> the Nuxt client is e.g. active for /news/new, or /imprint, or some other things
> 
> please double check
> 
> I have updated the screenshot as well

I have tried that in new lernstore which comes from nuxt, screenshots are just copy paste from client, but yeah I tried that in nuxt ;)

src/themes/int/components/legacy/TheSidebar.vue
```diff
@@ -0,0 +1,259 @@
...
+				}
+
+				&.active .list-content {
+					color: var(--color-sidebar-menu-item-active);
+				}
+			}
+		}
+	}
+}
+</style>
```
dominikbruechner, line 259: Why did you duplicate TheSidebar component?
Please revert, if you want to do style changes, you can override certain stylings - but that's a huge code redundancy. Please advise about the exact reasons

src/themes/int/components/legacy/TheFooter.vue
```diff
@@ -0,0 +1,152 @@
...
+	border: none;
+	&:focus,
+	&:hover {
+		text-decoration: underline;
+	}
+	&:visited {
+		color: var(--color-primary);
+	}
+}
+</style>
```
dominikbruechner, line 152: same as for TheSidebar, code redundancy, is it necessary here?

dominikbruechner:
> > I have just realized when you posted the screenshots: that were all legacy client screenshots.
> > the Nuxt client is e.g. active for /news/new, or /imprint, or some other things
> > please double check
> > I have updated the screenshot as well
> 
> @dominikbruechner I have tried that in new lernstore which comes from nuxt, screenshots are just copy paste from client, but yeah I tried that in nuxt ;)

ah alright
then we only have theSidebar/footer thingy open

src/themes/int/components/legacy/TheFooter.vue
ABialoskorski: In Footer I saw that colors of footer-links should be changed, and in sidebar padding of logo should be changed, if there is a better solution, maybe just scss classes will do the same work, then yeah we should delete 90% of those 2 components. I'm going on vacation till tuesday, if that ticket is necessary I told in team Freya that someone can pick it up ;)

ABialoskorski: is there an option to delete 90% of code, or it should stay as it is? @dominikbruechner

BlackGhostHD: I just saw that for each instance we duplicate the footer comopnente to 98% and thus we have 4 times nearly identical code, this has nothing to do with object oriented programming :( I would suggest to build a "base" footer componente and a special component, which then e.g. only passes the links. This will be overwritten for each instance and if necessary, individual styles can be overwritten. So we have the 98% part of the code only once.

We can do this in the course of this ticket or create a follow-up ticket for it, which will be processed directly after (or before this is merged here) if possible.

ABialoskorski: I would be up to create follow up ticket, to be sure that every instance had time to be checked, I think adding international theme is needed a little bit quickly.

ABialoskorski: I would say that this new ticket is 5/8 SP. In some components there are also html changes and so on

src/themes/int/styles/variables/_colors.scss
```diff
@@ -0,0 +1,16 @@
+$brighten: 90%;
+
+// extends the default theme
+:root {
+	/* main colors */
+	--color-primary: #064963;
+	--color-primary-light: #417491;
```
mwolowyk-cap, line 7: It is better to use

```	--color-primary-light: color-mod(var(--color-primary) tint(#{$brighten}));
	--color-primary-dark: color-mod(var(--color-primary) shade(#{$darken}));
	--color-secondary-light: color-mod(var(--color-secondary) tint(#{$brighten}));
	--color-secondary-dark: color-mod(var(--color-secondary) shade(#{$darken}));```

ABialoskorski: with brighten 15% I should add one more brighten because I use already 10%. I got hex colors from here https://github.com/hpi-schul-cloud/styles/blob/master/src/css/default/variables/_colors.scss.
and hex colors are used in other themes colors.scss, so I think I can stick with that

ABialoskorski:

src/themes/int/styles/variables/_colors.scss
dominikbruechner: fine @ABialoskorski

PR #1262, opened 2020-08-17
SC-3788 - matrix messenger integration

maxklenk:
# Description

- [x] Configure feature flag
- [x] Configure embed endpoint
- [x] Integrate Messenger in all loggedIn pages
- [x] Terminate Messenger on loggedOut pages
- [x] Pass current users language to the messenger
- [x] Configure CORS
- [x] Test Coverage

- https://ticketsystem.schul-cloud.org/browse/SC-3788

## Changes
- add Messenger component
- include component in layout views

## Datasecurity <sub><sup>details [on Confluence](https://docs.schul-cloud.org/x/2S3GBg)</sup></sub>
- the route `schools/find` has to be opened on the server to be accessible from the nuxt client

## Deployment
- add messenger environment variables that already exist in the client:
```
FEATURE_MATRIX_MESSENGER_ENABLED
FEATURE_MESSENGER_SCHOOL_SETTINGS_VISIBLE
FEATURE_MESSENGER_SCHOOL_ROOM_ENABLED
MATRIX_MESSENGER_EMBED_URI
MATRIX_MESSENGER_HOMESERVER_URI
```

## New Repos, NPM packages or vendor scripts
- Messenger resource are loaded when required

## Screenshots of UI changes
Screenshot 2020-08-17 at 3 31 13 PM

## Approval for review

### Link to Definition of Done


src/components/organisms/Messenger/Messenger.unit.js
```diff
@@ -0,0 +1,103 @@
...
+	});
+
+	it("do not initialize if feature is not enabled for school", async () => {
+		process.env.FEATURE_MATRIX_MESSENGER_ENABLED = "true";
+		school.features = [];
+
+		mount(Messenger, {
+			...createComponentMocks({ i18n: true, user: true, store: mockStores }),
+		});
+		await wait(1000);
```
maxklenk, line 60: I would like to get rid of those wait calls. Is there another way to wait for the promises inside the component. The promises are mocked so they should be return in the next cycle/tick.

BlackGhostHD, line 60:
```suggestion
		const wrapper = mount(Messenger, {
			...createComponentMocks({ i18n: true, user: true, store: mockStores }),
		});
		await wrapper.vm.$nextTick();
```

BlackGhostHD: You can try my suggestion, might solve your problem

src/layouts/loggedInFull.vue
```diff
@@ -3,15 +3,18 @@
...
 </template>
 
 <script>
 import LegacyLoggedIn from "@layouts/legacyLoggedIn";
+import Messenger from "../components/organisms/Messenger/Messenger";
 
 export default {
 	components: {
 		LegacyLoggedIn,
+		Messenger,
```
BlackGhostHD, line 17: I missed that in our conversation yesterday; you don't need to include the messenger component here, because the layout uses LegacyLoggedIn and the component has already been added there (otherwise the component is added twice)

src/components/organisms/Messenger/Messenger.unit.js
maxklenk: Awesome, I have to wait for each promise, so it requires multiple ticks, but works great.

src/components/organisms/Messenger/Messenger.vue
```diff
@@ -0,0 +1,170 @@
...
+				return Promise.resolve(false);
+			}
+
+			const query = {
+				query: {
+					_id: this.$user.schoolId,
+				},
+			};
+			return this.$store
+				.dispatch("schools/find", query)
```
BlackGhostHD, line 43: This API endpoint is not yet enabled for Nuxt in production. You may have to find another way or create a ticket to have this endpoint enabled.

BlackGhostHD (dismissed):
Looks good to me. 👍  Maybe @dominikbruechner can review this as well.

janrenz:
Why is this not merged yet? I prose a shared review session if this otherwise is hard to review

pkleybolte (changes requested):

src/components/organisms/Messenger/Messenger.vue
```diff
@@ -0,0 +1,125 @@
...
+	methods: {
+		loadMessengerEmbed() {
+			// load javascript
+			const riotScript = document.createElement("script");
+			riotScript.src = `${this.matrixAssetDomain}/embed.js`;
+			riotScript.type = "text/javascript";
+			document.head.appendChild(riotScript);
+		},
+		extractRoomTypeAndIdFromPath(path) {
+			const matches = RegExp("/(course|team)s/([0-9a-f]{24})").exec(path);
```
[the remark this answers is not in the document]
CeEv: const path = 'https://api.schul-cloud.de/courses/5d765d332d0323001a31c82222222222'; will match and it is cut the result after the second 2 at the end.

src/components/organisms/Messenger/Messenger.vue
```diff
@@ -0,0 +1,125 @@
...
+				riotConfig: "/riot_config.json",
+				indexeddbWorkerScript: "/indexeddb-worker.js",
+				assetDomain: `${this.matrixAssetDomain}/`,
+				language: this.userLanguage || "de",
+				forceToggled: true,
+			};
+
+			// force the selection of a specific room
+			if (matrixRoomId) {
+				options.roomId = matrixRoomId;
```
CeEv, line 105: If extractRoomTypeAndIdFromPath failed, then matrixRoomId and options.roomId = null;
Must we really execute some more stuff after this? 
Or is this fine and the null is what we want to know that we not show this room at this side.

src/store/auth.js
```diff
@@ -31,6 +31,19 @@ export const actions = {
 	async logout({ commit }) {
 		this.$cookies.remove("jwt");
 		localStorage.clear();
+		// Delete matrix messenger indexedDB databases
+		if (indexedDB) {
```
CeEv, line 35: Ok fine and is it is not defined what should be execute? Any error/ warning, post to Sentry?

src/store/messenger.js
```diff
@@ -0,0 +1,40 @@
+export const actions = {
+	async loadMessengerToken({ commit }) {
+		try {
+			const data = await this.$axios.$post("/messengerToken");
+			commit("setMessengerToken", data);
+		} catch (error) {
+			console.log(error);
+		}
+		// TODO what is supposed to happen on error?
```
CeEv, line 9: True todo, error pipline? Console.log is not a option. 

At the moment it is server side rendering. If it is possible that this error it throw by rending? Than we has the trouble with slow sync actions in production. By handlebars in old client the console.log execute on server side, not on client side.

src/store/messenger.unit.js
```diff
@@ -0,0 +1,59 @@
...
+		describe("serverName", () => {
+			it("returns null if no session present", () => {
+				expect(getters.serverName({})).toBeNull();
+			});
+			it("returns server name", () => {
+				expect(
+					getters.serverName({
+						session: { userId: "dummy-user:dummy-server.name" },
+					})
+				).toBe("dummy-server.name");
```
CeEv, line 14: Please add somethink like  { userId: "dummy-server.name" }, and  { userId: "dummy-user:dummy-server" }, as false input tests

src/store/messenger.unit.js
```diff
@@ -0,0 +1,59 @@
...
+						receivedUrl = url;
+						return Promise.resolve(dummyReturnObject);
+					},
+				};
+				const spyCommit = jest.fn();
+
+				await actions.loadMessengerToken({ commit: spyCommit });
+
+				expect(receivedUrl).toBe("/messengerToken");
+				expect(spyCommit.mock.calls).toHaveLength(1);
```
CeEv, line 34: By executing more then one expect in one it, i am happy if you can add what every expect should test. As feedback for test if they are failed.

expect(spyCommit.mock.calls, 'Should has lengh one, because ....').toHaveLength(1);

src/store/messenger.unit.js
```diff
@@ -0,0 +1,59 @@
...
+			it("triggers commit", async () => {
+				const dummyReturnObject = { someProperty: "dummy return message" };
+				let receivedUrl;
+				actions.$axios = {
+					$post: (url) => {
+						receivedUrl = url;
+						return Promise.resolve(dummyReturnObject);
+					},
+				};
+				const spyCommit = jest.fn();
```
CeEv, line 29: I know the syntax
const mocker= jest.fn();
const snapshot = new mocker();

How it work in this way? Can you please validate what is the right way to use it.

CeEv (changes requested):

src/components/organisms/Messenger/Messenger.unit.js
```diff
@@ -0,0 +1,147 @@
...
+	auth: {
+		state: () => ({
+			school: { features: ["messenger"] },
+		}),
+	},
+};
+
+describe("@components/organism/Messenger", () => {
+	beforeEach(() => {
+		process.env.FEATURE_MATRIX_MESSENGER_ENABLED = undefined;
```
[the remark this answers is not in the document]
pkleybolte: You can't have more than one beforeEach. So We can't generate an extra one for the toggles. I guess we will have to deal with this.
There should no be to many cases like this though, because most feature toggles should be server side only.

src/components/organisms/Messenger/Messenger.unit.js
```diff
@@ -0,0 +1,147 @@
...
+		});
+		await wrapper.vm.$nextTick(); // isActive
+		await wrapper.vm.$nextTick(); // getSchool
+
+		expect(window.Matrix).toBeUndefined();
+	});
+
+	it("init messenger from localStorage", async () => {
+		process.env.FEATURE_MATRIX_MESSENGER_ENABLED = "true";
+		window.localStorage.setItem("mx_hs_url", "domain");
```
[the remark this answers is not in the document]
pkleybolte: I'll remove them again at the end of the test just in case.

src/components/organisms/Messenger/Messenger.unit.js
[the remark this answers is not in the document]
tofixx: you may integrate commons.configuration to solve this...

src/components/organisms/Messenger/Messenger.vue
```diff
@@ -0,0 +1,125 @@
...
+</template>
+
+<script>
+export default {
+	data() {
+		return {};
+	},
+	computed: {
+		matrixFeatureFlag() {
+			return process.env.FEATURE_MATRIX_MESSENGER_ENABLED === "true";
```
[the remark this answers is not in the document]
pkleybolte: There is no config file like in the server in Nuxt.
And compile time they are written from the nuxt.config.js into that var. The nuxt.config.js is not available at runtime or in the tests though.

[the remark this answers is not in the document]
tofixx: sure, however the config must match client/server because of that i would load the values not as configuration but upon login as part of school/user view permissions/setting instead. for now this may stay while we could add schema validation using commons.configutration and export the whole configuration afterwards to be used...

src/store/messenger.unit.js
pkleybolte: I used the syntax for creating a mock function using jest only.
The are a couple of other libs for mocks. The syntax you mentioned is used by at least one other lib.

src/store/messenger.unit.js
pkleybolte: expect() does not provide the possibility to add a custom error message.

src/components/organisms/Messenger/Messenger.unit.js
```diff
@@ -0,0 +1,147 @@
...
+
+	it("init messenger from api", async () => {
+		process.env.FEATURE_MATRIX_MESSENGER_ENABLED = "true";
+
+		const wrapper = mount(Messenger, {
+			...createComponentMocks({ i18n: true, store: mockStores }),
+		});
+		await wrapper.vm.$nextTick(); // isActive
+		await wrapper.vm.$nextTick(); // getSchool
+		await wrapper.vm.$nextTick(); // getSession
```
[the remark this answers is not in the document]
pkleybolte: nextTick not necessary anymore. Removed.

src/components/organisms/Messenger/Messenger.vue
pkleybolte: The user might not be in a specific room at all, but he can still use the overall chat room. That is the roomID == null case.

src/store/auth.js
pkleybolte: If  it is not defined then nothing is stored there by matrix and we don't need to remove anything.

I don't know if the data is stored anywhere else permanently if no indexedDB is present or if it is only held in a var until the tab is closed. I can't find any information about that.

src/components/organisms/Messenger/Messenger.unit.js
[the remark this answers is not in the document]
CeEv: Yes and no. This should simulate the stuff that is executed in client browser. For test it should work with the common but it simulate it in the wrong way. The right way is to change process env, and re-execute the env file.

src/components/organisms/Messenger/Messenger.vue
CeEv: ok

src/store/messenger.unit.js
CeEv: https://github.com/hpi-schul-cloud/schulcloud-server/pull/1843/files#diff-9e55f73328d4c9b8eb2e6aa86f713f6badfaa32b45c80524cf9fd0ea8c9a2824R54  this work well

src/store/messenger.unit.js
CeEv: ok

src/components/organisms/Messenger/Messenger.vue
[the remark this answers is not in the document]
pkleybolte: I moved to the messenger's store.
I guess it makes sense to put the feature flags into the store so the components don't need to worry where they come from.
We can decide late if we put the feature flags into the store module that relates to the feature or we could have one store module for all the flags.

pkleybolte (commented):
The Changes I recommended are done

src/components/organisms/Messenger/Messenger.vue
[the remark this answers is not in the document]
CeEv: I like the solution with one store.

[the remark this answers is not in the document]
CeEv: But at the moment nuxt.config.js exist.

CeEv (approved):

PR #1277, opened 2020-08-24
SC-4260 - sampleRate added for sentry configuration

tofixx:
# Description

Adding `sampleRate`and fix `environment` in Sentry configuration. 

https://ticketsystem.schul-cloud.org/browse/SC-4260

## Approval for review

### Link to Definiton of Done

### DIFF
--- a/sentry.config.js
+++ b/sentry.config.js
@@ -1,6 +1,6 @@
 const pkg = require("./package");
 
-const environment = process.env.NODE_ENV || "development";
+const environment = process.env.NODE_ENV;
 const sampleRate = process.env.SENTRY_SAMPLE_RATE
 	? parseFloat(process.env.SENTRY_SAMPLE_RATE)
 	: 1.0;
--- a/sentry.config.js
+++ b/sentry.config.js
@@ -1,9 +1,15 @@
 const pkg = require("./package");
 
+const environment = process.env.NODE_ENV || "development";
+const sampleRate = process.env.SENTRY_SAMPLE_RATE
+	? parseFloat(process.env.SENTRY_SAMPLE_RATE)
+	: 1.0;
+
 module.exports = {
 	config: {
-		// environment: "{{@root.env}}", // TODO implement & enable
 		release: pkg.version,
+		environment,
+		sampleRate,
 		beforeSend: (event) => {
 			function removeIds(url) {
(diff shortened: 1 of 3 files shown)


sentry.config.js
```diff
@@ -1,9 +1,15 @@
 const pkg = require("./package");
 
+const environment = process.env.NODE_ENV || "development";
+const sampleRate = process.env.SENTRY_SAMPLE_RATE
+	? parseFloat(process.env.SENTRY_SAMPLE_RATE)
+	: 1.0;
```
CeEv, line 6: I like if we have the default value 0.0.

CeEv (changes requested):

sentry.config.js
tofixx: How to set it in every live system and develop @CeEv ?

sentry.config.js
```diff
@@ -1,9 +1,15 @@
 const pkg = require("./package");
 
+const environment = process.env.NODE_ENV || "development";
```
CeEv, line 3: Please move all ENVs to single point and set defaults one time. Please keep in mind that we want to render it in browser in future.

CeEv (changes requested):

sentry.config.js
tofixx: This is not part of this story. I removed the default value and keep it in my mind. There are no other places setting the value but 50 calls to process.env in 21 files which should be cleaned up in a follow up.

sentry.config.js
tofixx: @JanuszTaterka and @mcirikovic have been asked for input on this question and would support 1.0 as default.

CeEv: We already discuss it over month and after the last crash off all instances because adding 1.0 as default @janrenz follow my proposal to set it in every instance to 0.1 and the default to 0.0. 
If you want to change it, please discuss it again with @janrenz.

sentry.config.js
CeEv: Can you please create a ticket for it?

sentry.config.js
janrenz: i like default off or 0, unless we have mode specific defaults.

tofixx:
I will close this pr, I thought we wanted to enable this setting as additional option.

PR #1449, opened 2020-10-22
SC-6293 - restricts Lern-Store with permission

virgilchiriac:
# Description

[SC-6293](https://ticketsystem.hpi-schul-cloud.org/browse/SC-6293)
https://github.com/hpi-schul-cloud/schulcloud-server/pull/1864
https://github.com/hpi-schul-cloud/schulcloud-client/pull/2280

## Changes
- restricts Lern-Store access by new permission LERNSTORE_VIEW
- store auth, now loads full permissions for user, e.g. school permission too, not just the ones on his role. 
Sometimes, we can have permissions defined on schools, which can overwrite the ones from the role. This was not working in nuxt. The fix is simply doing the same we do in [client](https://github.com/hpi-schul-cloud/schulcloud-client/blob/develop/helpers/authentication.js#L118), thus also keeping it consistent. 

## Approval for review

### Link to Definiton of Done

### DIFF
--- a/src/store/auth.js
+++ b/src/store/auth.js
@@ -48,6 +48,10 @@ export const actions = {
 	},
 	async populateUser({ commit }) {
 		const user = await this.$axios.$get("/me");
+
+		const roles = await this.$axios.$get(`/roles/user/${user.id}`);
+		user.permissions = roles.reduce((acc, role) => [...new Set(acc.concat(role.permissions))], []);
+
 		commit("setUser", user);
 		if (user.schoolId) {
 			const school = await this.$axios.$get(`/schools/${user.schoolId}`);
--- a/src/store/auth.js
+++ b/src/store/auth.js
@@ -50,7 +50,10 @@ export const actions = {
 		const user = await this.$axios.$get("/me");
 
 		const roles = await this.$axios.$get(`/roles/user/${user.id}`);
-		user.permissions = roles.reduce((acc, role) => [...new Set(acc.concat(role.permissions))], []);
+		user.permissions = roles.reduce(
+			(acc, role) => [...new Set(acc.concat(role.permissions))],
+			[]
+		);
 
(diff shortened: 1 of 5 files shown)


Ojself (approved):

ABialoskorski (approved):

JenyMzo (approved):

src/store/auth.js
```diff
@@ -48,6 +48,13 @@ export const actions = {
 	},
 	async populateUser({ commit }) {
 		const user = await this.$axios.$get("/me");
+
+		const roles = await this.$axios.$get(`/roles/user/${user.id}`);
+		user.permissions = roles.reduce(
+			(acc, role) => [...new Set(acc.concat(role.permissions))],
+			[]
+		);
```
CordlessWool, line 56: use permissions/user route here

CordlessWool (changes requested):

src/store/auth.js
virgilchiriac: The permissions/user route does not take out the LERNSTORE_VIEW permission from the role, when school has 
```
"permissions": {
        "student": {
            "LERNSTORE_VIEW": false
        }
```
Moreover, we use the roles/user route in client [here](https://github.com/hpi-schul-cloud/schulcloud-client/blob/develop/helpers/authentication.js#L118)

virgilchiriac: As this behaviour is unclear if it is on purpose, I reported it as a possible bug, so that it can be solved separately and does not block this feature.
[ARC-125](https://ticketsystem.hpi-schul-cloud.org/browse/ARC-125)

CordlessWool: Should be not so complicate to solve this bug in the backend. There is nothing unclear, just solve the bug in the backend and use the correct route!

PR #1665, opened 2021-02-15
SC-8613 - LDAP-config-verify: Add checkbox to choose full or partial lo…

pkleybolte:
…ad of users

# Description

## Links to Tickets or other pull requests
https://ticketsystem.hpi-schul-cloud.org/browse/SC-8613
https://github.com/hpi-schul-cloud/schulcloud-server/pull/2274

## Approval for review

- [ ] All points were discussed with the ticket creator, support-team or product owner. The code upholds all quality guidelines from the PR-template.

> Notice: Please remove the WIP label if the PR is ready to review, otherwise nobody will review it.

### Link to Definiton of Done

More and detailed information on the _definition of done_ can be found [on Confluence](https://docs.schul-cloud.org/pages/viewpage.action?pageId=92831762)

PR #1666, opened 2021-02-17
Revert "SC-8613 - LDAP-config-verify: Add checkbox to choose full or partial lo…"

pkleybolte:
Reverts hpi-schul-cloud/nuxt-client#1665

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "schulcloud-nuxt",
-	"version": "25.6.2",
+	"version": "25.6.1",
 	"description": "The next Level of HPI Schul-Cloud",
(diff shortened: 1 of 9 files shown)


CordlessWool (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "schulcloud-nuxt",
-	"version": "25.6.2",
+	"version": "25.6.1",
```
alriese, line 3: Is the version correct? To avoid confusion we could keep the version as it is.

alriese (commented):
We are not sure about the right version, but otherwise the PR looks good to us.

PR #1759, opened 2021-04-29
SC-9004 - nuxt env config

murat-merdoglu-hpi:
# Short Description
- Config store created to fetch env variables from backend. It syncs the env variables between nuxt and server.
- These env variables are moved from the client to the server.

> 'ADMIN_TABLES_DISPLAY_CONSENT_COLUMN',
		'JWT_SHOW_TIMEOUT_WARNING_SECONDS',
		'JWT_TIMEOUT_SECONDS',
		'FEATURE_EXTENSIONS_ENABLED',
		'FEATURE_TEAMS_ENABLED',
		'NOT_AUTHENTICATED_REDIRECT_URL',
		'FEATURE_ES_COLLECTIONS_ENABLED',

- https://ticketsystem.hpi-schul-cloud.org/browse/SC-9004
- https://github.com/hpi-schul-cloud/schulcloud-server/pull/2522

## Changes

- Config store created to fetch env variables from backend
- https://docs.hpi-schul-cloud.org/display/SCPR/Nuxt+Environment+Variables

## Checklist before merging


http-headers.config.js
```diff
@@ -1,3 +1,20 @@
+/*
+=================================================================
```
CeEv, line 2: I love it. :D

src/store/env-config.js
```diff
@@ -0,0 +1,42 @@
...
+export const mutations = {
+	setEnv(state, payload) {
+		state.env = payload;
+	},
+	setError(state, error) {
+		state.error = error;
+	},
+};
+
+export const getters = {
```
CeEv, line 24: On the most places the .env.[name] is called directly from state. Is this fine, when the getter exist?

CeEv (changes requested):

nuxt.config.js
```diff
@@ -133,6 +110,7 @@ module.exports = {
 	 ** Plugins to load before mounting the App
 	 */
 	plugins: [
+		"@plugins/configInit",
```
Metauriel, line 113: comment that this needs to go first

src/store/env-config.js
```diff
@@ -0,0 +1,42 @@
...
+	getEnv(state) {
+		return state.env;
+	},
```
Metauriel, line 27: find a way of warning a developer about accessing a variable that doesnt exist

src/middleware/links-fallback.js
```diff
@@ -5,8 +5,9 @@ const isNuxtRoute = (url) =>
```
Metauriel, line 5: figure out what this file actually does

src/pages/administration/teachers/index.vue
```diff
@@ -296,9 +299,7 @@ export default {
 			return !this.school.isExternal;
 		},
 		showConsent() {
-			return process.env["ADMIN_TABLES_DISPLAY_CONSENT_COLUMN"] === "false"
-				? false
-				: true;
+			return this.env.ADMIN_TABLES_DISPLAY_CONSENT_COLUMN;
```
dominikbruechner, line 302: I just stumbled upon it... it is inverting the ENV variable? Just double-check because the logic seems a bit unclear. maybe !this.env.ADMIN_TABLES_DISPLAY_CONSENT_COLUM? I just do not understand it

src/plugins/configInit.js
```diff
@@ -0,0 +1,7 @@
+export default async ({ store }) => {
+	try {
+		await store.dispatch("env-config/get");
+	} catch (error) {
+		console.error(error);
+	}
+};
```
dominikbruechner, line 7: is it retrying if it fails? in what cases could it fail? do we have defaults?

src/store/auth.js
```diff
@@ -65,10 +65,10 @@ export const actions = {
 		}
 
 		//TODO Remove once added to User permissions SC-2401
-		if (process.env["FEATURE_EXTENSIONS_ENABLED"] === "true") {
+		if (rootState["env-config"].env.FEATURE_EXTENSIONS_ENABLED) {
 			commit("addUserPermission", "ADDONS_ENABLED");
 		}
-		if (process.env["FEATURE_TEAMS_ENABLED"] === "true") {
+		if (rootState["env-config"].env.FEATURE_TEAMS_ENABLED) {
```
dominikbruechner, line 71: this really should be moved to user perm at some point ^^

src/store/autoLogout.js
```diff
@@ -134,7 +140,9 @@ export const state = () => {
 	return {
 		active: false,
 		error: false,
-		remainingTimeInSeconds: defaultRemainingTimeInSeconds,
+		remainingTimeInSeconds: 3600 * 2,
+		showWarningOnRemainingSeconds: 3600,
+		defaultRemainingTimeInSeconds: 3600 * 2,
```
dominikbruechner, line 145: can we move this in some kind of configuration file? I think i f I would start looking for it, maybe I would search in the store, but maybe not. just a suggestion

src/store/env-config.js
dominikbruechner: +1

src/pages/administration/teachers/index.vue
pkleybolte: Reading from the console is always as a string. So if the string was "false" the boolean value had to be false.
The var loaded from the server in JSON format is a boolean already. No conversion needed anymore.

src/store/env-config.js
j-jmnz: There is actually no use for this getter as we are mostly reaching to the rootState to access this store. Can we get rid of this?

src/plugins/configInit.js
pkleybolte: I added a retry mechanism to the store module. If it fails it uses some defaults, but prints the error in the console.

src/store/autoLogout.js
pkleybolte: This exists until loading of the vars from the server is done or failed.
This is just to have a value so that the user is not shown the modal before the state is properly initialized.

src/store/env-config.js
pkleybolte: A var can always be null. From my point of view it is not necessary to check here. Depends on the var and how it is used in the component.
Some vars that are necessary to even start the client have defaults in `requiredVars`

CeEv (approved):

PR #1778, opened 2021-05-18
SC-9050 homeworks dashboard empty state [Odyssey] 

martynakrysinska:
# Short Description

Implemented empty state for the new homeworks dashboard. 

## Checklist before merging


src/pages/beta/dashboard.vue
```diff
@@ -1,16 +1,23 @@
 <template>
 	<v-container class="v-container">
-		<h1 class="h4">{{ $t("pages.homeworks.title") }}</h1>
-		<homeworks-list :homeworks="homeworks" />
+		<h1 v-if="homeworks.length > 0" class="h4">{{ $t("pages.homeworks.title") }}</h1>
+		<homeworks-list v-if="homeworks.length > 0" :homeworks="homeworks"/>
+		<tasks-empty-state 
+			v-if="homeworks.length === 0"
```
dominikbruechner, line 6: here you can do v-else

src/components/molecules/TasksEmptyState.vue
```diff
@@ -0,0 +1,40 @@
...
+import tasksEmptyState from "@assets/img/empty-state/Task_Empty_State.svg";
+import BaseImage from "@basecomponents/BaseImage";
+
+export default {
+	components: { BaseImage },
+	layout: "defaultVuetify",
+	props: {
+		image: {
+			type: String,
+			default: () => tasksEmptyState,
```
dominikbruechner, line 24: To make it reusable, I would suggest to make it _required: true_ and pass the image path always as a prop

dominikbruechner: imagePath: {
			type: String,
			required: true,
		},

src/components/molecules/TasksEmptyState.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+    <v-container class="v-container">
+		<base-image
+			:img-src="image"
+			:img-height="imgHeight"
+			fill="var(--color-primary)"
+			role="presentation"
+		/>
+		<h2 class="h4 text-center mt-8">{{ title }}</h2>
+		<h3 class="h4 text-center mt-3">{{ subtitle }}</h3>
```
dominikbruechner, line 10: as title and subtitle are not required, I would suggest to add a v-if (you would need to try this out, if it works with the default "" value)

Additionally: If we do not display a h1 element, then this needs to be changed accordingly

```suggestion
		<h1 v-if="title" class="h4 text-center mt-8">{{ title }}</h1>
		<h2 v-if="subtitle" class="h4 text-center mt-3">{{ subtitle }}</h2>
```

src/components/molecules/TasksEmptyState.vue
```diff
@@ -0,0 +1,40 @@
...
+			type: String,
+			default: () => "",
+		},
+		subtitle: {
+			type: String,
+			default: () => "",
+		},
+	},
+	data: () => ({
+		imgHeight: "400px"
```
dominikbruechner, line 36: I would add the imgHeight as a prop with default value

src/components/organisms/HomeworksList.vue
```diff
@@ -45,7 +45,7 @@ export default {
 	data() {
 		return {
 			fromNow,
-			taskImage: taskImage,
+			taskImage,
```
dominikbruechner, line 48: We should generally rename this to taskIconSvg, to make it clearer (just realizing after re-reading my code)

src/pages/beta/dashboard.vue
```diff
@@ -1,16 +1,23 @@
 <template>
 	<v-container class="v-container">
-		<h1 class="h4">{{ $t("pages.homeworks.title") }}</h1>
-		<homeworks-list :homeworks="homeworks" />
+		<h1 v-if="homeworks.length > 0" class="h4">{{ $t("pages.homeworks.title") }}</h1>
+		<homeworks-list v-if="homeworks.length > 0" :homeworks="homeworks"/>
```
dominikbruechner, line 4: this might or might not work with v-if="homeworks", or did you try it already ?

src/components/molecules/TasksEmptyState.unit.js
```diff
@@ -0,0 +1,54 @@
...
+
+let vuetify;
+let wrapper; 
+const title = "Test title";
+const subtitle = "Test subtitle";
+const image = "@assets/img/empty-state/Task_Empty_State.svg";
+
+describe("@components/molecules/TasksEmptyState", () => {
+
+	beforeEach(() => {
```
dominikbruechner, line 13: nice!

src/components/molecules/TasksEmptyState.unit.js
```diff
@@ -0,0 +1,54 @@
...
+			propsData: {
+				image,
+				title,
+				subtitle
+			}
+		});
+	});
+	
+	afterEach(() => {
+		wrapper.destroy();
```
dominikbruechner, line 33: also nice!

dominikbruechner (changes requested):
Hey there, generally nice work, particularly regarding the tests!

One thing that I thought about. This component already exists and is already used on many pages: https://storybook.test.hpi-schul-cloud.org/?path=/story/5-molecules-emptystate--default 

The EmptyState Layout for tasks looks a bit different. And we might want to use v-img. 
I really would like to have generic components, in my opinion, the case for keeping the new one would be that it plays nicely with vuetify. 

Suggestion: Rename the component to vCustomEmptyState and make it generic

Also, please add a storyBook entry, for now, you could place it next to the v-btn story which we have on the f-odyssey branch

cheers ✌️

src/pages/beta/dashboard.vue
martynakrysinska: Done

src/components/molecules/TasksEmptyState.vue
martynakrysinska: Done

src/components/molecules/TasksEmptyState.vue
martynakrysinska: I made the title required, but left subtitle optional in case we want to display a shorter message. And the h2 will only be shown if the subtitle is passed as a prop.

src/components/molecules/TasksEmptyState.vue
martynakrysinska: Done

src/pages/beta/dashboard.vue
martynakrysinska: This will actually display the HomeworksList component with no items on it, so the length check seems necessary here.

dominikbruechner (approved):

src/pages/tasks/open.vue
```diff
@@ -1,17 +1,31 @@
...
 
 <script>
 import HomeworksList from "@components/organisms/HomeworksList";
+import vCustomEmptyState from "@components/molecules/vCustomEmptyState";
+import tasksEmptyState from "@assets/img/empty-state/Task_Empty_State.svg";
 import { mapGetters } from "vuex";
 
 export default {
-	components: { HomeworksList },
-	layout: "defaultVuetify",
```
dominikbruechner, line 14: shouln't this stay here?

src/components/molecules/vCustomEmptyState.vue
```diff
@@ -0,0 +1,41 @@
...
+		<h2 v-if="subtitle" class="h4 text-center mt-3">{{ subtitle }}</h2>
+	</v-container>
+</template>
+
+<script>
+import BaseImage from "@basecomponents/BaseImage";
+
+export default {
+	components: { BaseImage },
+	layout: "defaultVuetify",
```
dominikbruechner, line 19: and go out of here?

src/pages/tasks/open.vue
dominikbruechner: the layouts should always be used on page level

src/components/molecules/vCustomEmptyState.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+	<v-container>
```
dominikbruechner, line 2: in this case v-container is put twice (once in the page, once in the component), better to keep it outside of the component?

martynakrysinska: The template root expects only one element, that's why I wrapped the BaseImage and headers in the v-container.

src/pages/tasks/open.vue
martynakrysinska: Done. I deleted the layout from HomeworksList as well. Now it is only in open.vue.

PR #1783, opened 2021-05-21
SC-9077 - store/use of store refactoring

j-jmnz:
# Short Description

https://ticketsystem.hpi-schul-cloud.org/browse/SC-9077

## Checklist before merging

### DIFF
--- a/src/components/organisms/DatasourceRunWebuntis.vue
+++ b/src/components/organisms/DatasourceRunWebuntis.vue
@@ -95,7 +95,7 @@ import ModalFooterConfirm from "@components/molecules/ModalFooterConfirm";
 import FormActions from "@components/molecules/FormActions";
 import DataTable from "@components/organisms/DataTable/DataTable";
 
-import { mapGetters } from "vuex";
+import { mapState } from "vuex";
 export default {
 	components: {
 		ModalBodyInfo,
@@ -152,8 +152,11 @@ export default {
 		};
 	},
 	computed: {
-		...mapGetters("webuntis-metadata", {
+		...mapState("webuntis-metadata", {
 			webuntisMetadata: "list",
+		}),
+		...mapState("datasourceRuns", {
+			runData: "list",
 		}),
 		tableDataObject() {
 			return this.tableData.reduce((obj, row) => {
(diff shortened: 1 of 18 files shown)


src/components/organisms/DatasourceRunWebuntis.vue
```diff
@@ -95,7 +95,7 @@ import ModalFooterConfirm from "@components/molecules/ModalFooterConfirm";
 import FormActions from "@components/molecules/FormActions";
 import DataTable from "@components/organisms/DataTable/DataTable";
 
-import { mapGetters } from "vuex";
+import { mapState } from "vuex";
```
CeEv, line 98: I think in generally we should not give any componet or side a direct access to the store, like $store.state.xxx..
Any data should comsume over $store.getters.xxx..
The same for mutations.

src/components/organisms/DatasourceRunWebuntis.vue
```diff
@@ -152,9 +152,12 @@ export default {
 		};
 	},
 	computed: {
-		...mapGetters("webuntis-metadata", {
+		...mapState("webuntis-metadata", {
```
CeEv, line 155: And really i think the service template in the stores is shit. :D 
But this is not a part of your PR.

CeEv (changes requested):

src/components/organisms/DatasourceRunWebuntis.vue
j-jmnz: According to vuex good practices, it is preferable to consume data by giving direct access to the store or by mapping the state to the component over having redundant getters that return the state as it is without any modification.
You can read a bit more about this over here: https://dev.to/timothyokooboh/vuex-best-practices-45dd

src/components/organisms/DatasourceRunWebuntis.vue
dominikbruechner: We need a service template v2, which delivers us all the standard methods for custom endpoints like GET/PATCH/DELETE/POST, while sticking to good practices. 

Particulary we need per default a #loading# state.

src/components/organisms/DatasourceRunWebuntis.vue
dominikbruechner: https://hackernoon.com/correct-and-efficient-way-to-use-vuex-part-i-wf15432eu 

Here as well.
I am talking to Cedric about this as well, after reading up on this I am also making the case for the use of MapState, if we do not need to filter data

src/components/organisms/DatasourceRunWebuntis.vue
j-jmnz: @dominikbruechner that's the follow-up of this ticket 👍

pkleybolte (approved):

PR #1789, opened 2021-05-27
Feature/odyssey homeworks loading state

martynakrysinska:
# Short Description

First implementation of the loading state for homeworks list and its headers, however the work is still in progress.

The templates of open.vue and HomeworkList.vue have way too many if/else statements. This still requires refactoring and consideration for error state that we want to add in the future. 
Mocking the store with different states seems  repetitive, im still not sure what would be the best solution for that. 

## Checklist before merging


src/pages/tasks/open.vue
```diff
@@ -1,11 +1,18 @@
...
+		<template v-else>
+			<h1 v-if="homeworks.length > 0" class="h4">
+				{{ $t("pages.homeworks.title") }}
+			</h1>
+		</template>
+
+		<homeworks-list :homeworks="homeworks" />
 		<v-custom-empty-state
-			v-else
+			v-if="homeworks.length === 0 && !loading.homeworks"
```
dominikbruechner, line 15: Probably you want to do something like:

`
IF loading.homeworks
   //show skeleton loader for headline
   //show skeleton loader for homeworks list
ELSE {
   IF homeworks.length > 0
      //show homeworks headline
      //show homeworksList
  ELSE
     //show emptystate
}
`

src/store/homeworks.js
```diff
@@ -5,10 +5,12 @@ const base = serviceTemplate("homework");
 const module = {
 	actions: {
 		getHomeworksDashboard: async function ({ commit }) {
+			commit("setLoading", { homeworks: true });
```
CeEv, line 8: try catch is missing, if you get an error from axios then you want to write it to store > error in future. But you must also setLoading to false in error case. 

Moving loading to service template bring you also to a problem. The page should not know what key is loading, it is loading or not. It is fine if you define more loading keys. Because it is possible to send more then on request over different dispatch. In a perfect world it should not because for this you must write a additional dispatch that combine it, but in real live you can run into this situations.
But the page will only the info is load or is not.

Also the loading getter not return homework : true ..it return all possible keys inside of loading are true. But the service template do not know all keys. You can set it directly by initial the store and go over Object.keys and array.every ..but this is hidden logic outside the main store.

CeEv (changes requested):

src/store/homeworks.js
```diff
@@ -5,10 +5,12 @@ const base = serviceTemplate("homework");
...
 			const res = await this.$axios.$get("/v3/task/dashboard/");
```
CeEv, line 9: The test if missing for axios throw an error.

CeEv (changes requested):

CeEv (approved):

dominikbruechner:

PR #1867, opened 2021-08-19
Feature/bc 5 chips

odalys-dataport:
# Short Description

## Checklist before merging


src/store/homeworks.js
```diff
@@ -14,12 +15,12 @@ const module = merge(base, {
 			try {
 				const data = await fetchAll(this.$axios, "/v3/task/dashboard/");
 				commit("set", {
-					items: data,
+					items: data.concat(homeworks),
```
martynakrysinska, line 18: oops, this shouldn't be here

martynakrysinska: basically all the changes in this file should be rolled back

src/components/molecules/vHomeworkItemStudent.vue
```diff
@@ -21,24 +21,28 @@
...
 			/>
-			<v-custom-chip-time-remaining
+			<v-custom-chip-task-state
 				v-else-if="isOverDue(homework.duedate)"
 				type="overdue"
 				:due-date="homework.duedate"
 				:shorten-date="$vuetify.breakpoint.xsOnly"
 			/>
+			<v-custom-chip-task-state
+				v-else-if="isGraded(homework.status.graded)"
```
martynakrysinska, line 37: There is a way to write this without if statements in the template:

We could create a computed property called smth like "type" or "taskState" that would return "warning", "overdue" or "graded" based on the condition (isCloseToDueDate, isOverDue or isGraded - these by the way should also be computed properties instead of methods) and pass that  as the "type" prop to the v-custom-chip-task-state.

martynakrysinska (changes requested):

PR #1876, opened 2021-08-25
BC-133 - filePaths and accounts module tests

j-jmnz:
# Short Description
Tests for filePaths and accounts modules

https://ticketsystem.hpi-schul-cloud.org/browse/BC-133

## Checklist before merging


virgilchiriac (approved):

src/store/accounts.unit.ts
```diff
@@ -0,0 +1,89 @@
+import { AccountsModule } from "./accounts";
```
alriese, line 1: We don't need to delete a old test file in this case, because the we did not have these tests before, right?

alriese (approved):

PR #1878, opened 2021-08-27
BC-132 - autologout, content, and env-config modules tests

j-jmnz:
# Short Description
Adds autologout, content, and env-config modules tests

https://ticketsystem.hpi-schul-cloud.org/browse/BC-113

## Checklist before merging


src/store/autoLogout.unit.ts
```diff
@@ -0,0 +1,112 @@
...
+		});
+		it("setReaminingTimeInSeconds should set remainingTimeInSeconds value", () => {
+			const autoLogoutModule = new AutoLogoutModule({});
+			const mockValue = 12345;
+
+			expect(autoLogoutModule.remainingTimeInSeconds).not.toBe(mockValue);
+			autoLogoutModule.setRemainingTimeInSeconds(mockValue);
+			expect(autoLogoutModule.remainingTimeInSeconds).toBe(mockValue);
+		});
+		it("setInit should set default values if no paramaters are defined", () => {
```
tofixx, line 51: describe('when initialize ithout setting default values', ()=>{
it('should set a default value for when show a warning', ()=>{
...})
it('should set a default remaining time value', ()=>{
...})
})

src/store/autoLogout.ts
```diff
@@ -180,8 +181,8 @@ export class AutoLogoutModule extends VuexModule {
 
 	@Mutation
 	setInit(
-		showWarningOnRemainingSeconds: number,
-		defaultRemainingTimeInSeconds: number
+		showWarningOnRemainingSeconds?: number,
+		defaultRemainingTimeInSeconds?: number
 	) {
 		this.showWarningOnRemainingSeconds = showWarningOnRemainingSeconds || 3600;
```
tofixx, line 187: this does not seem to be solid: the default value is set already in https://github.com/hpi-schul-cloud/nuxt-client/blob/e34cc71b6c30b5617d53475a973bdae033fd262b/src/store/autoLogout.ts#L150 on initialization. therefore the value must be given when creating a new instance of a new AutoLogoutModule() immediately. 
when needed, the method should call a prop setter which ensures the no invalid values can be given.

src/store/env-config.unit.ts
```diff
@@ -0,0 +1,201 @@
+import { EnvConfig } from "./env-config";
+import { initializeAxios } from "../utils/api";
+import { NuxtAxiosInstance } from "@nuxtjs/axios";
+
+const mockEnvs = {
```
tofixx, line 5: is this available as an defined interface?

src/store/env-config.unit.ts
```diff
@@ -0,0 +1,201 @@
...
+	DOCUMENT_BASE_DIR: "mockValue",
+	SC_TITLE: "mockValue",
+	SC_SHORT_TITLE: "mockValue",
+};
+
+const URL = "/v1/config/app/public";
+let requestPath: string;
+
+const axiosInitializer = (envs?: any, error?: boolean) => {
+	initializeAxios({
```
tofixx, line 33: has you found out how to solve this by mocking axios instead of using this workaround? @pkleybolte may help?

src/store/filePaths.unit.ts
```diff
@@ -112,15 +112,15 @@ describe("filePaths module", () => {
 			const mockSpecificFiles = mockSetSpecificFiles(mockUrl);
 
 			filePathsModule.setSpecificFiles(mockUrl);
-			expect(filePathsModule.getSpecificFiles).toEqual(mockSpecificFiles);
+			expect(filePathsModule.getSpecificFiles).toStrictEqual(mockSpecificFiles);
```
tofixx, line 115: toEqual should not be used on complex data (objects) good, are there other occurences?

src/store/accounts.unit.ts
```diff
@@ -33,11 +33,11 @@ describe("accounts module", () => {
 
 			accountsModule.resetBusinessError = spyResetError;
 			accountsModule.setStatus = spySetStatus;
-			expect(spyResetError).not.toBeCalled();
-			expect(spySetStatus).not.toBeCalled();
+			expect(spyResetError).not.toHaveBeenCalled();
```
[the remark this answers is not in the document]
j-jmnz: these are the only places 👍

src/store/env-config.unit.ts
j-jmnz: yes here the type is defined https://github.com/hpi-schul-cloud/nuxt-client/blob/9d15977dccd7f2b339d3592aec0164ce68b67b7c/src/store/env-config.ts#L20

src/store/env-config.unit.ts
j-jmnz: no solution yet, talked with @pkleybolte and he's also ok with this workaround

src/store/filePaths.unit.ts
j-jmnz: only occurrences

.eslintrc.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
 				"jest/expect-expect": "off",
 				"jest/no-disabled-tests": "off",
 				"jest/valid-title": "off",
+				"jest/no-focused-tests": 2,
```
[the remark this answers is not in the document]
j-jmnz: https://github.com/hpi-schul-cloud/nuxt-client/pull/1883 :)

src/store/env-config.unit.ts
tofixx: then i would suggest using it like
```suggestion
const mockEnvs: Envs = {
```

src/store/autoLogout.ts
tofixx:
```suggestion
if(showWarningOnRemainingSeconds !== undefined){
		this.showWarningOnRemainingSeconds = showWarningOnRemainingSeconds;
}
```
on this.showWarningOnRemainingSeconds a setter could be added with property validation like  is number and gt zero

src/store/env-config.ts
```diff
@@ -28,7 +28,7 @@ type Envs = {
 	FEATURE_TEAMS_ENABLED?: boolean;
 	FEATURE_SCHOOL_POLICY_ENABLED?: boolean;
 	FEATURE_VIDEOCONFERENCE_ENABLED?: boolean;
-	FEATURE_MATRIX_MESSENGER_ENABLED?: boolean;
+	FEATURE_MATRIX_MESSENGER_ENABLED?: string | undefined;
```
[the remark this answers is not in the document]
j-jmnz: the compiler was complaining if undefined was not added as a type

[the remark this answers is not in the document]
tofixx: hm, I see. the definition is a type but could be an interface too. 
i suggest always using interfaces, use types only if the requirements of types are needed, for example combining multiple types. as an interface it will work by using the ?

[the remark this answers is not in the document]
tofixx: hm, I see. the definition is a type but could be an interface too. 
i suggest always using interfaces, use types only if the requirements of types are needed, for example combining multiple types. as an interface it will work by using the ?

src/store/content.unit.ts
```diff
@@ -0,0 +1,685 @@
+import { Content } from "./content";
+import { initializeAxios } from "../utils/api";
+import { NuxtAxiosInstance } from "@nuxtjs/axios";
+
+const ESPath = "/v1/edu-sharing";
+const lessonsPath = "/v1/lessons";
+let requestPath: string;
+
+const mockResource = {
```
tofixx, line 9: these structures should implement an interface. where is this defined?

src/store/env-config.unit.ts
```diff
@@ -0,0 +1,201 @@
...
+		});
+		it("setStatus should set status", () => {
+			const envConfigModule = new EnvConfig({});
+			expect(envConfigModule.status).toBe("");
+			envConfigModule.setStatus("completed");
+			expect(envConfigModule.status).toBe("completed");
+		});
+	});
+	describe("getters", () => {
+		it("getFallbackLanguage should get 'de' if I18N__FALLBACK_LANGUAGE is not defined", () => {
```
tofixx, line 184: i like that test, beside the others above it is testing logic which is good as other tests are just checking a value has been set which is not telling us anything about functionality of the application which happens here, which is: we have german as default language ;)

tofixx (dismissed):
i don't get the requirement for the setter/getter tests but the changes look clear to me.

tofixx (approved):

PR #1902, opened 2021-09-17
BC-266 load baseUrl from api call

pkleybolte:
# Short Description

Get API_URL via http call, this prepares us if there ever will be nuxt without the middleware and also makes this configuration more transparent. This also eliminates the necessity of inline scripting to get that env var

https://ticketsystem.hpi-schul-cloud.org/browse/BC-266

## Checklist before merging


Loki-Afro (dismissed):

Loki-Afro (approved):

src/plugins/axios.js
```diff
@@ -2,7 +2,13 @@ import AuthModule from "@/store/auth";
 
 const unrecoverableErrorCodes = [401, 404, 500];
 
-export default function ({ $axios, store, error }) {
+export default async function ({ $axios, store, error }) {
+	const runtimeConfigJson = await $axios.get(
+		`${window.location.origin}/runtime.config.json`
+	);
```
virgilchiriac, line 8: is it not possible to request the json directly, rather than making a ajax call?

src/plugins/axios.js
```diff
@@ -2,7 +2,13 @@ import AuthModule from "@/store/auth";
...
+	window.schoolCloudRuntimeConfig = runtimeConfigJson.data;
+	$axios.setBaseURL(window.schoolCloudRuntimeConfig.apiURL);
+
```
virgilchiriac, line 11: I wonder if the baseUrl can be set in a simpler, dynamic way, in which API_URL env would be available
https://axios.nuxtjs.org/helpers/#setbaseurl

have you tried to create a new axios instance? https://axios.nuxtjs.org/extend

src/plugins/axios.js
Loki-Afro: how direct do you wanna request something if not like that? there needs to be an http call.

src/plugins/axios.js
Loki-Afro: this is the most simple solution possible, we like simplicity.

PR #1919, opened 2021-09-29
Hotfix/bc 365 fix pagination

odalys-dataport:
# Short Description

Fixing the pagination for task overview, where only 10 tasks max. were shown.

- https://ticketsystem.hpi-schul-cloud.org/browse/BC-365
- https://github.com/hpi-schul-cloud/schulcloud-server/pull/3033

## Checklist before merging


CeEv (approved):

uidp (approved):

src/store/tasks.js
```diff
@@ -63,9 +65,9 @@ const module = {
 			commit("resetBusinessError");
 			commit("setStatus", "pending");
 			try {
-				const response = await this.$axios.$get("/v3/tasks/");
+				const data = await fetchAll(this.$axios, "/v3/tasks/");
```
pkleybolte, line 68: Instead of using `fetchAll` which create a lot of sequential requests which should make one request that gets all. That might require in the server, but that is better than too many requests.

pkleybolte (dismissed):
Code changes without test changes -> tests are missing.

src/store/tasks.js
odalys-dataport: How would we achieve this?

CeEv: We want to increase the pagination limit from nest. But for the moment it is not possible to get more then 10. But we increase it to 50 or so and maybe teachers run in the situation that they have more. 
The ux do not want pagination for this side (maybe for completed - not added at the moment). 

The logic in the store based on the condition that we have:
- no side pagination in der UI
- limited pagination from server
- but for right sorting in the sections we need all tasks in store (otherwise we must split multiple small single request for loading)
- the adoption is that fetchAll call most time only one request, for some maybe 2, but not more then 3 request in worst case

If you have a other proposal for it, it is fine and we can use it. For this PR, that is already merged in the release i like to merge it and change it in a additional pr. At the moment fetchAll is used in the current live code and remove in a PR that is in 26.11, add again with this pr.

CeEv (approved):

PR #2043, opened 2022-02-14
BC-877 - Spike: Research mocking of typescript stores in Nuxt

uidp:
BC-877 - Spike: Research mocking of typescript stores in Nuxt

## Checklist before merging


src/utils/mock-store-module.ts
```diff
@@ -0,0 +1,44 @@
...
+	module: Function & Mod<M, any>,
+	statics: StaticsType
+) => {
+	Object.keys(module.actions as ActionTree<M, any>).forEach((key) => {
+		// eslint-disable-next-line jest/prefer-spy-on
+		statics[key] = jest.fn();
+	});
+};
+
+export function createModuleMocks<M extends VuexModule>(
```
CeEv, line 27: It make sense for easy overriding of data that what we want as results from the store methodes. But in the other way we override all other too. It remove the results from the default store values. That is good and bad. Good for remove the real store values from test. Bad (if i understand correct) that we must set every needed method that is executed by the component in test for each test. 

It is also possible to pass keys for overriding that not exist. Can we add a warning if someone add a overriding method key to a test that not exist in the store modul?

CeEv (changes requested):

CeEv (approved):

PR #2084, opened 2022-03-15
BC-1342 - Add task or lesson to the room board

MartinSchuhmacher:
# Short Description
As a teacher, I want to add tasks and lessons directly on the board via FAB so that I don't have to jump back and forth between pages.

The FAB provides the option for teachers to create lessons and tasks for the course via the legacy work flow.
Students do not see the FAB due to missing permissions.

Ticket: https://ticketsystem.hpi-schul-cloud.org/browse/BC-1342

## Checklist before merging


Metauriel (approved):

locale/de.json
```diff
@@ -703,6 +707,16 @@
...
+	"pages.room.lessonShare.step_1.text": "Info",
+	"pages.room.lessonShare.step_2.text": "Code einfügen",
+	"pages.room.lessonShare.modal.info": "Verteile folgenden Link an deine Schüler:innen, um sie zum Kurs einzuladen. Die Schüler:innen müssen dafür angemeldet sein.",
+	"pages.room.lessonShare.import.step_2.text": "Füge einen Code ein, um Themen anderer Nutzer:innen zu importieren. Im Thema enthaltene Dateien werden in deinem Kursordner abgelegt.",
+	"pages.room.lessonShare.import.step_1.info_2": "<b>Achtung:</b> Tausche im Thema enthaltene, mit Nutzerdaten befüllte Tools (z.B. neXboard, Etherpad, GeoGebra) nachträglich manuell aus, da Änderungen daran sich sonst auf den Original-Kurs auswirken! Dateien (Bilder, Videos, Audio) sowie verlinktes Material sind nicht betroffen und können unverändert bleiben.",
+	"pages.room.lessonShare.textValidation": "Dieses Feld kann nicht leer sein.",
+	"pages.room.lessonShare.codeError": "Der Thema-Code wird nicht verwendet.",
+	"pages.room.lessonShare.importError": "Leider ist etwas schief gegangen",
+	"pages.room.lessonShare.btn.continue": "Weiter",
+	"pages.room.lessonShare.btn.confirm": "Importieren",
```
odalys-dataport, line 719: The value "Importieren" without any other words already occurs 4 or 5 times in our locale file. Not sure what the preferred approach is here, but maybe we should make this a "common.actions.import" key.

Also while looking into that I have found these duplicates (same for student). Probably a merge conflict error and should be cleaned up
"pages.administration.teachers.fab.import.aria": "Lehrkraft importieren",
"pages.administration.teachers.fab.import.ariaLabel": "Lehrkraft importieren",

src/components/molecules/ImportLessonModal.vue
```diff
@@ -0,0 +1,196 @@
...
+						@click="onClickStepper(2)"
+					>
+						{{ $t("pages.room.lessonShare.step_2.text") }}
+					</v-stepper-step>
+				</v-stepper-header>
+			</v-stepper>
+			<div class="step-sections">
+				<div v-if="step === 1">
+					<!-- eslint-disable vue/no-v-html -->
+					<p v-html="$t('pages.room.lessonShare.import.step_1.info_2')"></p>
```
odalys-dataport, line 35: Why is the usage of v-html neccessary here and is there maybe a better solution?

src/components/molecules/ImportLessonModal.vue
```diff
@@ -0,0 +1,196 @@
...
+						class="mt-1 text-field-lesson-code"
+						:rules="[textFieldValidation.required]"
+					></v-text-field>
+					<div v-if="businessError.message === 'not-found'">
+						<v-alert dense outlined type="error" class="code-error">
+							{{ $t("pages.room.lessonShare.codeError") }}
+						</v-alert>
+					</div>
+					<div v-if="businessError.message === 'not-created'">
+						<v-alert dense outlined type="error" class="create-error">
```
odalys-dataport, line 54: When do these alerts disappear? Should they be dismissible? Can both errors occur at the same time and what happens then?

src/components/molecules/ImportLessonModal.vue
```diff
@@ -0,0 +1,196 @@
...
+			if (this.step === 2 && this.sharedLessonData.code === "") {
+				return;
+			}
+			if (this.step === 2) {
+				await RoomModule.confirmImportLesson(this.sharedLessonData.code);
+				if (this.businessError.statusCode === "") {
+					this.cancel();
+				}
+				return;
+			}
```
odalys-dataport, line 148: I think it would be cleaner to have the logic above in a separate method

src/components/molecules/RoomLessonCard.vue
```diff
@@ -128,6 +130,14 @@ export default {
...
 					});
 				}
+
+				if (EnvConfigModule.getEnv.FEATURE_LESSON_SHARE) {
+					roleBasedMoreActions[Roles.Teacher].push({
+						icon: this.icons.mdiShareVariant,
+						action: () => this.$emit("open-modal", this.lesson.id),
+						name: this.$t("pages.room.lessonCard.label.share"),
+					});
+				}
```
odalys-dataport, line 140: This is a huge method already, can this be refactored?

src/components/templates/RoomDashboard.vue
```diff
@@ -130,13 +131,46 @@
...
 			</div>
 		</div>
+		<vCustomDialog
+			ref="customDialog"
+			:is-open="lessonShare.isOpen"
+			class="room-dialog"
+			@dialog-closed="lessonShare.isOpen = false"
+		>
+			<div slot="title" class="room-title">
+				<h4>Copy code has been generated!</h4>
```
odalys-dataport, line 141: Translation key missing?

src/components/molecules/ImportLessonModal.vue
```diff
@@ -0,0 +1,196 @@
...
+						</v-alert>
+					</div>
+					<div v-if="businessError.message === 'not-created'">
+						<v-alert dense outlined type="error" class="create-error">
+							{{ $t("pages.room.lessonShare.importError") }}
+						</v-alert>
+					</div>
+				</div>
+			</div>
+			<div class="button-section mt-8">
```
odalys-dataport, line 60: The vCustomDialog already has cancel and confirm buttons. Why are they not used? And can the vCustomDialog component be adjusted to work for this case as well?

src/components/templates/RoomDashboard.vue
```diff
@@ -130,13 +131,46 @@
...
+						{{ $t("pages.room.lessonShare.modal.info") }}
+					</p>
+				</div>
+				<div>
+					<v-text-field :value="lessonShare.token" outlined></v-text-field>
+				</div>
+				<v-divider class="mb-4"></v-divider>
+				<div class="share-cancel-button">
+					<v-btn
+						class="dialog-back-button"
```
odalys-dataport, line 156: I think we should avoid adding classes for the sake of unit testing, but I haven't come up with a better idea myself yet either

src/pages/rooms/_id/index.unit.ts
```diff
@@ -1,8 +1,10 @@
-import RoomModule from "@store/room";
+import RoomsModule from "@store/room";
```
odalys-dataport, line 1: In RoomDashboard it's "RoomModule". Do we have both?

src/pages/rooms/_id/index.vue
```diff
@@ -54,6 +75,70 @@ export default {
 		};
 	},
 	computed: {
+		fab() {
```
odalys-dataport, line 78: Let's call this `fabItems` as well to be clearer

src/pages/rooms/_id/index.vue
```diff
@@ -54,6 +75,70 @@ export default {
...
+			if (
```
odalys-dataport, line 79: Can this be simplified? Maybe by rejecting false values instead or separating the definition of fab items from deciding which are shown

odalys-dataport (changes requested):

locale/de.json
monaweitzenberg: I totally agree. I will solve this key issues in the next sprint with the ticket BC-1447.

monaweitzenberg: I would use line 716: Der Code wird nicht verwendet. And now we could create a generic one which is reusable. Same for all other keys.

src/components/molecules/ImportLessonModal.vue
MartinSchuhmacher: Both error messages can not be happening at the same time and are dismissable due to the v-if directive.

src/components/molecules/ImportLessonModal.vue
MartinSchuhmacher: After talking to Murat the method is already pretty small and there is no need to separate it further.

src/components/molecules/RoomLessonCard.vue
MartinSchuhmacher: We had a look on the method and find that keeping it together like this makes it easier to maintain and to see the logic for each role. So we recommend keeping it like this.

src/components/molecules/ImportLessonModal.vue
MartinSchuhmacher: Currently we need three buttons for the importing process. We will refactor the vCustomDialog component in the upcoming sprints to make it usable for this case as well.

src/pages/rooms/_id/index.vue
MartinSchuhmacher: We reworked the object now to fabItems but do not see the positive effect of simplifing the rest of it as it - like mentioned above - in this case also makes it harder to maintain.

src/components/molecules/ImportLessonModal.vue
MartinSchuhmacher: We use html tags in the language file and thats why the v-html is necessary. Right now we research for a better solution, so be welcome to make any suggestions :)

src/pages/rooms/_id/index.unit.ts
MartinSchuhmacher: This was solved now as well

src/pages/rooms/_id/index.vue
MartinSchuhmacher: Was done

PR #2117, opened 2022-04-01
BC-1408 - enabling school admin feature toggles

odalys-dataport:
# Short Description

## Checklist before merging


src/components/organisms/administration/PrivacySettings.vue
```diff
@@ -245,7 +245,7 @@ export default {
 		lernStoreVisibility() {
 			return this.permissions.student
 				? this.permissions.student.LERNSTORE_VIEW
-				: false;
+				: true;
```
CeEv, line 248: For the PR, yes this is the needed changed. But in generally should we really have helper like lernStoreVisibility that include his own default state in code for the configuration settings?

Loki-Afro: if this is not the only place where this is checked yes, otherwise: unnecessary

CeEv (dismissed):

src/components/organisms/administration/PrivacySettings.vue
CeEv: The primäre problem is that the remove of the permission from students, should result in lernstore is not visible for students.
But that is also wrong implement im legacy client and must be fixed for it too.
Therefore approve from my side.

uidp (approved):

src/store/auth.ts
```diff
@@ -178,6 +178,7 @@ export class Auth extends VuexModule {
 		return this.accesToken || false;
 	}
 
+	// TODO - why are we using toLowerCase() on permissions here?
```
CeEv, line 181: oh yes good questions... that make it hard to find something and work with it.

CeEv (approved):

PR #2153, opened 2022-05-05
BC-1685 - Substitute teacher can see course edit three dots/options

MartinSchuhmacher:
# Short Description

Substitute teachers can see the course edit option for the assigned course, which is expected to be seen by only main teacher of the course.

Ticket: https://ticketsystem.dbildungscloud.de/browse/BC-1685
## Checklist before merging


src/store/room.ts
```diff
@@ -281,11 +282,24 @@ export default class RoomModule extends VuexModule {
 		}
 	}
 
+	@Action
+	async fetchScopePermission(payload: object | any): Promise<void> {
```
virgilchiriac, line 286: we should specify the payload object, and not allow any, since we need the userId and courseId always
something like payload: {userId: string, courseId: string} should work

src/store/room.ts
```diff
@@ -281,11 +282,24 @@ export default class RoomModule extends VuexModule {
...
+		const ret_val = await $axios.$get(
+			`/v1/courses/${payload.courseId}/userPermissions?userId=${payload.userId}`
```
virgilchiriac, line 288: is the userId needed? It should take the current user, and a user should not be able to access permission of another user

src/store/room.ts
```diff
@@ -281,11 +282,24 @@ export default class RoomModule extends VuexModule {
...
+		);
+		this.setPermissionData(Object.values(ret_val[payload.userId]));
```
virgilchiriac, line 290: is Object.array is needed for typescript? Because that result is already an array

MartinSchuhmacher: you are right, I forgot to remove that, not needed actually

src/store/room.ts
MartinSchuhmacher: will do that, thx

virgilchiriac (approved):

PR #2176, opened 2022-05-20
BC-1644 - Disable name editing for synced school

christian-darsow:
# Short Description

Changing the name of the school on the new admin page should only be possible for non-synced schools, otherwise editing should be disabled.

https://ticketsystem.dbildungscloud.de/browse/BC-1644

## Checklist before merging


src/store/schools.unit.ts
```diff
@@ -761,5 +761,61 @@ describe("schools module", () => {
...
+					{
+						_id: "id_1",
+						type: "ldap",
+						ldapConfig: {
+							provider: "general",
+						},
+					},
+				];
+				schoolsModule.setSystems(systems);
+				expect(schoolsModule.schoolIsSynced).toStrictEqual(false);
```
MartinSchuhmacher, line 817: since we still support active moodle systems there should also be a test for this system in getIsSynced

src/components/organisms/administration/GeneralSettings.unit.js
```diff
@@ -166,6 +184,35 @@ describe("GeneralSettings", () => {
...
+
+		it("school name should be editable if the school is not synced", async () => {
+			const wrapper = mount(GeneralSettings, {
+				...createComponentMocks({
+					i18n: true,
+					store: generateMockStore,
+					vuetify: true,
+				}),
+			});
+			schoolsModule.setSystems([]);
```
MartinSchuhmacher, line 209: I am not sure if this test is correct or at least if there should be another test. Because even if a school has the `provider: "general"` system it is by definition not a synced school. There should be one systems array with only the `iserv-idm` case and another systems array for this case where `general`is set as provider and it would still be a editable. A combination of `general` and `iserv-idm` inside one array is no realistic case. Maybe you can then have it a third case like this is now, if there is no system at all.

MartinSchuhmacher (changes requested):

src/store/schools.unit.ts
christian-darsow: To test all possible cases for synced and non-synced schools tests for moodle and itslearning schools have now been added

src/components/organisms/administration/GeneralSettings.unit.js
christian-darsow: An explicit test for non-synced school has been added and the mocked systems collection has been adapted to be realistic

src/components/organisms/administration/GeneralSettings.unit.js
```diff
@@ -166,6 +183,56 @@ describe("GeneralSettings", () => {
...
+				...createComponentMocks({
+					i18n: true,
+					store: generateMockStore,
+					vuetify: true,
+				}),
+			});
+			const localMockData = {
+				localSchool: {
+					...mockData.localSchool,
+					systems: [{ $oid: "0000d186816abba584714c91" }],
```
MartinSchuhmacher, line 211: I recommend using the variable here as otherwise the value here needs to be changed as well once the value in the variable is changed.

MartinSchuhmacher (changes requested):

virgilchiriac (approved):

MartinSchuhmacher (dismissed):

virgilchiriac (dismissed):

MartinSchuhmacher (approved):

PR #2191, opened 2022-06-02
feature/BC-1552 - copy process via flag on roomoverview

murat-merdoglu-hpi:
# Short Description
- Added feature flag to switch copy process between old (redirecting to legacy) and new ones.

https://ticketsystem.dbildungscloud.de/browse/BC-1552

## Checklist before merging


src/components/templates/RoomDashboard.unit.ts
```diff
@@ -591,5 +595,21 @@ describe("@components/templates/RoomDashboard.vue", () => {
 			expect(wrapper.vm.copyProcess.data).toStrictEqual({});
 			expect(wrapper.vm.copyProcess.isOpen).toBe(false);
 		});
+
+		it("should redirect to legacy client if FEATURE_TASK_COPY_ENABLED flag is not set true", async () => {
+			const location = window.location;
```
virgilchiriac, line 600: is this working? should it not be mocked?

src/components/templates/RoomDashboard.vue
```diff
@@ -328,6 +328,10 @@ export default {
 			await roomModule.finishTask({ itemId, action: "restore" });
 		},
 		async copyTask(itemId) {
+			if (!envConfigModule.getEnv.FEATURE_TASK_COPY_ENABLED) {
+				window.location = `/homework/${itemId}/copy?returnUrl=rooms/${this.roomDataObject.roomId}`;
```
virgilchiriac, line 332: should it be window.location.href instead?

src/components/templates/RoomDashboard.unit.ts
murat-merdoglu-hpi: According to link you suggested: The `window.location` is a string value and typescript is complaining like 'assigning a fn() to a string value' or something like this. We can set `as any` properties but it is not a best practice.
Since `location.href` value is already being tested in the window object, I think no need to test it in another mock function.

MartinSchuhmacher (approved):

virgilchiriac (approved):

PR #2207, opened 2022-06-16
Feature/BC-1557 - Copy course

murat-merdoglu-hpi:
# Short Description

- https://ticketsystem.dbildungscloud.de/browse/BC-1557
- https://github.com/hpi-schul-cloud/schulcloud-server/pull/3506
- https://github.com/hpi-schul-cloud/dof_app_deploy/pull/287

## Checklist before merging


src/store/room.ts
```diff
@@ -321,7 +321,28 @@ export default class RoomModule extends VuexModule {
...
+	@Action
+	async copyRoom(courseId: string): Promise<void> {
+		this.resetBusinessError();
+		this.setLoading(true);
+		try {
+			const copyResult = await this.roomsApi.roomsControllerCopyCourse(
+				courseId
+			);
+
+			this.setCopyResult(copyResult.data || {});
```
uidp, line 345: Is it correct that we set an empty js-object when `copyResult.data` is falsy? The default value of `store.copyResult` seems to look differently after all.

src/components/organisms/CopyProcess.vue
```diff
@@ -102,27 +93,44 @@ export default {
...
+		getItemTitleAndStatus(title, status) {
+			const titleObj = {
+				metadata: this.$t("components.molecules.copyResult.metadata"),
+				description: this.$t("common.labels.description"),
+				tasks: this.$t("common.words.tasks"),
+				lessons: this.$t("common.words.topics"),
+				coursegroups: this.$t("common.words.courseGroups"),
+				times: this.$t("common.words.times"),
+			};
+			if (title === "files" && status === "not-implemented") {
```
uidp, line 105: Wouldn't it be better to use the type property here?

src/store/notifier.ts
```diff
@@ -0,0 +1,24 @@
...
+@Module({
+	name: "notifier",
+	namespaced: true,
+	stateFactory: true,
+})
+export default class NotifierModule extends VuexModule {
+	notifier: any = {};
+
+	@Action
+	show(payload: {}) {
```
uidp, line 12: How about making the store track the visibility state itself? Then we wouldn't have to watch the changes of the notifier data in the components.
We could also have an action to hide the notification then and even timouts etc. in the store.

src/store/room.ts
```diff
@@ -321,7 +321,28 @@ export default class RoomModule extends VuexModule {
...
+			this.setBusinessError({
+				statusCode: error?.response?.status,
+				message: error?.response?.statusText,
+				...error,
+			});
+		}
+	}
+
+	@Action
+	async copyRoom(courseId: string): Promise<void> {
```
uidp, line 337: Maybe we should move the course/room copying to a seperate store module? We could then move some additional logic there as well which is currently handled inside components (e.g. CopyProcess).

uidp (changes requested):

PR #2214, opened 2022-06-23
BC-1996 - fix eslint warnings in nuxt

NFriedo:
# Short Description

When starting the nuxt client, there are eslint warnings that should be fixed.

https://ticketsystem.dbildungscloud.de/browse/BC-1996

## Checklist before merging

### DIFF
--- a/src/components/organisms/ContentCard.vue
+++ b/src/components/organisms/ContentCard.vue
@@ -10,67 +10,63 @@
 				}"
 				:no-style="true"
 			>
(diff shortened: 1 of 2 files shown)


murat-merdoglu-hpi (approved):
Actually, the ContentCard component needs a lot of refactoring (ex: many methods should be in computed sections, and calling a method without parameter, etc. ), but it definitely was not in the ticket's scope.

PR #2225, opened 2022-07-07
BC-1956 configure blog url

uidp:
# Short Description
Provide the blog base URL using the config environment.

## Checklist before merging


src/utils/navbarBaseItems.js
```diff
@@ -1,17 +0,0 @@
-export default [
```
odalys-dataport, line 1: Thanks for deleting this, I wanted to do that as well :D

src/layouts/loggedOut.unit.ts
```diff
@@ -0,0 +1,54 @@
...
+			...createComponentMocks({
+				i18n: true,
+			}),
+			setup() {
+				provide("envConfigModule", envConfigModuleMock);
+			},
+			...attrs,
+			stubs: {
+				Nuxt: true,
+				NuxtLink: true,
```
odalys-dataport, line 25: If the links are nuxt links we should probably use "to" instead of "href" in the navbar links

src/layouts/loggedOut.vue
```diff
@@ -1,6 +1,6 @@
 <template>
```
odalys-dataport, line 1: Thanks for the renaming and implementing tests!

src/store/env-config.ts
```diff
@@ -41,6 +41,7 @@ export default class EnvConfigModule extends VuexModule {
 		DOCUMENT_BASE_DIR: "",
 		SC_TITLE: "",
 		SC_SHORT_TITLE: "",
+		GHOST_BASE_URL: "",
```
odalys-dataport, line 44: Is this a nuxt only thing? Is it coming from the server?

uidp: Yes. It's coming from the server. We have that environmant variable in both the server and the legacy client.

odalys-dataport (dismissed):

PR #2227, opened 2022-07-07
BC-1377 - integrated and updated prettier

odalys-dataport:
# Short Description

https://ticketsystem.dbildungscloud.de/browse/BC-1377

## Checklist before merging


.prettierrc.js
```diff
@@ -1,5 +1,13 @@
...
+	useTabs: true,
+	trailingComma: "es5",
+	bracketSpacing: true,
+	jsxBracketSameLine: false,
+	arrowParens: "always",
+	proseWrap: "never",
+	htmlWhitespaceSensitivity: "css",
+	endOfLine: "lf",
+	semi: true,
 	singleQuote: false,
```
MartinSchuhmacher, line 12: should we stick to the usage from https://github.com/hpi-schul-cloud/lint-configs/blob/master/packages/prettier-config/index.js and put `singleQuote: true`?

odalys-dataport: If you check the prettierrc.js in the nuxt client repo right now, we're overwriting that rule. That's why I decided to go with `singleQuote: false`

virgilchiriac (approved):

PR #2242, opened 2022-07-22
BC-2159 - Renaming Room in copy modal into Learning content

MartinSchuhmacher:
# Short Description
Change Naming of "Room" into "Learning content" im modal

BC-2159
## Checklist before merging


src/components/organisms/CopyProcess.vue
```diff
@@ -106,7 +106,7 @@ export default {
 	methods: {
 		getItemTitle(item) {
 			const titles = {
-				BOARD: this.$t("common.labels.room"),
+				BOARD: this.$t("common.words.learnContent"),
 				CONTENT: this.$t("components.molecules.copyResult.label.content"),
 				COURSE: this.$t("common.labels.room"),
```
hoeppner-dataport, line 111: Found other instances of common.labels.room - shouldn't they get replaced, too? (L. 111)

MartinSchuhmacher: no, because look at `COURSE` it uses the room label still and should do so. Before that (and in my opinion that was a bad decision) for BOARD and COURSE they used the same label, so its actually better now that we do not use the same label for two different usecases

hoeppner-dataport: Locked and mentioned L.111 which is "COURSE:" and which is the line I attached my comment to. Okay, then I only could contribute, that the title of the PR and of the ticket is maybe to general.

hoeppner-dataport (approved):

PR #2253, opened 2022-08-10
use consent feature flag

odalys-dataport:
# Short Description

https://ticketsystem.dbildungscloud.de/browse/BC-2289

## Checklist before merging

### DIFF
--- a/src/pages/administration/StudentOverview.page.vue
+++ b/src/pages/administration/StudentOverview.page.vue
@@ -294,11 +294,17 @@ export default {
 		schoolIsExternallyManaged() {
 			return schoolsModule.schoolIsExternallyManaged;
 		},
-		env() {
-			return envConfigModule.getEnv;
+		isConsentNecessary() {
+			return (
+				envConfigModule.getEnv &&
+				envConfigModule.getEnv.FEATURE_CONSENT_NECESSARY
+			);
 		},
 		showConsent() {
-			return this.env && this.env.ADMIN_TABLES_DISPLAY_CONSENT_COLUMN;
+			return (
+				envConfigModule.getEnv &&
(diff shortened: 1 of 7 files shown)


src/pages/administration/StudentOverview.page.vue
```diff
@@ -398,10 +398,7 @@ export default {
 			};
 		},
 		isConsentNecessary() {
-			if (process.env.SC_THEME === "brb") {
-				return false;
-			}
-			return true;
+			return this.env.FEATURE_CONSENT_NECESSARY;
```
virgilchiriac, line 401: why calling it with this.env, while above envConfigModule
in some other cases, I see being called like envConfigModule.getEnv.FEATURE_CONSENT_NECESSARY

odalys-dataport: Since envConfigModule.getEnv is already stored in "env" I just went with the flow that was already in place. I can change it though.

MartinSchuhmacher (dismissed):
tests missing in src/store/env-config.unit.ts
also think about test in src/pages/administration/student-overview.unit.js

src/store/env-config.ts
```diff
@@ -42,6 +42,7 @@ export default class EnvConfigModule extends VuexModule {
 		SC_TITLE: "",
 		SC_SHORT_TITLE: "",
 		GHOST_BASE_URL: "",
+		FEATURE_CONSENT_NECESSARY: true,
```
virgilchiriac, line 45: the logic in a bit hard to follow. So here we set a default in nuxt, which is then overwritten by the value we get from server, right?

odalys-dataport: This confused me as well, this is due to typescript complaining about it not being declared

virgilchiriac (approved):

PR #2254, opened 2022-08-12
BC-2218 - adjust bulk options

christian-darsow:
# Short Description
Remove bulk option "Analoge Einverständniserklärung" on student overview page (administration) if consent is not necessary

https://ticketsystem.dbildungscloud.de/browse/BC-2218

## Checklist before merging

### DIFF
--- a/src/pages/administration/student-overview.unit.js
+++ b/src/pages/administration/student-overview.unit.js
@@ -579,26 +579,22 @@ describe("students/index", () => {
 		});
 		expect(envConfigModule.getEnv.FEATURE_CONSENT_NECESSARY).toBe(false);
 
-		// user row exists
 		const dataRow = wrapper.find(`[data-testid="table-data-row"]`);
 		expect(dataRow.exists()).toBe(true);
-		// user row checkbox is clicked
+
 		const checkBox = dataRow.find(".select");
 		expect(checkBox.exists()).toBe(true);
+
 		await checkBox.trigger("click");
-		// user is selected
 		expect(dataRow.vm.selected).toBe(true);
 
-		// selection row component is rendered
 		const selectionBar = wrapper.find(".row-selection-info");
 		expect(selectionBar.exists()).toBe(true);
 
-		// contextMenu is rendered
 		const openContextButton = wrapper.find(".context-menu-open");
 		expect(openContextButton.exists()).toBe(true);
-		// contextMenu is clicked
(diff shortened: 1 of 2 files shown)


src/pages/administration/student-overview.unit.js
```diff
@@ -564,4 +564,42 @@ describe("students/index", () => {
...
+		});
+		const wrapper = mount(StudentPage, {
+			...createComponentMocks({
+				i18n: true,
+				store: mockStore,
+			}),
+		});
+		expect(envConfigModule.getEnv.FEATURE_CONSENT_NECESSARY).toBe(false);
+
+		// user row exists
```
MartinSchuhmacher, line 582: I would suggest removing the comments here as the code should be structured and named in a way that speaks for itself (which it does in my opinion)

MartinSchuhmacher: foremost reason is that comments can get deprecated quite quickly once logic changes and in case something changes, the comments need to be renewed and nobody does it because people forget ^^

christian-darsow: I agree, that those are actually not needed and will remove them at least in the new test. Similar comments will remain - or we might just remove those comments everywhere (also in former tests) at least in this file.

MartinSchuhmacher (approved):

src/pages/administration/student-overview.unit.js
```diff
@@ -564,4 +564,38 @@ describe("students/index", () => {
...
+			ADMIN_TABLES_DISPLAY_CONSENT_COLUMN: true,
+			FEATURE_CONSENT_NECESSARY: false,
+		});
+		const wrapper = mount(StudentPage, {
+			...createComponentMocks({
+				i18n: true,
+				store: mockStore,
+			}),
+		});
+		expect(envConfigModule.getEnv.FEATURE_CONSENT_NECESSARY).toBe(false);
```
virgilchiriac, line 580: do we have a positive test, aka when FEATURE_CONSENT_NECESARY = true, then the action button should be visible?

christian-darsow: Nope, I will add a test that checks for presence but also for action triggering, for some reason this is missing only on the consent bulk button

christian-darsow: As the consent action takes place after redirect, I now only added a test now for necessary button visibilty

PR #2273, opened 2022-09-12
BC-2362 - show alert in top bar

virgilchiriac:
# Short Description

BC-2362
https://github.com/hpi-schul-cloud/dof_app_deploy/pull/359

## Checklist before merging


src/components/legacy/TheTopBar.unit.js
```diff
@@ -1,4 +1,9 @@
 import TheTopBar from "./TheTopBar";
```
odalys-dataport, line 1: Please remove comments, if they aren't necessary

src/components/molecules/StatusAlerts.vue
```diff
@@ -0,0 +1,134 @@
...
+							{{ getUrl(item.url) }}
+						</a>
+					</div>
+				</v-card-text>
+			</div>
+		</v-card>
+	</div>
+</template>
+
+<script>
```
odalys-dataport, line 52: If this is a new component, it should be implemented with typescript and the composition api

src/plugins/datetime.js
```diff
@@ -261,6 +261,10 @@ export const fromNowToFuture = (date, unit) => {
 	} else return null;
 };
 
+export const timeAgo = (date) => {
```
odalys-dataport, line 264: This looks like it's doing the same or something similar like the function `fromNow` in this plugin, so maybe we can refactor this?

src/utils/store-accessor.ts
```diff
@@ -39,18 +41,19 @@ function initializeStores(store: Store<any>): void {
...
 	importUsersModule = getModule(ImportUsersModule, store);
 	newsModule = getModule(NewsModule, store);
+	notifierModule = getModule(NotifierModule, store);
 	roomModule = getModule(RoomModule, store);
 	roomsModule = getModule(RoomsModule, store);
 	schoolsModule = getModule(SchoolsModule, store);
+	statusAlertsModule = getModule(StatusAlertsModule, store);
 	taskModule = getModule(TaskModule, store);
-	notifierModule = getModule(NotifierModule, store);
-	copyModule = getModule(CopyModule, store);
```
odalys-dataport, line 53: Thanks for keeping this sorted! :)

odalys-dataport (changes requested):

src/plugins/datetime.js
virgilchiriac: is indeed a method of datejs, but doing the opposite. Can be refactored so that it looks the same and taking account the UTC convertion

tofixx: merged the bahaviour into fromNow. however "to" and "fromNow" seems to work each in both directions (past/future). just kept using local time, added this as new flag in fromNow

src/plugins/datetime.js
```diff
@@ -240,9 +240,13 @@ export const createInputDateTime = (date) => {
 /**
  * Returns date difference to current local time
  * @param {String} date
+ * @param {boolean} isLocalTimeZone set true if input date is to be handled in local time zone instead of utc
  * @return {String} Date difference based on current timezone
  */
-export const fromNow = (date) => {
+export const fromNow = (date, isLocalTimeZone) => {
+	if (isLocalTimeZone == true){
```
tofixx, line 247: fromNow and daysjs.to seems to have same behaviour, just checked here for local time to skip utc here

murat-merdoglu-hpi:
The implementation should be the 'mobile first' design approach.

image

src/components/legacy/TheTopBar.unit.js
```diff
@@ -69,4 +87,33 @@ describe("@components/legacy/TheTopBar", () => {
 		expect(wrapper.findAll(".top-sidebar")).toHaveLength(0);
 		expect(wrapper.findAll(".fullscreen-button-active")).toHaveLength(1);
 	});
+
+	it("render with Status Alerts", async () => {
+		const wrapper = shallowMount(TheTopBar, {
```
murat-merdoglu-hpi, line 92: Try to avoid using the `shallowMount` method. `shallowMount`method doesn't render its child components. But the newly created component `StatusAlert.vue` is its child.

murat-merdoglu-hpi (approved):

src/store/status-alerts.ts
```diff
@@ -0,0 +1,58 @@
...
+	setStatusAlerts(statusAlerts: StatusAlert[]): void {
+		this.statusAlerts = statusAlerts;
+	}
+
+	get getStatusAlerts(): StatusAlert[] {
+		return this.statusAlerts;
+	}
+
+	@Action
+	async fetchStatusAlerts() {
```
OliverHappe, line 47: missing return type void

src/components/legacy/TheTopBar.vue
```diff
@@ -9,67 +8,90 @@
 			<base-button
 				:class="{ 'menu-button': true, 'expanded-menu': expandedMenu }"
 				design="text icon"
+				data-test-id="top-menu-btn"
```
OliverHappe, line 11: i think this should be data-testid

src/components/legacy/TheTopBar.vue
```diff
@@ -9,67 +8,90 @@
...
+				<popup-icon
+					v-if="showStatusAlertIcon"
+					source="fa"
+					icon="exclamation-triangle"
+					:title="$t('global.topbar.actions.alerts')"
+					:aria-label="$t('global.topbar.actions.alerts')"
+					:fill="statusAlertColor"
+					class="item"
+					centered
+					data-test-id="status-alerts-icon"
```
OliverHappe, line 26: data-testid

src/components/legacy/TheTopBar.vue
```diff
@@ -130,21 +146,41 @@ export default {
...
+		return {
+			statusAlerts: [],
+		};
 	},
 	computed: {
 		role() {
 			const roleName = this.user.roles.map((r) => r.name);
 			return this.$t(`global.topbar.roleName.${roleName[0]}`);
 		},
+		showStatusAlertIcon() {
```
OliverHappe, line 158: show[...] suggest to me that this function is a toggle/ setter to actually show the statusalert-icon.

i suggest to instead call it "isShowStatusAlertIcon" to clearly convey that this is a getter describing a specific state of the component.

src/components/legacy/TheTopBar.vue
```diff
@@ -130,21 +146,41 @@ export default {
...
+			return this.statusAlerts.length !== 0;
+		},
+		statusAlertColor() {
+			const color = this.statusAlertDanger()
```
OliverHappe, line 162: i think statusAlertDanger() is a superflous codesplit. 

suggestion:

`const isDangerStatusAlert: boolean = <body of statusAlertDanger()>`

use isDangerStatusAlert as a condition in the `color` assignment. 

remove statusAlertDanger() method.

src/components/legacy/TheTopBar.vue
```diff
@@ -130,21 +146,41 @@ export default {
...
+				? "var(--color-error-base)"
+				: "var(--color-secondary-darken1)";
+			return color;
```
OliverHappe, line 165: i think the color variable is superflous - we can return the result immediately without assigning it first

src/components/molecules/StatusAlerts.unit.ts
```diff
@@ -0,0 +1,61 @@
...
+		...createComponentMocks({
+			i18n: true,
+			vuetify: true,
+		}),
+		propsData: props,
+		...options,
+	});
+};
+
+describe("@components/molecules/StatusAlerts", () => {
```
OliverHappe, line 23: a test to check if nothing is rendered if there are no statusalerts would be a nice addition

src/components/molecules/StatusAlerts.vue
```diff
@@ -0,0 +1,139 @@
...
+		width: 400px;
+	}
+}
+
+.action-button {
+	color: var(--color-primary);
+}
+
+.alert-item:not(:first-child) {
+	padding-top: var(--space-xs-4);
```
OliverHappe, line 112: please use the vuetify spacing utility classes. e.g. `pt-2`

src/components/molecules/StatusAlerts.vue
```diff
@@ -0,0 +1,139 @@
...
+		class="alerts"
+		data-testid="status-alerts"
+	>
+		<div
+			v-for="(item, index) in statusAlerts"
+			:key="index"
+			class="alert-item"
+			:data-testid="`alert-item-${index}`"
+		>
+			<v-card-text class="py-2">
```
OliverHappe, line 14: the parent div also has padding. i suggest to merge the two elements if possible. i don't see a reason for the outer div to wrap v-card-text.

src/components/molecules/StatusAlerts.unit.ts
odalys-dataport: this is tested in the top bar

OliverHappe: i think it's important to view one component as a relevant unit when writing unit tests. 

desired behavior should not only be tested in components outside of that unit. the topbar should only test if StatusAlert component exists. Behavior of StatusAlert should be tested in StatusAlert unit tests

odalys-dataport: aah ok, done :)

OliverHappe (approved):

PR #2282, opened 2022-09-30
BC-2528 - enable vuetify everywhere

odalys-dataport:
# Short Description

## Checklist before merging


src/themes/base/vuetify.options.js
```diff
@@ -14,6 +14,9 @@ import langIconUa from "@assets/icons/lang-icon-ua.vue";
 
 export default {
 	theme: {
+		options: {
```
uidp, line 17: Why is that moved here from nuxt.config.js?

src/layouts/default.vue
```diff
@@ -1,17 +1,25 @@
 <template>
 	<legacy-logged-in>
-		<base-content-container id="main-content" tag="main" class="content">
-			<Nuxt />
-		</base-content-container>
+		<v-app class="content">
```
uidp, line 3: Why are `class` and `id` split up to `<v-app>` and `<v-main>`?

src/layouts/default.vue
```diff
@@ -1,17 +1,25 @@
 <template>
 	<legacy-logged-in>
```
uidp, line 2: It looks like `<legacy-logged-in>` is only used here (in this component). Maybe we can refactor that a little differently. Also `<v-app>` is not the root component of the whole layout. Can it be achieved somehow that `<v-app>` wraps the topbar and sidebar as well?

uidp (commented):
I would suggest that we look in how we can build a more general default layout that wraps everything in `<v-app>`. Although I'm not sure if that's possible at the current state of things.

src/themes/base/vuetify.options.js
odalys-dataport: Because it wasn't working in the nuxt.config.js, moving this here made vuetify css vars globally accessible

src/layouts/default.vue
odalys-dataport: That's how it was before in `defaultVuetify`. But I can look into it while trying to make `<v-app>` the root component

uidp (dismissed):

christian-darsow (approved):

PR #2284, opened 2022-10-04
BC-2426 - theme and colors

christian-darsow:
# Short Description

Define colors per theme only via vuetify settings. 
Map old color vars to vuetify color system and remove colors.scss.

https://ticketsystem.dbildungscloud.de/browse/BC-2426

## Checklist before merging


src/themes/base/styles/utility/_colors.scss
```diff
@@ -1,12 +1,12 @@
 // apply to toasts
```
odalys-dataport, line 1: Can we move these styles to toast component itself?

odalys-dataport (changes requested):

src/themes/base/styles/utility/_colors.scss
christian-darsow: Lets please investigate on this in a follow up ticket

src/themes/base/styles/vuetify-custom.scss
```diff
@@ -52,7 +52,7 @@ $btn-focus-opacity: 0;
 
 .v-btn.primary {
 	&:hover {
-		background-color: var(--color-primary-dark) !important;
+		background-color: var(--v-primary-darken1) !important;
```
odalys-dataport, line 55: Can we use this instead? I'm not sure, but if it works it would be prettier

[$btn-states](https://vuetifyjs.com/en/api/v-btn/#sass-btn-states)	
```(
  'states': (
    'activated': $btn-active-opacity,
    'focus': $btn-focus-opacity,
    'hover': $btn-hover-opacity
  )
);```

File: src/themes/base/vuetify.options.js [hunk 3]
```diff
@@ -14,15 +14,38 @@ import langIconUa from "@assets/icons/lang-icon-ua.vue";
...
-				accent: "#E98404",
-				error: "#FF1134",
-				info: "#0A7AC9",
-				success: "#13BA98",
-				warning: "#FF8311",
+				black: {
+					base: "#1b1b1b",
+				},
+				white: {
+					base: "#ffffff",
```
odalys-dataport, line 26: white is set to #ffffff by vuetify itself, why are we overwriting this?

src/themes/base/vuetify.options.js
```diff
@@ -14,15 +14,38 @@ import langIconUa from "@assets/icons/lang-icon-ua.vue";
...
+				},
+				gray: {
+					base: "#aaaaaa",
+					lighten1: "#eeeeee",
+					lighten2: "#e0e0e0",
+					darken1: "#444444",
```
odalys-dataport, line 32: Are these discussed with UX? Maybe we can use these colors provided by vuetify instead. The only big difference would be in our base gray (#aaaaaa) and vuetify's base grey (#9e9e9e)

grey lighten-3 -> #EEEEEE
grey lighten-2 -> #E0E0E0
grey darken-3 -> #424242 (veeery similar)

https://vuetifyjs.com/en/styles/colors/#material-colors

src/themes/brb/vuetify.options.js
```diff
@@ -6,9 +6,18 @@ const custom = {
 		dark: false,
 		themes: {
 			light: {
-				primary: "#0a9396",
-				secondary: "#38677a",
-				accent: "#E4032E",
+				primary: {
+					base: "#0a9396",
+					darken1: "#086e71",
```
odalys-dataport, line 11: Just out of curiosity, does defining these mean we "lose" all other variants like primary-darken3?

src/themes/base/vuetify.options.js
```diff
@@ -14,15 +14,38 @@ import langIconUa from "@assets/icons/lang-icon-ua.vue";
...
+		},
 		themes: {
 			light: {
-				primary: "#9e292b",
-				secondary: "#54616e",
-				accent: "#E98404",
-				error: "#FF1134",
-				info: "#0A7AC9",
-				success: "#13BA98",
-				warning: "#FF8311",
```
odalys-dataport, line 25: Why are primary, secondary and accent color being removed being removed? I see they are defined in default theme. I think I'd prefer it if they were defined in base theme and "inherited" by default theme, if that's possible, but I have no strong opinion on this.

odalys-dataport (changes requested):

src/themes/brb/vuetify.options.js
christian-darsow: Yeap, if we define the color vars in this way, vuetify "only" generates the defined ones. So lose darken2, darken3, ...

src/themes/base/vuetify.options.js
christian-darsow: Primary, secondary and accent are only valid for the default theme (and different for brb, n21, ...). In my opion it's therefore more explicit and clearer to define them there. all values in base are in this way valid for all themes --- well, the base :)

src/themes/base/vuetify.options.js
christian-darsow: Yes, all values are discussed with UX, we also had tried other gray values but then decided to stay with the values we currently use

src/themes/base/vuetify.options.js
christian-darsow: I did that just to have a color var defined in the same way, we can use all other colors. var(--v-white-base)

src/themes/base/styles/vuetify-custom.scss
christian-darsow: I agree, I like this style! I didnt know that sass var before, but it looks like a way to define opacity values for all buttons at once and not define bg color only for eg primary button. If you have any idea of how to map that, let's do it!

src/themes/base/vuetify.options.js
odalys-dataport: Could we run into issues with not having a fall back color defined in base in case the other themes fail to load properly in any way?

src/themes/base/vuetify.options.js
odalys-dataport: Alright, then maybe let's find a better name for our own color, because having "grey" and "gray" could be confusing and cause human errors in the future

src/themes/brb/vuetify.options.js
odalys-dataport: We could think about setting this option then, if we don't need variations and/or creating them ourselves

```
theme: {
    options: { variations: false },
  },
  ```

src/themes/base/vuetify.options.js
christian-darsow: I would say a missing or miss-configuration of a theme is always an issue. I don't have a strong opinion on what wrong color is worse, vuetify fallback or other-instance fallback. If you say fallback to the default instance is better, lets reintroduce it

monaweitzenberg (approved):
hex-values looking fine

src/themes/base/vuetify.options.js
christian-darsow: very good point. I would suggest to then call it "grey" and overwrite the material values.

odalys-dataport (approved):

PR #2291, opened 2022-10-17
BC-2529 - global css

christian-darsow:
# Short Description

Reduce build css size by using global css

https://ticketsystem.dbildungscloud.de/browse/BC-2529

## Checklist before merging


src/pages/RoomDetails.page.vue
```diff
@@ -397,7 +397,6 @@ export default defineComponent({
 </script>
 <style lang="scss" scoped>
 @import "~vuetify/src/styles/styles.sass";
```
OliverHappe, line 399: i think we should not do this. 

in this component we need the breakpoint mixin from vuetify.

i suggest to use the breakpoint display classes in the template in favor of importing all vuetify styles just for a mixin.

alternatively: is it possible to just import the mixin?

christian-darsow: this vuetify import could be removed from some components complety as none of the imported functionality was used. for the others I would say we can stay with this import, as it only imports mixins / vars and nothing that gets compiled into the css. I would say that this is also with the docs https://vuetifyjs.com/en/features/sass-variables/#usage-in-templates

src/pages/RoomDetails.page.vue
```diff
@@ -397,7 +397,6 @@ export default defineComponent({
...
-@import "@variables";
```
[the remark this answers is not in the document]
christian-darsow: goid point. unneeded imports are removed from several components now

src/themes/base/styles/index.scss
```diff
@@ -1,2 +1,13 @@
-@import "./utility.scss";
```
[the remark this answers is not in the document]
christian-darsow: the index.scss from the themes now import it as mixins from the base

src/pages/RoomDetails.page.vue
OliverHappe: thanks for the clarification!

odalys-dataport (approved):

PR #2298, opened 2022-10-27
BC-2603 - task create

odalys-dataport:
# Short Description

BC-2603
https://github.com/hpi-schul-cloud/schulcloud-server/pull/3694

## Checklist before merging


src/utils/store-accessor.ts
```diff
@@ -37,6 +38,7 @@ let roomModule: RoomModule;
 let roomsModule: RoomsModule;
 let schoolsModule: SchoolsModule;
 let statusAlertsModule: StatusAlertsModule;
+let tasksModule: TasksModule;
```
virgilchiriac, line 41: shoudl this be singluar, is for Task?

MartinSchuhmacher:
I still am unsure about this whole task/tasks separation to be honest, even though we do the same for room/rooms, which does not mean we should continue with doing so. I find it very confusing actually (which is the same with room/rooms), so maybe we can talk about that.

src/pages/TaskCreate.page.vue
```diff
@@ -0,0 +1,60 @@
...
+
+		const breadcrumbs = [
+			{
+				text: i18n.t("common.words.tasks"),
+				to: "/tasks",
+			},
+		];
+
+		const save = () => {
+			taskModule.createTask({
```
christian-darsow, line 42: I'm also getting confused now :) While I see good reasons to have two stores for single/multiple I'm wondering if this more generic action should not be in the tasks store (like the deleteTask action)

MartinSchuhmacher: agree to that

src/store/task.ts
```diff
@@ -0,0 +1,110 @@
...
+		if (!this._taskApi) {
+			this._taskApi = TaskApiFactory(undefined, "/v3", $axios);
+		}
+		return this._taskApi;
+	}
+
+	@Action
+	async createTask(params: TaskCreateParams): Promise<void> {
+		// check for permission here or in page?
+		// probably checked in server?
```
virgilchiriac, line 42: server authorization is done in any case, the logic should therefor be more on the level of page load / show button

MartinSchuhmacher (approved):

PR #2300, opened 2022-11-01
BC-2409 - vue3 i18n

christian-darsow:
# Short Description

Vue 3 POC: Implement I18n

https://ticketsystem.dbildungscloud.de/browse/BC-2409

## Checklist before merging

### DIFF
--- a/vue3_poc/package.json
+++ b/vue3_poc/package.json
@@ -19,11 +19,13 @@
 		"roboto-fontface": "*",
 		"universal-cookie": "^4.0.4",
 		"vue": "^3.2.37",
+		"vue-i18n": "^9.2.2",
 		"vue-router": "^4.1.2",
 		"vuetify": "^3.0.0-beta.0",
 		"webfontloader": "^1.0.0"
 	},
 	"devDependencies": {
+		"@intlify/unplugin-vue-i18n": "^0.7.0",
 		"@mdi/js": "^7.0.96",
 		"@types/express": "^4.17.14",
 		"@types/webfontloader": "^1.0.0",
--- a/vue3_poc/src/components/LanguageSelector.spec.ts
+++ b/vue3_poc/src/components/LanguageSelector.spec.ts
@@ -32,14 +32,11 @@ describe("LanguageSelector", () => {
 
 	it("should change language from de to en when en btn clicked", () => {
 		const wrapper = getWrapper();
-		const selectLanguageSpy = vi.spyOn(wrapper.vm, "selectLanguage");
 		const englishBtn = wrapper.find(`[data-test-id="changeLanguageToEnglish"]`);
 
 		expect(i18n.global.locale).toStrictEqual("de");
 
 		englishBtn.trigger("click");
-
-		expect(selectLanguageSpy).toHaveBeenCalled();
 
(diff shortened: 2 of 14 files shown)


vue3_poc/src/store/types/Language.ts
```diff
@@ -0,0 +1,4 @@
+export enum Language {
+	English = "en",
+	German = "de",
```
virgilchiriac, line 3: why not going with all 4 langs already?

christian-darsow: Actually only because we focused in that POC ticket on en + de (and also the lang switch therefore only provides those two options etc)

vue3_poc/package.json
```diff
@@ -19,11 +19,13 @@
 		"roboto-fontface": "*",
 		"universal-cookie": "^4.0.4",
 		"vue": "^3.2.37",
+		"vue-i18n": "^9.2.2",
 		"vue-router": "^4.1.2",
 		"vuetify": "^3.0.0-beta.0",
 		"webfontloader": "^1.0.0"
 	},
 	"devDependencies": {
+		"@intlify/unplugin-vue-i18n": "^0.7.0",
```
uidp, line 28: What exactly are the advantages of using this plugin? Do we really want to have that now at this point?

vue3_poc/src/components/LanguageSelector.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+describe("LanguageSelector", () => {
+	it("renders properly", () => {
+		const wrapper = getWrapper();
+		expect(wrapper.text()).toContain("Deutsch");
+	});
+
+	it("should change language from de to en when en btn clicked", () => {
+		const wrapper = getWrapper();
+		const selectLanguageSpy = vi.spyOn(wrapper.vm, "selectLanguage");
```
uidp, line 35: The method `selectLanguage` belongs to the internal implementation of the component and shouldn't be tested IMHO. We should only test the external behaviour of a component. So checking that `i18n.global.locale` gets altered by the click should be enough.

vue3_poc/src/plugins/vuetify.ts
```diff
@@ -26,5 +30,11 @@ export default createVuetify({
 			mdi,
 		},
 	},
+	// https://next.vuetifyjs.com/en/features/internationalization/#getting-started
+	locale: {
```
uidp, line 34: Why do we have to do that again seperately although we have the language config in `plugins/i18n.ts` already?

uidp (changes requested):

vue3_poc/package.json
christian-darsow: While probably not needed for the poc, this plugin takes care of proper build configuration. In the end it reduces bundle size and therefore improves performance. If you think it might be better (clearer) to not have such things now, lets remove it and come back to this docs later https://vue-i18n.intlify.dev/guide/advanced/optimization.html

vue3_poc/src/components/LanguageSelector.spec.ts
christian-darsow: Thanks for that :) Makes absolutely sense to me and I changed that test

uidp (approved):

PR #2306, opened 2022-11-03
N21-411 improve sidebar highlighting

MarvinOehlerkingCap:
- use regex to activate the highlighting on sidebar items instead of string.includes(). this prevents collisions e.g. in subpaths like /files/teams and /teams
- introduce type for sidebar items

# Short Description

https://ticketsystem.dbildungscloud.de/browse/N21-411

## Checklist before merging


arnegns (approved):

src/utils/sidebarBaseItems.ts
```diff
@@ -1,23 +1,42 @@
-export default [
+export type SidebarItem = {
+	title: string;
+	href?: string;
```
OliverHappe, line 3: i suppose a SidebarItem without href is a parent of a SidebarItem group and requires children?

I appreciate the effort to switch to typescript and create types. :) 

I suggest to create several interfaces if we want to achieve a better typing. Also I suggest to try to avoid optional properties whenever possible. 

Interface suggestion: 

```ts
export type SidebarItem = SidebarItemLink & SidebarItemCategory;

type SidebarItemBase = {
    title: string;
    icon: string;
    to?: string;
    testId?: string; // should be required
    permission?: string;
    excludedPermission?: string;
    /**
     *  should be required -> an empty array serves the same purpose than undefined but 
     *  avoids errors (or conditional) when using array functions e.g. Array.includes()
     */
    activeForUrls?: string[]; // should be required and an empty array if no values are present
}

export type SidebarItemLink = SidebarItemBase & {
    href: string;
}

// I am open for better name suggestions :)
export type SidebarItemCategory = {
    children: [];
}

```

The main type to be used is still SidebarItem which covers all options in the tree of SidebarItems. In Subcomponents 
the distinction between SidebarItemLink and ...Category give us the advantage to require less Typechecking for existing properties because based on the component context we could type a `prop` to be exactly SidebarItemLink.

I know that TS support in components is lackluster at the moment but in the future this will give us a great benefit for a small effort today :)

**Note**
Personally i have no preferance among `type` or `interface`. I think that Interfaces are easier to understand for people who are used to OOP (in the example above it would read as `export interface SidebarItemCategory extends SidebarItemBase`) but i want to leave the decision between the two up to you :)

src/layouts/legacyLoggedIn.vue
```diff
@@ -95,25 +95,26 @@ export default {
...
-					(item.activeForUrls && item.activeForUrls.some(isInRoute));
+			// Sidebar highlights
+			const { path } = this.$route;
+			const isActive = (item) =>
+				item.activeForUrls &&
+				item.activeForUrls.some((activeFor) =>
+					new RegExp(activeFor).test(path)
+				);
+
+			sidebarItems.forEach((item) => {
```
OliverHappe, line 106: I have to say that I am not too happy about this function.

1. It adds properties to the object that do not exist in the Type you defined. Unfortunately TS support in components is basically non-existant at the moment so TS does not raise an issue here, but there will be once we manage to enable TS. 
2. I think the functions that are added muddle the distinction between configurational data for our view and the behavior of the view. The Sidebar should handle which item are active or not since it is only concerning the view of the sidebar. It should not be part of the data we use to configure the sidebar.

**Suggestion**
 * remove the business logic regarding `isActive` from this component. Pass in the SidebarItems as they are defined in `SidebarItems.ts` 
 * pass in the currently active url `this.$route.path` to the sidebar (this way the sidebar stays a dumb component)
 * create a computed in sidebar which returns the currently active sidebar items (e.g. return string[] of their titles) using your improved isActive function
 * check whether an Item is active by looking it up in the activeItems-Array

**Note**
I know this is quite an extensive refactor of this component. I'd like to make clear that I will approve this PR when you decide against implement my suggestion. If you don't want to implement them, please consider creating a Ticket to do it next Sprint? I really appreciate that you took the time to refactor this component already. :)

src/layouts/legacyLoggedIn.vue
```diff
@@ -95,25 +95,26 @@ export default {
...
-				const isActive =
-					this.$route.path.includes(item.href) ||
-					this.$route.path.includes(item.to) ||
-					(item.activeForUrls && item.activeForUrls.some(isInRoute));
+			// Sidebar highlights
+			const { path } = this.$route;
+			const isActive = (item) =>
+				item.activeForUrls &&
+				item.activeForUrls.some((activeFor) =>
+					new RegExp(activeFor).test(path)
```
OliverHappe, line 103: much better than before, thanks!

src/utils/sidebarBaseItems.ts
```diff
@@ -1,23 +1,42 @@
-export default [
```
OliverHappe, line 1: please use kebab-case for filenames

src/utils/sidebarBaseItems.ts
MarvinOehlerkingCap: I don't see a benefit of splitting the types for our current changes. This change should be done when the complete component is changed to TypeScript, because only then you know how the interfaces should be split exactly and how they should look like. Improvements to the data layout seem like a bigger refactoring.

src/layouts/legacyLoggedIn.vue
MarvinOehlerkingCap: I understand what you mean and I agree, but I don't think that your suggestion is optimal.

Using an isActive function which generates a simple string array comes with a loss in performance, since route.active and route.childActive are called often in the component and iterating over a list is suboptimal. It is also hard to do it with a simple string array, since you need the active and childActive property separately. There is the possibility to use a map like `{ "title": { active: boolean; childActive: boolean; } }`, but this is just unnecessarily complicated and you can just build the correct routes array with the active information, that is used in the template section.

The component would also not be that dumb anymore when using the computed to determine, which item is active. Its logic is related to the currently visited route and not any other route, so I don't see a problem/reason to not use the state information.

The `SidebarItems.ts` can also not be directly passed, since the items are filtered by permissions.

My Suggestion would be to only allow one active parent and one active child. Everything else would not make sense to me. I would use two functions `isActive` and `isChildActive` and two variables that hold the the currently active title and active parent title.

src/utils/sidebarBaseItems.ts
OliverHappe: I agree that the benefit is marginal considering that TS support in the components is broken. I prefer the sentiment of "refactor now over refactor later". Especially considering that TS support will be working in a few weeks after the nuxt removal pr is merged.

src/layouts/legacyLoggedIn.vue
OliverHappe: As far as I understand your suggestion I think it also solves the two problems I pointed out about the sidebar so please go ahead :)

OliverHappe (dismissed):

OliverHappe (dismissed):

PR #2307, opened 2022-11-04
N21-402 implement collaborative files page

arnegns:
# Short Description
This PR contains the first implementation of the new collaborative files page.

https://ticketsystem.dbildungscloud.de/browse/N21-221

## Screenshots of UI changes
image
image

## Checklist before merging


src/store/files.ts
```diff
@@ -0,0 +1,159 @@
...
+import { FileTypeResponse } from "@/serverApi/v3/files/file-meta.response";
+import { File, FileType } from "@store/types/file";
+import { filesModule } from "@utils/store-accessor";
+
+@Module({
+	name: "files",
+	namespaced: true,
+	stateFactory: true,
+})
+export default class FilesModule extends VuexModule {
```
uidp, line 13: The name `FilesModule` is quite general for just the collaborative files. Can we find a more specific name an keep the general name for general files when we need it?

src/store/types/file.ts
```diff
@@ -0,0 +1,60 @@
+export enum FileType {
```
uidp, line 1: Naming? See above.

uidp (changes requested):

src/pages/files/FilesOverview.page.vue
```diff
@@ -1,6 +1,140 @@
...
+	},
+	setup() {
+		const { t } = i18n();
+		const { path, params } = useRoute().value;
+
+		const title: Ref<string> = ref(t("pages.files.overview.headline"));
+
+		const breadcrumbs: Ref<Breadcrumb[]> = ref([]);
+
+		const headers: DataTableHeader[] = [
```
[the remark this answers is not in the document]
arnegns: Done

src/components/base/BaseTable/BaseTable.vue
```diff
@@ -0,0 +1,76 @@
+<template>
```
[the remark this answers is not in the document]
Metauriel: It probably makes sense to use the vuetify component. maybe look at the task overview as an example

src/utils/i18n-util.ts
```diff
@@ -0,0 +1,26 @@
+import VueI18n, { Locale } from "vue-i18n";
+import { inject } from "@vue/composition-api";
+import { ChangeLanguageParamsLanguageEnum } from "@/serverApi/v3";
+
+export type TranslationFunction = (key: string) => string;
+
+export function i18n() {
```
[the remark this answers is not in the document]
arnegns: We removed our wrapper for that

src/pages/files/FilesOverview.page.vue
```diff
@@ -1,6 +1,140 @@
...
+
+		switch (currentCategory) {
+			case "cfiles": {
+				title.value = t("pages.files.overview.headline");
+
+				loadFilesFunc = async () => filesModule.fetchFilesOverview();
+				break;
+			}
+			case "teams": {
+				if (pathArray.length == 2) {
```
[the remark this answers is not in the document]
arnegns: We have moved all functions to a file like this

src/serverApi/v3/files/file-meta-list.response.ts
```diff
@@ -0,0 +1,8 @@
+import { FileMetaResponse } from "@/serverApi/v3/files/file-meta.response";
```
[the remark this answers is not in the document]
arnegns: Allright :) We have moved the files to another place. Thanks for the advice

src/store/files.ts
arnegns: We named the module files module because it should be the general files module. As you can see in the screenshot the sidebar entry call "Dateien" and will replace "Meine Dateien".

src/utils/i18n-util.ts
```diff
@@ -0,0 +1,26 @@
...
+		const translateResult = i18nLib?.t(key);
+		if (typeof translateResult === "string") {
+			return translateResult;
+		}
+		return "unknown translation-key:" + key;
+	};
+
+	const locale = (): Locale => {
+		if (i18nLib?.locale === ChangeLanguageParamsLanguageEnum.Ua) {
+			return "uk"; // TODO https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes use everywhere uk (language code)
```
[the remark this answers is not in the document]
arnegns: I dont get what you mean with fallback.. The reason why we needs a mapping from UA to uk is because UA is the country code and uk the language code which is wrongly used in the bildungscloud. We wanted to change this in i18n.js but the codes also come from the server.

src/router/routes.js
```diff
@@ -152,4 +152,16 @@ export const routes = [
...
 		name: "files",
 	},
+	{
+		path: "/cfiles/teams",
+		component: () =>
+			interopDefault(import("@pages/files/FilesOverview.page.vue")),
+		name: "teamfiles",
+	},
+	{
+		path: "/cfiles/teams/:catchAll(.*)",
```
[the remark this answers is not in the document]
arnegns: It has to be that dynamic because we dont know how many folders exists in one team. Maybe we change everything behind the /teams/... to a query param

src/store/files.ts
arnegns: I thought it will replace it for every instance and not only for nds. UX designed a mockup for this.
But its okay for us, then we ll give an other name and then it can be renamed in the future.

src/store/types/file.ts
```diff
@@ -0,0 +1,60 @@
...
+const FileTypeIconMapping = {
+	[FileType.FILE]: FileIcon.FILE,
+	[FileType.PDF]: FileIcon.FILE,
+	[FileType.UNKNOWN]: FileIcon.FILE,
+	[FileType.DIRECTORY]: FileIcon.FOLDER,
+	[FileType.SHARED_DIRECTORY]: FileIcon.SHARED_FOLDER,
+	[FileType.FAVORITES]: FileIcon.FAVORITE,
+};
+
+export class File {
```
OliverHappe: please do not use classes. 

Always use `interface` or `type`.

please move this file to `/store/types` (a better project structure is currently being developed in the frontend arch group atm. let's use the old places even if they are unintuitive for easier refactoring later)

src/components/templates/default-wireframe.types.ts
```diff
@@ -0,0 +1,24 @@
+export type Breadcrumb = {
+	text: string;
+	href?: string;
+	to?: string;
+	disabled?: boolean;
```
OliverHappe, line 5: i htink boolean fields should never be undefined (optional) if we do not have to distinguish between the value being `false` and `undefined`.

src/components/templates/default-wireframe.types.ts
```diff
@@ -0,0 +1,24 @@
...
+};
+
+type FabAction = {
+	icon: String;
+	label: String;
+	href?: String;
+	to?: String;
+	dataTestid?: String;
```
OliverHappe, line 13: dataTestId should always be required. It makes writing tests (esp. e2e-tests) much easier if all elements have an ID

src/components/templates/default-wireframe.types.ts
```diff
@@ -0,0 +1,24 @@
...
+	class?: String;
+	testId?: String;
```
OliverHappe, line 15: what is the difference between testId and dataTestId? I suggest to pick one (preferably `testId`) if they cover the same functionality

src/components/templates/default-wireframe.types.ts
```diff
@@ -0,0 +1,24 @@
...
+	ariaLabel?: String;
+};
+
+export type Fab = {
+	actions?: FabAction[];
```
OliverHappe, line 20: actions should be required. 

instead of not specifying the property we should use an empty array. it gives us stricter typing in components and also does not warrant to check for the properties existance whenever we might want to use Array-Functions

src/store/files.ts
hoeppner-dataport: I see... I think "Files"-MenuItem will replace "My Files"-MenuItem... but this store accesses the collaborativeFiles or NextCloud-Files and should be named this way... and there is/might be another store, that accesses the fileStorage-Service...  and maybe both stores will be used on the files.page in the future.

src/pages/files/file-table-item.ts
```diff
@@ -0,0 +1,10 @@
+export interface FileTableItem {
+	name: string;
+	path: string;
+	icon: {
+		name: string;
+		colored?: boolean;
+	};
+	size: string;
+	lastChanged: Date;
```
OliverHappe, line 9: is this an actual Date instance in the data or just a date string?

We did not have this discussion in the arch group yet but i suggest to not use the Date-Class in interfaces since it is an actual Class instance which is not "clonable" usiong object spread.

Instead i suggest to use a timestamp string in the data. 

@uidp @hoeppner-dataport @odalys-dataport  what do you think?

src/pages/files/FilesOverview.page.vue
```diff
@@ -1,6 +1,129 @@
...
+
+		const locale = (): Locale => {
+			if (i18n.locale === ChangeLanguageParamsLanguageEnum.Ua) {
+				return "uk"; // TODO https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes use everywhere uk (language code)
+			}
+			return i18n.locale || "de";
+		};
+
+		const timesAgo = (value: Date): string => {
+			return moment(value).locale(locale()).fromNow();
```
OliverHappe, line 104: could you please verify that this function is called only on the initial render of the page and does not refresh constantly?

src/pages/files/FilesOverview.page.vue
```diff
@@ -1,6 +1,129 @@
...
+		};
+
+		const router: VueRouter = useRouter();
+		const click = function (item: FileTableItem): void {
```
OliverHappe, line 108: please use fat-arrow syntax like in line 101

src/pages/files/file-table-utils.ts
```diff
@@ -0,0 +1,138 @@
...
+import { Breadcrumb } from "@components/templates/default-wireframe.types";
+
+export type FilesPage = {
+	title: string;
+	breadcrumbs: Breadcrumb[];
+	loadFilesFunction: () => Promise<void>;
+};
+
+export function getHeaders(t: (key: string) => string): DataTableHeader[] {
+	return [
```
OliverHappe, line 18: I do not think that moving component related business logic to another TS file is a good practice in view. 

I suggest to either 

**A**

Keep these functions in the component. It is ok for Smart components to include business logic.

**B**

Create a [Composable](https://vuejs.org/guide/reusability/composables.html) covering this business logic. 

or **C**

create subcomponents for the different tabs

**Edit**: lets discuss tomorrow which way is the best for us. :)

------
keep in mind that everything that is exported here, currently is exported on a global scope within the whole project.

Functions should be named according to the scope they are relevant for. e.g. `getDeepBreadcrumbs` sounds like a quite generic implementation to create a breadcrumb format based on a string-Array. But in the implementation here it is clear that it is only supposed to be used in the context of the Files-Subpage. We have to careful with global exports (read: any exports) until a better project structure is in place.

src/pages/files/file-table-utils.ts
```diff
@@ -0,0 +1,138 @@
+import { File, FileType } from "@store/types/file";
+import { FileTableItem } from "@pages/files/file-table-item";
+import { filesModule } from "@utils/store-accessor";
+import { Route } from "vue-router";
+import { DataTableHeader } from "vuetify";
+import { Breadcrumb } from "@components/templates/default-wireframe.types";
+
+export type FilesPage = {
```
OliverHappe, line 11: I think FilesPage as a type name is a bit unfortunate. It can easily be confused with the type of a FilesPage component. 

I suggest `FilesPageConfig` since this closer describes what this object actually contains. Configurational data for the view of a FilesPage :)

src/pages/files/file-table-utils.ts
arnegns: As we discussed, we build a composable

src/pages/files/file-table-utils.ts
arnegns: Good suggestion :)

src/pages/files/FilesOverview.page.vue
arnegns: I verified this :)

src/pages/files/file-table-utils.composable.ts
```diff
@@ -0,0 +1,194 @@
...
+					text: t("pages.files.overview.headline"),
+					to: "/cfiles/",
+				},
+			],
+			loadFilesFunction: async () => collaborativeFilesModule.fetchTeams(),
+		};
+	}
+
+	function getTeamsPage(
+		t: (key: string) => string,
```
OliverHappe, line 104: can we think of usecases where it is useful to give different implementations of the translate function? 

if not it might be cleaner to supply t once to the composable and bind it in the composables internal scope.

src/pages/files/file-table-utils.composable.ts
```diff
@@ -0,0 +1,194 @@
...
+	FileMetaResponse,
+	FileTypeResponse,
+} from "@store/collaborative-files/file-meta.response";
+import { FilesPageConfig } from "@pages/files/file-page-config.type";
+import CollaborativeFilesModule from "@store/collaborative-files";
+
+export function fileTableComposable(
+	collaborativeFilesModule: CollaborativeFilesModule
+) {
+	function getHeaders(t: (key: string) => string): DataTableHeader[] {
```
uidp, line 21: Functions should be defined consistently using the arrow notation like `const doStuff = () => {}` and not like `function doStuff() {}`. We do not have a linter rule for that yet.

src/pages/files/file-table-utils.composable.ts
```diff
@@ -0,0 +1,194 @@
...
+import { Breadcrumb } from "@components/templates/default-wireframe.types";
+import { FileMetaListResponse } from "@store/collaborative-files/file-meta-list.response";
+import {
+	FileMetaResponse,
+	FileTypeResponse,
+} from "@store/collaborative-files/file-meta.response";
+import { FilesPageConfig } from "@pages/files/file-page-config.type";
+import CollaborativeFilesModule from "@store/collaborative-files";
+
+export function fileTableComposable(
```
OliverHappe, line 18: composable naming convention as suggested by vue is `use<FeatureName>` so i suggest to rename this to  `useFileTableUtils`

src/router/guards/permission.guard.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function createPermissionGuard(
+	permission: string,
+	fallbackRoute: string
+): NavigationGuard {
+	return (to: Route, from: Route, next: NavigationGuardNext) => {
+		if (authModule.getUserPermissions.includes(permission)) {
+			next();
+		} else {
+			next(fallbackRoute);
```
uidp, line 13: Shouldn't we display an error page instead of quietly redirecting to another page if the user doesn't have the right permissions to access a route?

uidp (changes requested):

src/components/templates/default-wireframe.types.ts
```diff
@@ -0,0 +1,24 @@
...
+	label: String;
+	href?: String;
+	to?: String;
+	dataTestid?: String;
+	class?: String;
+	testId?: String;
+	ariaLabel?: String;
+};
+
+export type Fab = {
```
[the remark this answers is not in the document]
arnegns: Honestly I cant answer these questions. 
In this case we only moved all types which were defined in the component. We only wanted to reuse the Breadcrumb. If you want we revert this change and only move the breadcrumb.

OliverHappe (approved):

src/pages/files/FilesOverview.page.vue
```diff
@@ -1,6 +1,130 @@
...
+
+		const route: Route = useRoute().value;
+		const filesPage: FilesPageConfig = getFilesPageForRoute(t, route);
+
+		const title: Ref<string> = ref(filesPage.title);
+		const breadcrumbs: Ref<Breadcrumb[]> = ref(filesPage.breadcrumbs);
+
+		const locale = (): Locale => {
+			if (i18n.locale === ChangeLanguageParamsLanguageEnum.Ua) {
+				return "uk"; // TODO https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes use everywhere uk (language code)
```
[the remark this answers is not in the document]
arnegns: I hope it is temporary but not because the translation is missing, more because the usage is wrong. Not only in client, also in the server

src/router/guards/permission.guard.ts
OliverHappe: i think that behavior is in development in our sprint right now? 

maybe we approve this routeguard and refactor it to use the new pattern for permission error pages once it is done?

src/pages/files/file-table-utils.composable.ts
arnegns: Thanks for the advice :)

OliverHappe (approved):

PR #2323, opened 2022-11-22
BC-2727 - Implement Element Cards for Edit Page (future card edit)

odalys-dataport:
# Short Description

## Checklist before merging


src/components/task-form/TaskContentElement.unit.ts
```diff
@@ -0,0 +1,53 @@
...
+		},
+		propsData: props,
+		...options,
+	});
+};
+
+// TODO Promise rejection - CKEditorError: bo.window.ResizeObserver is not a constructor
+describe("@components/task-form/TaskContentElement", () => {
+	it("should render component", () => {
+		const wrapper = getWrapper({ value: "abc" });
```
christian-darsow, line 22: Do we need this prop? also in the following tests?

src/components/task-form/TaskTitleElement.vue
```diff
@@ -0,0 +1,39 @@
...
+			type: String,
+			default: "",
+		},
+		value: {
+			type: String,
+			default: "",
+		},
+	},
+	setup(props) {
+		const name = ref(props.value);
```
christian-darsow, line 25: I would suggest to rename that to something more specific like 'title'

src/components/task-form/TaskTitleElement.vue
```diff
@@ -0,0 +1,39 @@
...
+
+		watch(
+			() => props.value,
+			(newValue) => {
+				name.value = newValue;
```
christian-darsow, line 30: After thinking about that, I'm not really sure, in which case we actually want or need to support this - which side effect can happen, that the prop changes?

src/pages/TaskForm.page.unit.ts
```diff
@@ -0,0 +1,105 @@
...
+				authModule.setUser(mockAuthStoreDataStudent);
+				expect($router.go).toHaveBeenCalled();
+				// QUESTION should not render TaskForm page?
+			});
+		});
+
+		describe("when user does have permission HOMEWORK_CREATE", () => {
+			it.todo("should do what?"); // QUESTION
+		});
+	});
```
christian-darsow, line 104: tests for delete element, add element will follow in the next iteration, right?

src/store/task.ts
```diff
@@ -9,17 +10,24 @@ import {
...
-	taskData: Object = {
-		taskId: "",
-		courseId: "",
-		title: "",
-		description: "",
+	// TODO - implement/use proper type
+	taskData: TempTask = {
+		id: "",
+		name: "",
+		description: { content: "", type: RichTextTypeEnum.RichtextCk5 },
```
christian-darsow, line 30: As I simply don't know: the API so far is not ready for our array of texts?

src/components/task-form/TaskTitleElement.vue
odalys-dataport: I agree, I've kept it like this for now because those are the property names from the backend and I didn't want to map them. I would change this when the new backend is ready.

src/components/task-form/TaskContentElement.unit.ts
odalys-dataport: If you mean the prop "value" I'd be happy to receive some help for a better solution

src/store/task.ts
odalys-dataport: I haven't received any news on being able to use the new backend yet

src/pages/TaskForm.page.unit.ts
odalys-dataport: Yes exactly :)

src/components/task-form/TaskContentElement.unit.ts
christian-darsow: I'm confused :) I don't see any prop "value" in that component. Is that a leftover from before, when it was the text component? if I make the props object empty, all tests still pass

src/components/task-form/TaskTitleElement.vue
christian-darsow: yeap, you're right. I'm still wondering if there is as case that the prop changes in the parent component at all - and that we need to watch for that? feels like an ugly side effect

src/components/task-form/TaskContentElement.unit.ts
odalys-dataport: Oh might be, I'll check

christian-darsow (approved):

PR #2373, opened 2023-01-24
N21-500 migration page

MarvinOehlerkingCap:
# Short Description
- add new public page for the migration flow https://docs.dbildungscloud.de/x/UADjDQ

https://ticketsystem.dbildungscloud.de/browse/N21-500

## Screenshots of UI changes
image

## Checklist before merging


IgorCapCoder (dismissed):

src/store/user-migration.ts
```diff
@@ -0,0 +1,72 @@
...
+
+			const mappedLinks: MigrationLinks = {
+				proceedLink: links.data.proceedButtonUrl,
+				cancelLink: links.data.cancelButtonUrl,
+			};
+			this.setMigrationLinks(mappedLinks);
+			this.setLoading(false);
+		} catch (e) {
+			console.log(`Some error occurred while loading migration links: ${e}`);
+			this.setLoading(false);
```
odalys-dataport, line 69: Please add error handling

src/store/systems.ts
```diff
@@ -0,0 +1,68 @@
...
+			const mappedSystems: System[] = systems.data.data.map(
+				(system: SystemResponse): System => ({
+					id: system.id,
+					name: system.displayName || "",
+				})
+			);
+			this.setSystems(mappedSystems);
+			this.setLoading(false);
+		} catch (e) {
+			console.log(`Some error occurred while loading system data: ${e}`);
```
odalys-dataport, line 64: Please add error handling

src/pages/user-migration/UserMigration.page.vue
```diff
@@ -0,0 +1,166 @@
+<template>
+	<div v-show="hasData" class="text-center mx-auto">
+		<img src="@assets/img/migration/migration.svg" alt="" />
```
odalys-dataport, line 3: Please add an alternative text for the image

src/pages/user-migration/UserMigration.page.vue
```diff
@@ -0,0 +1,166 @@
...
+		<h1 class="pl-4 pr-4">
+			{{ $t("pages.userMigration.title") }}
+		</h1>
+		<div>
+			<p
+				class="text-left pa-4"
+				data-testId="text-description"
+				v-html="
```
odalys-dataport, line 11: Is there a way to avoid usage of v-html?

src/pages/user-migration/UserMigration.page.vue
```diff
@@ -0,0 +1,166 @@
...
+					})
+				"
+			></p>
+			<div class="d-flex flex-wrap justify-center mt-8">
+				<v-btn
+					class="mx-8 mb-8"
+					depressed
+					data-testId="btn-cancel"
+					:to="cancelLink"
+				>
```
odalys-dataport, line 24: Please add an aria-label

src/pages/user-migration/UserMigration.page.vue
```diff
@@ -0,0 +1,166 @@
...
+						)
+					}}
+				</v-btn>
+				<v-btn
+					class="mx-8 mb-8"
+					color="primary"
+					depressed
+					data-testId="btn-proceed"
+					:href="proceedLink"
+				>
```
odalys-dataport, line 39: Please add an aria-label

src/pages/user-migration/UserMigration.page.vue
```diff
@@ -0,0 +1,166 @@
...
+			type: String,
+			required: true,
+		},
+		mandatory: {
+			type: Boolean,
+		},
+	},
+	setup(props) {
+		const { createApplicationError } = useApplicationError();
+		if (!props.targetSystem || !props.sourceSystem || !props.origin) {
```
odalys-dataport, line 86: This logic should be in the store in my opinion. I think if you set a business error in the store, this will be dealt with automatically, but I'm not 100% sure about that

odalys-dataport (changes requested):
I wasn't able to test this on the dev instance under this link
https://nbc-n21-500-migration-page.cd.dbildungscloud.dev/migration

Do you have another testing instance?

src/pages/user-migration/UserMigration.page.vue
MarvinOehlerkingCap: I don't know of a way to use line breaks in a translation without v-html. If you know more let me know.

src/pages/user-migration/UserMigration.page.vue
MarvinOehlerkingCap: The button has text that describes what it is doing, so I don't think an aria-label helps here. Or are we adding aria-labels to all buttons now, because I can't see a lot of v-btns in the project that use aria-labels?

src/pages/user-migration/UserMigration.page.vue
odalys-dataport: I see, thanks for explaining

src/pages/user-migration/UserMigration.page.vue
MarvinOehlerkingCap: This is not store logic, but a validation of the query parameters that the page gets called with. The user should not be able to access the page without proper parameters. That's why you didn't see anything on the deployed page, but it is supposed to take you to an error page. I will look into that again.

odalys-dataport: I see, thanks for clarifying

src/layouts/loggedOut.vue
```diff
@@ -1,27 +1,31 @@
...
+			<div class="page">
+				<div class="topbar" data-testid="logged-out-top-bar">
+					<navigation-bar
+						:buttons="true"
+						:img="require('@assets/img/logo/logo-image-mono.svg')"
+						:links="navbarItems"
+					/>
+				</div>
+				<div :class="isMobile ? 'small-wrapper' : 'wrapper'">
+					<Nuxt />
```
OliverHappe, line 13: did you already merge in the nuxt removal branch?

src/pages/user-migration/UserMigration.page.unit.ts
```diff
@@ -0,0 +1,327 @@
...
+import createComponentMocks from "@@/tests/test-utils/componentMocks";
+import { provide } from "@vue/composition-api";
+import { createModuleMocks } from "@utils/mock-store-module";
+import Vue from "vue";
+
+describe("UserMigration", () => {
+	let systemsModule: jest.Mocked<SystemsModule>;
+	let userMigrationModule: jest.Mocked<UserMigrationModule>;
+
+	const setup = (props: object) => {
```
OliverHappe, line 19: i would appreciate a better typing than object.

while it is fine for now, this does not give any information to people adding or refacturing tests in the future.

src/pages/user-migration/UserMigration.page.unit.ts
```diff
@@ -0,0 +1,327 @@
...
+						sourceSystem: "sourceSystemId",
+						origin: "sourceSystemId",
+					});
+
+				expect(func).toThrow("ApplicationError");
+			});
+		});
+
+		describe("when prop origin is missing", () => {
+			it("should throw an error", () => {
```
OliverHappe, line 91: just for my understanding: having this in a callback is required to check if dependency injection failures get handled correctly?

src/layouts/loggedOut.vue
MarvinOehlerkingCap: yes

src/pages/user-migration/UserMigration.page.unit.ts
MarvinOehlerkingCap: you have to give the `expect()` function a callback function in order for jest to handle exceptions in synchronous function calls correctly. Otherwise the test will just throw the error before jest can intercept it.

PR #2382, opened 2023-01-30
N21 241 school admin create tool config

IgorCapCoder:
# Short Description
Schooladmin can add an external tool to be used in the school. The configuration of custom parameters on school scope is also possible including validations.

https://ticketsystem.dbildungscloud.de/browse/N21-241
https://github.com/hpi-schul-cloud/schulcloud-server/pull/3868

## Screenshots of UI changes
image

## Checklist before merging


src/components/administration/external-tool/ExternalToolConfigParameter.vue
```diff
@@ -0,0 +1,105 @@
...
+		const { validateParameter } = useExternalToolValidation(t);
+
+		const getLabelText = (): string => {
+			if (parameter.value.isOptional) {
+				return parameter.value.name;
+			}
+			return `${parameter.value.name} *`;
+		};
+
+		const toolParameterTypeEnumNumber = ToolParameterTypeEnum.Number;
```
OliverHappe, line 93: is it possible to just expose the enum to the template?

`const ToolParameterType = ToolParameterTypeEnum;`

in template: 

`<template v-if="parameter.type === ToolParameterType.Number">`

src/store/external-tool/school-external-tool-status.enum.ts
```diff
@@ -0,0 +1,5 @@
+export enum SchoolExternalToolStatusEnum {
```
OliverHappe, line 1: i don't like having enum in the name. not a strong opinion though....

But having SchoolExternalToolStatus**Enum** suggests to me that there is also SchoolExternalToolStatus\<Something else or nothing>

src/components/administration/external-tool/ExternalToolConfigParameter.vue
arnegns: Yes this works :) But with script setup it would be nicer

src/store/external-tool/school-external-tool-status.enum.ts
arnegns: Good point. I renamed it

src/store/external-tool/tool-configuration-scope.ts
```diff
@@ -0,0 +1,3 @@
+export enum ToolConfigurationScope {
+	SCHOOL = "school",
```
OliverHappe, line 2: SCHOOL is the only enum with caps. i think we should keep it consistent

src/store/external-tool/tool-parameter.enum.ts
```diff
@@ -0,0 +1,8 @@
+export enum ToolParameterType {
+	String = "string",
+	Number = "number",
+	Boolean = "boolean",
+	AutoCourseid = "auto_courseid",
```
OliverHappe, line 5: is this name intentional?

to me, a correct PascalCase-Name would be AutoCourse**Id**

src/store/external-tool/tool-parameter.enum.ts
```diff
@@ -0,0 +1,8 @@
...
+	AutoCoursename = "auto_coursename",
+	AutoSchoolid = "auto_schoolid",
```
OliverHappe, line 7: AutoSchoolId ?

src/store/external-tool/tool-parameter.ts
```diff
@@ -0,0 +1,23 @@
...
+import { ToolParameterType } from "./tool-parameter.enum";
+
+export interface ToolParameter {
+	name: string;
+
+	default?: string;
+
+	regex?: string;
+
+	regexComment?: string;
```
OliverHappe, line 12: i think this is a misleading interface.

as far as i understand it when there is a regex there will always be a regexComment?

if that is true, this interface permits comments without regex and vice versa.

src/store/external-tools.ts
```diff
@@ -45,46 +88,155 @@ export default class ExternalToolsModule extends VuexModule {
...
 		try {
+			this.setLoading(true);
 			if (authModule.getUser?.schoolId) {
 				const resp =
 					await this.toolApi.toolSchoolControllerGetSchoolExternalTools(
 						authModule.getUser.schoolId
 					);
 				const schoolExternalTools: SchoolExternalTool[] =
-					useSchoolExternalToolUtils().mapSchoolExternalToolSearchListResponse(
+					useExternalToolUtils().mapSchoolExternalToolSearchListResponse(
```
OliverHappe, line 106: i have a bad feeling using composables like this. composables are part of vue's entities. while they can be used like simple factories from a technical view i would not expect a composable being used in a store.

If there are functions that are required in a composable and a store they should be outside of a composable scope and imported into the composable instead.

@uidp what do you think about this?

src/store/external-tools.ts
```diff
@@ -45,46 +88,155 @@ export default class ExternalToolsModule extends VuexModule {
...
 						resp.data
 					);
 				this.setSchoolExternalTools(schoolExternalTools);
 			}
 			this.setLoading(false);
-		} catch (e) {
+		} catch (e: any) {
```
OliverHappe, line 112: never use any typing. 

if you don't know (or don't care about) the type us `unknown`

src/store/external-tools.ts
```diff
@@ -45,46 +88,155 @@ export default class ExternalToolsModule extends VuexModule {
...
+	}
+
+	@Action
+	async saveSchoolExternalTool(
+		toolTemplate: ToolConfigurationTemplate
+	): Promise<void> {
+		try {
+			this.setLoading(true);
+			this.resetBusinessError();
+			if (authModule.getUser?.schoolId) {
```
OliverHappe, line 218: for conditions like this i prefer conditional guards.

instead of checking the condition and wrapping all your code within the if, check for the opposite condition and break if it is met. this makes it easier to quickly skim the code and understand hierarchies of conditions.

nesting deeply is difficult to understand.

example: 

```ts

doSomethingComplicated() { 

  if (condition) {
    // all the code that will run if condition is met
   // ...
   // ... 
   // ...
  // ... 
   // ...
   // ... 
  }
}

// compared to: 

doSomethingComplicated() { 

 if (!condition) {
  return 
 }

  // all the code that will run if condition is met
 // ...
 // ... 
 // ...
 // ... 
 // ...
 // ... 
}
```

src/store/external-tools.ts
```diff
@@ -45,46 +88,155 @@ export default class ExternalToolsModule extends VuexModule {
...
+					toolId
+				);
+			const toolConfigurationTemplate: ToolConfigurationTemplate =
+				useExternalToolUtils().mapExternalToolConfigurationTemplateResponse(
+					configTemplate.data
+				);
+			this.setLoading(false);
+
+			return toolConfigurationTemplate;
+		} catch (e: any) {
```
OliverHappe, line 197: no any -> unknown

src/store/external-tools.ts
```diff
@@ -45,46 +88,155 @@ export default class ExternalToolsModule extends VuexModule {
...
+					);
+
+				this.setToolConfigurations(
+					useExternalToolUtils().mapToolConfigurationListResponse(
+						availableTools.data
+					)
+				);
+			}
+			this.setLoading(false);
+		} catch (e: any) {
```
OliverHappe, line 166: no any -> unknown

src/store/external-tools.ts
```diff
@@ -45,46 +88,155 @@ export default class ExternalToolsModule extends VuexModule {
...
+						toolTemplate,
+						authModule.getUser.schoolId
+					);
+
+				await this.toolApi.toolSchoolControllerCreateSchoolExternalTool(
+					schoolExternalToolPostParams
+				);
+			}
+			this.setLoading(false);
+		} catch (e: any) {
```
OliverHappe, line 230: no any -> unknown

tests/test-utils/factory/businessErrorFactory.ts
```diff
@@ -0,0 +1,12 @@
+import { BusinessError } from "@/store/types/commons";
+
+export const businessErrorFactory = (
```
OliverHappe, line 3: please introduce this change to the arc group and encourage people to use it in the existing stores as well. we should create a ticket to establish this as a best practice not only for tests when people agree.

tests/test-utils/factory/schoolExternaToolFactory.ts
```diff
@@ -0,0 +1,16 @@
+import {
+	SchoolExternalTool,
+	SchoolExternalToolStatus,
+} from "@/store/external-tool";
+
+export const schoolExternalToolFactory = (
```
OliverHappe, line 6: i appreciate the idea. i also think we have to talk about this in the arc group because it can possibly get pretty messy pretty quickly if we do not agree on a pattern to organize these factories

src/pages/administration/SchoolSettings.page.vue
```diff
@@ -49,7 +49,7 @@
 						<v-skeleton-loader type="table-thead, table-row, table-row" />
 					</template>
 					<auth-systems v-else :systems="systems"></auth-systems>
-					<external-tools-section></external-tools-section>
+					<external-tools-section id="tools"></external-tools-section>
```
OliverHappe, line 52: might be a really specific ID. can it be something more specific so that we can make sure that ID's in the rendered dom never collide?

ideally create a UUID composable that creates a UUID for this instance that gets used as the id-prefix within a component.

that way we never (almost at least :D ) have colliding id's

src/composables/external-tool-utils.composable.ts
```diff
@@ -0,0 +1,210 @@
...
+	};
+
+	const mapCustomParameterResponse = (
+		parameters: CustomParameterResponse[]
+	): ToolParameter[] => {
+		return parameters.map((resp: CustomParameterResponse) => {
+			return {
+				name: resp.name,
+				// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+				// @ts-ignore
```
OliverHappe, line 118: never use @ts-ignore. there has to be a better solution. let's talk about it.

src/components/administration/external-tool/external-tool-validation.composable.ts
```diff
@@ -0,0 +1,52 @@
+import VueI18n from "vue-i18n";
+import { ToolParameter, ToolParameterType } from "@/store/external-tool";
+
+export function useExternalToolValidation(
```
OliverHappe, line 4: 👏

src/components/administration/external-tool/ExternalToolConfigParameter.vue
```diff
@@ -0,0 +1,133 @@
...
+				text: t("common.words.yes"),
+				value: "true",
+			},
+			{
+				text: t("common.words.no"),
+				value: "false",
+			},
+		]);
+
+		const ToolParameterType = toolParameterType;
```
OliverHappe, line 121: just out of curiosity:

can you return the imported Enum in the setup() immediately?

so not creating the const and just have 

```ts
import { Enum } from '...';
setup() {
 //....
 return {
   // ....
  Enum
 }
}
```

anyway i appreciate the change though having to rename the import makes it a bit clunky unfortunately :(

arnegns: No, unfortunately not.. but with script setup only importing the enum is enough. 
One more thing for us the get vue 3 :)

src/pages/administration/SchoolSettings.page.vue
arnegns: Oh sorry.. this is a remaining id from my anchor scrolling :D I will remove it

src/store/external-tool/tool-parameter.enum.ts
arnegns: Unfortunately it is :( 
We will fix this in the next server ticket and rename it to a correct pascal case name

src/store/external-tool/tool-configuration-scope.ts
arnegns: You are right. Renamed it

src/pages/administration/external-tool/ExternalToolSelectionRow.vue
```diff
@@ -0,0 +1,42 @@
...
+import { ToolConfiguration } from "@/store/external-tool";
+
+export default {
+	name: "ExternalToolSelectionRow",
+	props: {
+		item: {
+			type: Object as PropType<ToolConfiguration>,
+			required: true,
+			default() {
+				return {};
```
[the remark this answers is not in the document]
arnegns: Yes, thanks :)

src/pages/administration/external-tool/ExternalToolSelectionRow.vue
```diff
@@ -0,0 +1,42 @@
...
+			},
+		},
+		maxHeight: {
+			type: String,
+			default() {
+				return "30";
```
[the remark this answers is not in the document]
arnegns: Good point :)

src/store/external-tool/tool-configuration-template.ts
```diff
@@ -0,0 +1,21 @@
+import { ToolParameter } from "./tool-parameter";
+
+export class ToolConfigurationTemplate {
```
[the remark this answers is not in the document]
uidp: I agree. As long as we don't have to use decorators we can use plain objects. In NestJS (in the server) we need decorators and that requires us to use classes. In the frontend that's not necessary.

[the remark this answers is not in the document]
arnegns: I refactor this. Thanks for the explanation :)

src/store/external-tools.ts
uidp: Well in the pinia documentation they says _"Composables are functions that leverage Vue Composition API to encapsulate and reuse stateful logic."_. And they explicitly encourage using composables in stores:
https://pinia.vuejs.org/cookbook/composables.html
Maybe we shouldn't name composables like "<xyz>Utils" and more according to the actual purpose or feature of a composable, e.g. `useExternalToolsMappings` (just as an example).

tests/test-utils/factory/schoolExternaToolFactory.ts
arnegns: I put it on the topic list :)

tests/test-utils/factory/businessErrorFactory.ts
arnegns: I agree with you.. it is already on the topic list

src/store/external-tools.ts
arnegns: Maybe we can look on it together tomorrow. Of course i wanted to use something else than any...

src/composables/external-tool-utils.composable.ts
```diff
@@ -0,0 +1,210 @@
...
+				type: ToolParamTypeMapping[resp.type],
+				scope: ToolParamScopeMapping[resp.scope],
+				// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+				// @ts-ignore
+				value: resp.default ? resp.default : undefined,
+			};
+		});
+	};
+
+	const mapExternalToolConfigurationTemplateResponse = (
```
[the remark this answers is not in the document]
arnegns: Thanks :) will test it tomorrow

tests/test-utils/factory/businessErrorFactory.ts
uidp: We also had a brief attempt on test factories a couple of month ago with tasks:
https://github.com/hpi-schul-cloud/nuxt-client/blob/948b46a865cf6c3d2e3e9e451055e6ca0e2ad354/src/store/task.filter.unit.ts#L17
In the server we are using a library called "Fishery" which is basically a TS implementation of the well-known "Factory Girl" library by Thougthbot:
https://github.com/thoughtbot/fishery

tests/test-utils/factory/schoolExternaToolFactory.ts
uidp: Yes. But maybe we can take a similar approach like in the server repository.

src/composables/external-tool-utils.composable.ts
arnegns: i changed the server api and now the field name is defaultValue instead of default

src/pages/administration/external-tool/ExternalToolConfigOverview.page.vue
```diff
@@ -0,0 +1,183 @@
...
+				to: "/administration/",
+			},
+			schoolSetting,
+			{
+				text: t("pages.tool.title"),
+				disabled: true,
+			},
+		];
+
+		const {getTranslationKey} = useExternalToolMappings();
```
OliverHappe, line 124: i think there are whitespaces missing. pls format the file

src/pages/administration/external-tool/ExternalToolConfigOverview.page.vue
```diff
@@ -0,0 +1,183 @@
...
+				);
+			parametersValid.value = true;
+		};
+
+		const router: VueRouter = useRouter();
+		const onCancel = () => {
+			router.push({path: schoolSetting.to});
+		};
+
+		const saveTool = async () => {
```
OliverHappe, line 160: onSaveTool

src/pages/administration/external-tool/ExternalToolSelectionRow.vue
```diff
@@ -0,0 +1,36 @@
...
+export default {
+	name: "ExternalToolSelectionRow",
+	props: {
+		item: {
+			type: Object as PropType<ToolConfiguration>,
+			required: true,
+			default: () => ({})
+		},
+		maxHeight: {
+			type: [Number, String],
```
OliverHappe, line 27: is it neccessary that this value can be a number or a string?

arnegns: Yes, because the attribute make this to string. I will research this in the next edit ticket :)

OliverHappe (approved):

PR #2384, opened 2023-02-03
BC-2918 - Accessible inline edit (label, aria)

MartinSchuhmacher:
# Short Description
- the title and text element fields do not have visible labels in the context of inline editing like a typical form field 
- it is therefore important, especially for screen readers etc., to pay particular attention to the requirements for low-barrier use
- e.g. an invisible label or aria-label are possible

BC-2918

## Checklist before merging


src/components/card-elements/CardElement.vue
```diff
@@ -19,6 +19,7 @@
 				outlined
 				color="secondary"
 				class="drag-element-btn handle"
+				:aria-label="$t('components.cardElement.dragElement')"
```
virgilchiriac, line 22: do we need also title attr besides area-label?

odalys-dataport: a title attr is not necessary :)

src/components/card-elements/TitleCardElement.vue
```diff
@@ -11,6 +11,7 @@
 			solo
 			flat
 			:placeholder="placeholder"
+			:aria-label="$t('components.cardElement.titleElement')"
```
odalys-dataport, line 14: Screen Reader is reading "Titleelement Title" here, I think it might be reading the placeholder too. We should fix that

odalys-dataport (changes requested):

src/components/card-elements/RichTextCardElement.vue
```diff
@@ -7,6 +7,7 @@
 			v-model="content"
 			:placeholder="placeholder"
 			:disabled="disabled"
+			:aria-label="$t('components.cardElement.richTextElement')"
```
christian-darsow, line 10: the CK5 already comes with an aria-label on itself, so no need to add this here

odalys-dataport:
Implemented in BC-2923

PR #2387, opened 2023-02-06
BC-3168 - The new file store is not available locally after removing nuxt

SevenWaysDP:
# Short Description

https://github.com/hpi-schul-cloud/nuxt-client/pull/2387

## Checklist before merging

### DIFF
--- a/server-proxy.js
+++ b/server-proxy.js
@@ -9,8 +9,9 @@ const { createProxyMiddleware } = require("http-proxy-middleware");
 const {
 	createLegacyClientProxy,
 	createServerProxy,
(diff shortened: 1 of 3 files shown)


uidp (changes requested):
Maybe we should specify the requirements for server routes and fileStorage routes more clearly, something like:
```
const isFileStorage = (path) => {
	return path.startsWith("/api/v3/file");
};
const isServer = (path) => {
	return path.startsWith("/api/v") && !isFileStorage(path);
};
```
This would exclude the `fileStorage` path explicitly from the server routes.

uidp (approved):

PR #2390, opened 2023-02-09
BC-3098 create api definition for LEI

uidp:
# Short Description

## Checklist before merging

### DIFF
--- a/api-definitions.ts
+++ b/api-definitions.ts
@@ -45,7 +45,7 @@
             On each board-mutation, we return the full board-skeleton and update to minimize potential conflicts.			
 */
 
-type BoardCardType = "task" | "content" | "legacy-task" | "legacy-lesson";
+type BoardCardType = "task" | "content" | "legacy-task-reference" | "legacy-lesson-reference";
 
 // GET - Load a Board
 // /boards/:id
--- a/api-definitions.ts
+++ b/api-definitions.ts
@@ -7,9 +7,9 @@
     rename column ✅
     move column ✅
     delete column ✅
-    create card
+    create card ✅
     move card ✅
-    delete card
+    delete card ✅
 
     Actions to support later: 
     Move cards between boards
@@ -96,6 +96,10 @@ interface CourseBoardCard {
 interface VisibilitySettings {
 	publishedAt: string | undefined; // undefined = invisible | Date specifies the visibility from
 }
+
+declare type AnyCourseBoardCard =
+	| CourseBoardLegacyLessonCard
+	| CourseBoardLegacyTaskCard; // union of all CourseBoardCard-Interfaces
 
 interface CourseBoardLegacyLessonCard extends CourseBoardCard {
 	cardType: "legacy-lesson";
@@ -239,28 +243,29 @@ interface Response {
 	version: number;
 }
 
-// POST - Create a Column
-// /boards/:boardId/column
-
 /*
-  Default title of a Column is an empty string. A translated fallback-title is created in the Frontend. 
-  This way we always have a name for a11y and we can be context aware one-dimensional / n-dimensional layout.
+  ---------------------------
+  POST - Create a Column
+  /boards/:boardId/column
 
-  column.cards will be an empty array by default.
+  - title: defaults to an empty string
+    A translated fallback-title is created in the Frontend.
+	This way we always have a name for a11y and we can be context aware one-dimensional / n-dimensional layout.
+  - cards: defaults to empty array
 
-  Validation needs to factor in a DoS-Attack by adding to many columns. Maybe limit to a 100 cols?
+  hint: Validation needs to factor in a DoS-Attack by adding to many columns. Maybe limit to a 100 cols?
 */
 
 interface Payload {}
 
 // Responses
-// 200 OK - No Conflict -> Column was renamed
+// 200 OK - No Conflict -> Column was created
 
 // interface CourseBoard {
 
 // }
 
-// 400 Bad Request - Validation Error -> Column was not renamed
+// 400 Bad Request - Validation Error -> Column was not created
 interface Response {
 	version: number;
 }
@@ -280,7 +285,52 @@ interface Response {
 
 // }
 
-// 400 Bad Request - Validation Error -> Column was not renamed
+// 400 Bad Request - Validation Error -> Column was not removed
+interface Response {
+	version: number;
+}
+
+/*
+  ---------------------------
+  POST - Create a Card
+  /boards/:boardId/column/:columnId/createcard
+  ---------------------------
+
+  - title: defaults to an empty string
+  - elements: defaults to empty array
+*/
+interface Payload {
+	type: CourseBoardCardType;
+}
+
+// Responses
+// 200 OK - No Conflict -> Card was created
+// interface AnyCourseBoardCard {
+
+// }
+
+// 400 Bad Request - Validation Error -> Card was not created
+interface Response {
+	version: number;
+}
+
+/*
+  ---------------------------
+  DELETE - Delete a Card
+  /boards/:boardId/card/:cardId
+  ---------------------------
+
+  - Deleting a Card also updates the board.
+ */
+
+// Responses
+// 200 OK - No Conflict -> Card was removed
+
+// interface CourseBoard {
+
+// }
+
+// 400 Bad Request - Validation Error -> Column was not removed
 interface Response {
 	version: number;
 }
--- a/api-definitions.ts
+++ b/api-definitions.ts
@@ -57,7 +57,6 @@ interface Board {
 	id: string;
 	title: string;
 	columns: BoardColumn[];
-	version: number;
 	timestamps: BoardTimestamps;
 }
 
--- a/api-definitions.ts
+++ b/api-definitions.ts
@@ -1,0 +1,175 @@
+
+/**  
+    Actions to support in MVP:
+    read board ✅
+    read cards ✅
+    rename board
+    create column
+    rename column
+    move column ✅
+    delete column
+    create card
+    move card ✅
+    delete card
+
+    Actions to support later: 
+    Move cards between boards
+    create board
+    delete board
+
+    Other thoughts:
+        - we have to support cards being added/removed from other places (task overview)
+        - we have to support cards being added/removed from legacy (tasks and lessons)
+        - we might want to load board information from rough to detailed (we had the idea to first render loading states for all cards, and then query each card individually)
+        - create and delete board is not necessarily needed in the first version
+        - single requests for moving single elements, or reorder?
+
+*/
+
+
+interface CourseBoardTaskCard {
+
+}
+
+interface CouseBoardContentCard {
+
+}
+
+interface CourseBoardElement {
+
+}
+
+type CourseBoardCardType = 'task' | 'content' | 'legacy-task' | 'legacy-lesson';
+
+// GET - Load a CourseBoard
+// /boards/:id
+
+interface CourseBoard {
+    id: string;
+    title: string;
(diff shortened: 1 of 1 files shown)


api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+interface CourseBoardCard {
+	id: string;
+	height: number;
+	elements: ContentElement[];
+	cardType: CourseBoardCardType;
+	visibility: VisibilitySettings;
+}
+
+interface VisibilitySettings {
+	publishedAt: string | undefined; // undefined = invisible | Date specifies the visibility from
```
uidp, line 97: Can the `publishedAt` property be also relevant for other aspects then the visibility of a card? In that case we shouldn't tie it too close to only the visibility.

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+
+  Option 2: We track a Board-Version for each board mutation. When there is a Version mismatch greater than +1 
+            (e.g. last known version is 10, backend returns a 12) the Frontend reloads the board.
+
+
+  **Intermediary decision**
+  Option 3: We ignore multi-tenant edits until we have websockts. 
+            On each board-mutation, we return the full board-skeleton and update to minimize potential conflicts.			
+*/
+
```
uidp, line 44: What are these 3 interfaces used for? There are other similiar types defined below.

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+interface CourseBoardTimestamps {
+	lastUpdatedAt: string;
+	createdAt: string;
+	deletedAt: string;
+}
+
+interface CourseBoardColumn {
+	id: string;
+	title: string;
+	cards: CourseBoard_Skeleton_Card[]; // ?
```
uidp, line 73: The `CourseBoard_Skeleton_Card` type has only `id` and `height` properties. Is it intended that we cannot get the list of "full" cards from a column (in terms of lazy loading)?

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+}
+
+declare type AnyCourseBoardCard =
+	| CourseBoardLegacyLessonCard
+	| CourseBoardLegacyTaskCard; // union of all CourseBoardCard-Interfaces
+
+interface CourseBoardLegacyLessonCard extends CourseBoardCard {
+	cardType: "legacy-lesson";
+	lessonId: string;
+	elements: ContentElement[]; // narrowed down to allowed Elements: LegacyLesson
```
uidp, line 107: Is the relation between card and legacy lesson a 1:1 or 1:n relationship?

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+
+	// some properties
+}
+
+interface CourseBoardLegacyTaskCard extends CourseBoardCard {
+	cardType: "legacy-task";
+	taskId: string;
+	elements: ContentElement[]; // narrowed down to allowed Elements: LegacyTask
```
uidp, line 115: Is the relation between card and legacy task a 1:1 or 1:n relationship?

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+	// some properties
+}
+
+declare type AnyContentElement = TextContentElement | TitleContentElement; // union of all Element-Interfaces
```
uidp, line 119: Is that union supposed to include all other element interfaces as well? (like `image`, `task`, `legacy-lesson`...)

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+
+interface TextContentElement {}
+
+interface TitleContentElement {}
+
+// **IDEA**
+// We want to prevent collisions or at least notify about collisions by versioning the board with an autoincrement ID on database manipulation
+// how we transfer the version on each request is up for debate.
+
+// PUT - Move a Card on the board
```
uidp, line 142: Maybe it's more like `PATCH` because we are doing a partial update of the board?

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+	deletedAt: string;
+}
+
+interface CourseBoardColumn {
+	id: string;
+	title: string;
+	cards: CourseBoard_Skeleton_Card[]; // ?
+}
+
+interface CourseBoard_Skeleton_Card {
```
uidp, line 76: Why the underscores?

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+// 400 Bad Request - some id did not exist -> Card was not moved, reload
+interface Response {
+	version: number;
+}
+// 409 Conflict - Action conflicts -> reload board, card was not moved
+interface Response {
+	version: number;
+}
+
+// PUT - Move a Column on the board
```
uidp, line 166: partial update => `PATCH`?

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+// 400 Bad Request - some id did not exist -> Column was not moved, reload
+interface Response {
+	version: number;
+}
+// 409 Conflict - Action conflicts -> reload board, Column was not moved
+interface Response {
+	version: number;
+}
+
+// PUT - Rename a Board
```
uidp, line 212: partial update => PATCH?

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+// 200 OK - No Conflict -> Board was renamed
+interface Response {
+	version: number;
+}
+// 400 Bad Request - Validation Error -> board was not renamed
+interface Response {
+	version: number;
+}
+
+// PUT - Rename a Column
```
uidp, line 229: partial update => PATCH?

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+interface CouseBoardContentCard {}
+
+interface CourseBoardElement {}
+
+type CourseBoardCardType = "task" | "content" | "legacy-task" | "legacy-lesson";
+
+// GET - Load a CourseBoard
+// /boards/:id
+
+interface CourseBoard {
```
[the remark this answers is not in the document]
OliverHappe: possibly. 

generally i try to avoid single-word interface names but this might be a case where `Board` is the eprfect description

api-definitions.ts
OliverHappe: I think this Date is introduced specifically to set a Date where the card will become visible. 

We want to make the VisibilitySettings interface reusable, having a publishedAt date for that is a useful property. 

I don't think we should extract the property from this interface.

api-definitions.ts
OliverHappe: yes. it is a skeleton card because it will be used to render card-skeletons specifically. 

if a use-case comes up were we need to load alls cards of a board in one request we should implement a specific endpoint at that time. 

lazy loading the cards is a decent start for the board

api-definitions.ts
OliverHappe: yes.

AnyContentElement will collect all known ContentElement-Types so we can describe any mutation of elements on cards.

api-definitions.ts
OliverHappe:
```suggestion
// PATCH - Move a Card on the board
```

api-definitions.ts
OliverHappe: the name **Skeleton**Card sparked some discussion that we skipped in favor of making progress.

the underscores are there to indicate that. 

I think Skeleton is a well name describing the purpose of the interface.

```suggestion
interface CourseBoardSkeletonCard {
```

api-definitions.ts
OliverHappe:
```suggestion
// PATCH - Rename a Board
```

api-definitions.ts
OliverHappe:
```suggestion
// PATCH - Rename a Column
```

api-definitions.ts
OliverHappe:
```suggestion
	cards: CourseBoardSkeletonCard[]; // ?
```

api-definitions.ts
OliverHappe: we did not specify their content since we started with the minimum required interfaces. They might be unneccessary though is suspect we will need them soon :)

api-definitions.ts
```diff
@@ -0,0 +1,336 @@
...
+    rename column ✅
+    move column ✅
+    delete column ✅
+    create card ✅
+    move card ✅
+    delete card ✅
+
+    Actions to support later: 
+    Move cards between boards
+    create board
```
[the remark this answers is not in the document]
OliverHappe: initially every course will have one board by default.

createBoard will be more relevant (at least for the API) when we want to create subboards

api-definitions.ts
OliverHappe: the card itself won't have the relationship but the element on the card. each element will have a 1:1 relationship to a legacy entity.

api-definitions.ts
OliverHappe: the card itself won't have the relationship but the element on the card. each element will have a 1:1 relationship to a legacy entity.

api-definitions.ts
OliverHappe:
```suggestion
// PATCH - Move a Column on the board
```

api-definitions.ts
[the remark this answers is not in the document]
hoeppner-dataport: I think we said, that "Board" is already in use - and we do not want to replace it immediatly with the first commits.

api-definitions.ts
```diff
@@ -0,0 +1,274 @@
...
+  Option 2: We track a Board-Version for each board mutation. When there is a Version mismatch greater than +1 
+            (e.g. last known version is 10, backend returns a 12) the Frontend reloads the board.
+
+
+  **Intermediary decision**
+  Option 3: We ignore multi-tenant edits until we have websockts. 
+            On each board-mutation, we return the full board-skeleton and update to minimize potential conflicts.			
+*/
+
+type BoardCardType = "task" | "content" | "legacy-task" | "legacy-lesson";
```
OliverHappe, line 48:
```suggestion
type BoardCardType = "task" | "content" | "legacy-task-reference" | "legacy-lesson-reference";
```

api-definitions.ts
```diff
@@ -0,0 +1,266 @@
...
+	| "legacy-task-reference"
+	| "legacy-lesson-reference";
+
+// GET - Load a Board
+// /boards/:id
+interface Board {
+	id: string;
+	title: string;
+	columns: BoardColumn[];
+	version: number;
```
OliverHappe, line 60:
```suggestion
```

api-definitions.ts
```diff
@@ -0,0 +1,265 @@
...
+}
+
+interface VisibilitySettings {
+	publishedAt: string | undefined; // undefined = invisible | Date specifies the visibility from
+}
+
+declare type AnyBoardCard = BoardLegacyLessonCard | BoardLegacyTaskCard; // union of all BoardCard-Interfaces
+
+interface BoardLegacyLessonCard extends BoardCard {
+	cardType: "legacy-lesson";
```
NFriedo, line 102:
```suggestion
	cardType: "legacy-lesson-reference";
```
Should it be legacy-lesson-reference?

api-definitions.ts
```diff
@@ -0,0 +1,265 @@
...
+	lessonId: string;
+	elements: ContentElement[]; // narrowed down to allowed Elements: LegacyLesson
+
+	// some properties
+}
+
+interface BoardLegacyTaskCard extends BoardCard {
+	cardType: "legacy-task";
```
NFriedo, line 110:
```suggestion
	cardType: "legacy-task-reference";
```

Should it be legacy-task-reference?

PR #2415, opened 2023-02-24
BC-3243 - Implement view of existing taskcard for students

MartinSchuhmacher:
# Short Description

Create a page for students of an existing taskcard based on infos from backend. (course, title, text etc.) No possibillity to edit (no input fields, no editor - just view) 

BC-3243

https://github.com/hpi-schul-cloud/schulcloud-server/pull/3927
## Changes

- task card page is used for new task cards, for teacher edit view and for student view (according to permission)
- task and course in task will now be correctly populated in task card, so the access can be checked correctly
- edit view has the same features (draggable, editable) as before (just with cleaner code)
- simple view has the same fields just not editable, also add element buttons are removed as well as draggable is not possible now (buttons will be done in a future ticket)
## Screenshots of UI changes
Teacher view:
image

Student view:
image

## Checklist before merging


src/pages/TaskCard.page.vue
```diff
@@ -4,9 +4,13 @@
...
 		headline="Task Card"
 	>
-		<v-form class="d-flex flex-column">
-			<card-element-wrapper v-model="title.model" v-bind="title.props" />
-			<card-element-list v-model="elements" />
+		<v-form v-if="isEditMode" class="d-flex flex-column">
+			<card-element-wrapper
+				v-model="title.model"
+				v-bind="title.props"
+				:editMode="isEditMode"
```
virgilchiriac, line 11: since the if condition above, why not put true for the value? it makes code more readable

src/pages/TaskCard.page.vue
```diff
@@ -4,9 +4,13 @@
 		:breadcrumbs="breadcrumbs"
 		headline="Task Card"
 	>
-		<v-form class="d-flex flex-column">
-			<card-element-wrapper v-model="title.model" v-bind="title.props" />
-			<card-element-list v-model="elements" />
+		<v-form v-if="isEditMode" class="d-flex flex-column">
```
virgilchiriac, line 7: is correct we don't have .value for isEditMode, right?

MartinSchuhmacher: yes that is correct, for booleans seems we dont need the separate .value call

christian-darsow: Just as information: this .value syntax for the reactive vars is not required in templates -> https://vuejs.org/guide/essentials/reactivity-fundamentals.html#ref-unwrapping-in-templates

src/router/routes.ts
```diff
@@ -166,10 +166,10 @@ export const routes: Array<RouteConfig> = [
 		name: "tasks",
 	},
 	{
-		path: `/task-cards/:id(${REGEX_ID})/edit`,
+		path: `/task-cards/:id(${REGEX_ID})`,
 		component: () => import("../pages/TaskCard.page.vue"),
-		name: "task-card-edit",
-		beforeEnter: createPermissionGuard(["task_card_edit"]),
+		name: "task-card-view",
```
virgilchiriac, line 171: route name could be task-card or task-card-view-edit

src/router/routes.ts
```diff
@@ -166,10 +166,10 @@ export const routes: Array<RouteConfig> = [
...
+		beforeEnter: createPermissionGuard(["task_card_view"]),
```
virgilchiriac, line 172: side note: create permissions guard could implement an OR condition. In that case, page could be accesible with either task_card_view or task_card_edit permission

src/pages/TaskCard.page.vue
```diff
@@ -176,12 +201,20 @@ export default defineComponent({
 			router.go(-1);
 		};
 
+		const getUserPermissions = ref(authModule.getUserPermissions);
+
+		const isEditMode = computed(() => {
+			if (getUserPermissions.value.includes("task_card_edit")) return true;
```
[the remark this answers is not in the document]
MartinSchuhmacher: created a ticket for that: https://ticketsystem.dbildungscloud.de/browse/BC-3346

src/router/routes.ts
MartinSchuhmacher: done

src/router/routes.ts
MartinSchuhmacher: good idea, as long as we go with task_card_view for both user roles I guess its enough like this right now.
if we decide to only put one of both per user role that we should implement as you say

src/components/card-elements/CardElementWrapper.vue
```diff
@@ -83,6 +86,10 @@ export default defineComponent({
 		editable: {
 			type: Boolean,
 		},
+		editMode: {
```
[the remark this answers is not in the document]
MartinSchuhmacher: removed editable from this page, still left editMode (page level as well as for button logic) and editable (on element level if this should be editable or not) separate for now

src/pages/TaskCard.page.vue
```diff
@@ -4,23 +4,46 @@
...
+					color="primary"
+					depressed
+					@click="save"
+					data-testid="save-btn"
+				>
 					{{ $t("common.actions.save") }}
 				</v-btn>
 			</div>
 		</v-form>
+		<v-form v-else class="d-flex flex-column">
```
christian-darsow, line 34: Maybe we use something different then v-form that fits more from markup aspect?

MartinSchuhmacher: Dont know what that means ^^ can you make a suggestion?

src/components/card-elements/CardElementWrapper.unit.ts
```diff
@@ -18,58 +19,150 @@ const getWrapper = (props?: object, options?: object) => {
...
+			const deleteBtn = wrapper.find('[data-testid="delete-element-btn"]');
 
-	it("should render dynamic component", async () => {
-		const wrapper = getWrapper({
-			component: CardElementComponentEnum.RichText,
+			expect(deleteBtn.exists()).toBe(false);
 		});
 
-		expect(wrapper.findComponent(RichTextCardElement).exists()).toBe(true);
+		it("add element button for title should not be present", () => {
```
christian-darsow, line 110: This should actually not even be present (visible) in editMode. Maybe we add it in the above section at well?

src/pages/TaskCard.page.vue
christian-darsow: I thought about something like <article>

MartinSchuhmacher: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article
makes sense to me, will try to implement

src/pages/TaskCard.page.unit.ts
```diff
@@ -21,16 +21,23 @@ jest.mock("@/store", () => ({
...
+const getWrapper = (
+	userPermission: string,
+	props?: object,
+	options?: object
+) => {
 	const componentOptions = createComponentMocks({ i18n: true });
 	const { localVue } = componentOptions;
 	localVue.use(VueRouter);
-	const router = new VueRouter({ routes: [{ path: "/tasks/new" }] });
+	const router = new VueRouter({ routes: [{ path: "task-card-view-edit" }] });
```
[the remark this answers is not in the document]
MartinSchuhmacher: you are right, wanted to change that but forgot. will adapt your suggestion

src/components/card-elements/CardElementWrapper.unit.ts
MartinSchuhmacher: 👍

src/pages/TaskCard.page.vue
```diff
@@ -4,23 +4,46 @@
...
+					color="primary"
+					depressed
+					@click="save"
+					data-testid="save-btn"
+				>
 					{{ $t("common.actions.save") }}
 				</v-btn>
 			</div>
 		</v-form>
+		<article v-else class="d-flex flex-column">
```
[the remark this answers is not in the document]
MartinSchuhmacher: see Christians comment above regarding Markup

src/pages/TaskCard.page.vue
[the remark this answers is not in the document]
MartinSchuhmacher: 👍

src/components/card-elements/CardElementWrapper.vue
```diff
@@ -28,7 +28,7 @@
 		<component
 			:is="component"
 			v-model="model"
-			:editable="editable"
+			:editable="true"
```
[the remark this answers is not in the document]
MartinSchuhmacher: no, as we need to also set/unset all the buttons, and its much cleaner codewise that way (I think) instead of putting all the conditions on every component

odalys-dataport (approved):

PR #2424, opened 2023-03-05
BC-3107 null check fix

sszafGCA:
# Short Description

- https://ticketsystem.dbildungscloud.de/browse/BC-3107

## Checklist before merging

### DIFF
--- a/src/pages/administration/TeacherOverview.page.vue
+++ b/src/pages/administration/TeacherOverview.page.vue
@@ -313,7 +313,7 @@ export default {
 			);
 
 			// filters out the QR bulk action is user is not an admin
-			if (!this.user.roles.some((role) => role.name === "administrator")) {
+			if (!this.user?.roles?.some((role) => role.name === "administrator")) {
 				editedActions = editedActions.filter(
 					(action) =>
 						action.label !==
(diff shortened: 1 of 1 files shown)


MateuszSapalaPassCon (approved):

odalys-dataport (approved):

src/pages/administration/TeacherOverview.page.vue
```diff
@@ -313,7 +313,7 @@ export default {
 			);
 
 			// filters out the QR bulk action is user is not an admin
-			if (!this.user.roles.some((role) => role.name === "administrator")) {
+			if (!this.user?.roles?.some((role) => role.name === "administrator")) {
```
MartinSchuhmacher, line 316: we should always have a user right? why making it optional?

sszafGCA: Because it is null when we logging out and that's why error is thrown

OliverHappe: i think instead of adding many question marks, it might be better to add an additional condition:

`if(this.user && !this.user.roles....)` 

what do you think @sszafGCA ?

PR #2425, opened 2023-03-06
BC-3222 - show task card on room board

odalys-dataport:
# Short Description

https://ticketsystem.dbildungscloud.de/browse/BC-3222
https://ticketsystem.dbildungscloud.de/browse/BC-3221

## Checklist before merging


src/components/molecules/RoomTaskCard.vue
```diff
@@ -367,7 +384,11 @@ export default {
 		},
 		handleClick() {
 			if (!this.dragInProgress) {
-				window.location = `/homework/${this.task.id}`;
+				if (this.isBetaTask) {
```
MartinSchuhmacher, line 387: I was thinking if it makes sense to put this redirect into a separate private function like `redirectTask` or something, as I am not a huge fan of two if statements

src/locales/en.json
```diff
@@ -861,6 +861,7 @@
 	"pages.room.taskCard.label.open": "Open",
 	"pages.room.taskCard.label.done": "Finish",
 	"pages.room.taskCard.label.edit": "Edit",
+	"pages.room.taskCard.label.betaTask": "Beta-Task",
```
MartinSchuhmacher, line 864: english should not use "-", just use `Beta task`

src/locales/es.json
```diff
@@ -854,6 +854,7 @@
 	"pages.room.taskCard.label.open": "Abrir",
 	"pages.room.taskCard.label.done": "Terminar",
 	"pages.room.taskCard.label.edit": "Editar",
+	"pages.room.taskCard.label.betaTask": "Tarea Beta",
```
MartinSchuhmacher, line 857: `Tarea beta` would be my suggestion

src/locales/ua.json
```diff
@@ -842,6 +842,7 @@
 	"pages.room.taskCard.label.taskDone": "Завдання виконано",
 	"pages.room.taskCard.student.label.overdue": "Відсутня",
 	"pages.room.taskCard.teacher.label.overdue": "Термін дії минув",
+	"pages.room.taskCard.label.betaTask": "Бета-завдання",
```
MartinSchuhmacher, line 845: same as above, ukrainian also does not use hiphens here (at least not in russian and its the similar grammar), so simply `Бета завдання` is enough

src/pages/RoomDetails.page.vue
```diff
@@ -318,12 +316,20 @@ export default defineComponent({
...
-	async created() {
-		await roomModule.fetchContent(this.courseId);
+	// async created() {
+	// 	await roomModule.fetchContent(this.courseId);
+	// 	await roomModule.fetchScopePermission({
+	// 		courseId: this.courseId,
+	// 		userId: authModule.getUser.id,
+	// 	});
+	// },
+	async beforeRouteEnter(to, from, next) {
```
MartinSchuhmacher, line 326: in case this new behaviour substitutes the old one, I would remove the commented function

src/components/molecules/RoomTaskCard.vue
```diff
@@ -178,6 +184,15 @@ export default {
 				new Date(scheduledDate).getTime() - delay > new Date().getTime()
 			);
 		},
+		isBetaTask() {
+			return !!this.task.taskCard;
```
MartinSchuhmacher, line 188: not a big fan of conversions, maybe a straight `!== null` would give the straight answer to the check here

src/components/molecules/RoomTaskCard.vue
```diff
@@ -178,6 +184,15 @@ export default {
...
+		},
+		titleIcon() {
+			return this.isBetaTask ? "$taskDoneFilled" : "$tasks";
+		},
+		titleIconColor() {
+			return this.isBetaTask ? "task" : "";
```
MartinSchuhmacher, line 194: are there plans for other colors as well? because below you name a class "beta-task" with separate background color. does this "task" here mean a different task then or are they supposed to be used for the same "type" of task? then a matching naming might come in handy

src/pages/RoomDetails.page.vue
odalys-dataport: While it does work on local, apparantly it does not work on dev instance

src/router/routes.ts
```diff
@@ -9,6 +9,7 @@ import {
 	REGEX_UUID,
 } from "@/utils/validationUtil";
 
+// routes configuration sorted in alphabetical order
```
MartinSchuhmacher, line 12: I dont think that comment is necessary for future work

src/router/routes.ts
```diff
@@ -121,6 +122,7 @@ export const routes: Array<RouteConfig> = [
 			isPublic: true,
 		},
 	},
+	// deprecated?
```
MartinSchuhmacher, line 125: can we handle the deprecated routes/create a follow up ticket instead of writing a comment here which will not be tackled in the future?

OliverHappe (changes requested):

src/components/molecules/RoomTaskCard.unit.ts
```diff
@@ -223,6 +245,26 @@ describe("@/components/molecules/RoomTaskCard", () => {
...
+			const wrapper = getWrapper({ ...betaTaskTestProps, role });
+			const title = wrapper.find(".title-section");
+
+			expect(title.element.textContent).toContain("Beta-Aufgabe – Entwurf");
+		});
+
+		it("should redirect to homework page", () => {
+			const { location } = window;
+			const wrapper = getWrapper({ ...betaTaskTestProps, role });
+			const taskCard = wrapper.find(".task-card");
```
OliverHappe, line 261: a card should not be a button without role=button

OliverHappe, line 261: imo it's cleaner to not use classes as a selector for wrappers.

in this case .findComponent(VCard) should be used. especially since the class you are trying to find here is added dynamically. it should be obvious which element you are selecting when looking at the template.

src/components/molecules/RoomTaskCard.vue
```diff
@@ -15,78 +15,96 @@
...
 	>
-		<v-card-text data-testid="content-card-task-content">
-			<div class="top-row-container mb-0">
-				<div class="title-section" tabindex="0">
-					<v-icon size="14">{{ icons.mdiFormatListChecks }}</v-icon>
-					{{ cardTitle(task.duedate) }}
+		<template v-if="isBetaTask">
+			<v-card-text data-testid="content-card-task-content">
+				<div class="top-row-container mb-0">
+					<div class="title-section" tabindex="0">
```
OliverHappe, line 21: titles should use an <hX> element to be accessible for screenreaders

OliverHappe: why is there a tabindex on the title?

src/components/molecules/RoomTaskCard.vue
```diff
@@ -15,78 +15,96 @@
...
+				</div>
+				<div class="text-h6 text--primary mb-2 task-name">
+					{{ task.name }}
+				</div>
+				<!-- eslint-disable vue/no-v-html -->
+				<div
+					v-if="canShowDescription"
+					class="text--primary mt-1 mb-0 pb-0 text-description"
+					tabindex="0"
+					v-html="task.description"
```
OliverHappe, line 58: why no interpolation for the description string? v-html poses a potential xss vulnerability!

OliverHappe (changes requested):

src/components/molecules/RoomTaskCard.unit.ts
odalys-dataport: good point, I added the role to v-card

odalys-dataport: You're right, I just copied from the existing tests and didn't care too much. Adjusted it for my new test, would not like to adjust it for all tests though.

src/components/molecules/RoomTaskCard.vue
odalys-dataport: good question, I removed it for both old and new

I made the name of the task an h6 now. The title section is the subtitle above the task name. I didn't name it like that, it's from the existing code

src/components/molecules/RoomTaskCard.vue
odalys-dataport: this is also from the existing code and not from my changes

OliverHappe (approved):

PR #2453, opened 2023-03-20
BC-3507 - Fix beta task reload

christian-darsow:
# Short Description

After creating a beta task the task is not immediately shown on the room details page. The page needs to be reloaded for the task to be visible.

After changing the title of an existing beta-task the updated title is not immediately shown on the room details page. The page needs to be reloaded for the task to be updated.

https://ticketsystem.dbildungscloud.de/browse/BC-3507

## Checklist before merging


MartinSchuhmacher (approved):

src/components/molecules/RoomTaskCard.vue
```diff
@@ -408,10 +408,15 @@ export default {
 		},
 		handleClick() {
 			if (!this.dragInProgress) {
-				const route = this.isBetaTask
-					? `/task-cards/${this.task.taskCardId}`
-					: `/homework/${this.task.id}`;
+				if (this.isBetaTask) {
```
MartinSchuhmacher, line 411: to be honest I would like that part of "seting the route" being in a separate private function

src/pages/TaskCard.page.vue
```diff
@@ -156,11 +157,27 @@ export default defineComponent({
...
+						content: {
+							value: "",
+							inputFormat:
+								CardRichTextElementResponseInputFormatEnum.RichtextCk5,
+						},
+					},
+				];
 
-				taskCardModule.setCourseId(course.value);
 				dueDate.value = endOfSchoolYear.toISOString();
```
MartinSchuhmacher, line 179: the seting of the courseId above here is not necessary because it is done in the `createTaskCard` function, correct?

MartinSchuhmacher (changes requested):

src/components/molecules/RoomTaskCard.vue
christian-darsow: yeap, good idea! created a new computed property for that

MartinSchuhmacher (dismissed):

src/pages/TaskCard.page.vue
```diff
@@ -156,11 +157,27 @@ export default defineComponent({
...
+					{
+						id: "",
+						cardElementType: CardElementResponseCardElementTypeEnum.RichText,
+						content: {
+							value: "",
+							inputFormat:
+								CardRichTextElementResponseInputFormatEnum.RichtextCk5,
+						},
+					},
+				];
```
odalys-dataport, line 177: Shouldn't we adjust the initial data in store instead of here?

christian-darsow: I think (now) it's easier to read like this having all the initial values next to each other, initial cardElements, duedate etc and not getting one of those from the store. and just in case we want ot have other initial values somewhere, e.g two text contents, also more flexible.  what do you think?

odalys-dataport: mmh let's leave it like this for now and think about if it makes more sense in the store

src/pages/TaskCard.page.vue
```diff
@@ -149,11 +150,27 @@ export default defineComponent({
 						title: roomData.title,
 					},
 				];
-				const taskCardData = taskCardModule.getTaskCardData;
+				const initialCardElements = [
```
odalys-dataport, line 153: with the title refactor pr, this needs to be refactored as well, right?

odalys-dataport (changes requested):

src/pages/TaskCard.page.vue
christian-darsow: yeap, right. I thought that merging main into it would solve this or at least give me a hint of a conflict. but not

odalys-dataport (approved):

PR #2456, opened 2023-03-21
BC-3403 content card

OliverHappe:
# Short Description

Adds a title and context menu to CardHost components on the board

- https://ticketsystem.dbildungscloud.de/browse/BC-3403

## Checklist before merging


src/components/feature-board/CardHeaderTitleInput.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+	<VTextarea
```
OliverHappe, line 2: This will also get an accessible h3 tag later, we need to wait for the results of another ux-meeting to know how it should behave. then we will add better a11y

src/components/feature-board/CardHostMenu.vue
```diff
@@ -0,0 +1,36 @@
...
+				v-bind="attrs"
+				v-on="on"
+				:ripple="false"
+			>
+				<v-icon>{{ mdiDotsVertical }}</v-icon>
+				<span class="d-sr-only">Card-Menu</span>
+			</v-btn>
+		</template>
+		<v-list>
+			<slot></slot>
```
christian-darsow, line 19: While I like this approach by just building the menu with a slot, I also wonder if it wouldn't be more clear on the long run by passing the items as a prop.

src/components/feature-board/CardState.composable.ts
```diff
@@ -1,28 +1,73 @@
-import { onMounted, ref } from "vue";
+import { onMounted, reactive, ref, toRef, unref } from "vue";
```
christian-darsow, line 1: Just wanted to understand the use-case of unref when I saw its not used :D

src/components/feature-board/CardState.composable.ts
```diff
@@ -1,28 +1,73 @@
...
 	onMounted(() => fetchCard(id));
 
 	return {
 		fetchCard,
-		card,
-		isLoading,
+		updateTitle,
+		deleteCard,
+		updateCardHeight,
+		card: toRef(cardState, "card"),
```
christian-darsow, line 70: the props of the reactive cardState would loose reactivity when returning like this `card: cardState.card` ?

src/components/feature-board/CardHostMenu.vue
OliverHappe: https://maecapozzi.com/blog/composition-vs-configuration

i personally try to follow a composition approach because it keeps most of the complexity to the template. I prefer seeing a more verbose template over a component with many objects as props. But as the blogpost suggests, there are pros and cons for both approaches :)

src/components/feature-board/CardState.composable.ts
OliverHappe: i will remove the unused import

PR #2463, opened 2023-03-24
Devcontainer

janrenz:
# Short Description

This adds support for devcontainers (see also https://github.com/hpi-schul-cloud/schulcloud-server/pull/4020).

Running devcontainer for the SPA frontend will also require running  server and legacy client in docker (see how we set the URLs to the docker host).


Loki-Afro:
hi @janrenz docker-compose is dead. devcontainers just inherit the issues of docker-compose.
it may be transparent to debug but ultimately (docker/podman .. ) it very platform/environment depended. 

therefore i have to reject your pr.

PR #2464, opened 2023-03-27
Bc 3402 create element list

OliverHappe:
# Short Description

This PR adds the Element-List on ContentCards.

Main pieces are: 
 * CardHostInteractionHandler.vue
  > Renderless Wrapper Component for catching and dispatching mouse / keyboard events to components on the card
 * useCardHostInteractionHandler.composable.ts
  > Composable to listen for events done on the card and get a callback if current component is targeted
 * useContentElementState.composable.ts
  > Component State for each Element. Generically typed and includes targeted focus for elements
 * TextContentElement.vue
  > Draft implementation for a basic text element. 

- https://ticketsystem.dbildungscloud.de/browse/BC-3402
- https://ticketsystem.dbildungscloud.de/browse/BC-3403 (included - approved beforehand)

## Checklist before merging


src/components/feature-board/content-elements/ContentElementList.vue
```diff
@@ -0,0 +1,37 @@
+<template>
+	<VCardText>
+		<template v-for="(element, index) in elements">
+			<TextContentElement
+				v-if="element.type === 'text'"
```
christian-darsow, line 5: see below about the idea of changing the type of ContentElementType

src/components/feature-board/content-elements/TextContentElement.vue
christian-darsow, line 1: I was also searching for something like that - but unsuccesfull: do you know any way to tell that this `extends AnyContentElement `

src/components/feature-board/types/Card.ts
```diff
@@ -1,28 +1,10 @@
...
 	visibility: VisibilitySettings;
 	title: string;
 }
-
-export interface LegacyLessonCard extends BaseCard {
-	// cardType: BoardCardType.LegacyLesson;
-	lessonId: string;
-	elements: ContentElement[]; // narrowed down to allowed Elements: LegacyLesson
-}
-
```
christian-darsow, line 23: Its all "just" cards now, right? for now don't care about this legacy stuff? Just ask to also understand as we are still working a lot with the old tasks/homework

src/components/feature-board/card/CardHostInteractionHandler.vue
```diff
@@ -0,0 +1,63 @@
...
+	props: {
+		isEditMode: {
+			type: Boolean,
+			required: true,
+		},
+	},
+	emits: ["start-edit-mode", "end-edit-mode", "move-card-keyboard"],
+	setup(props, { emit }) {
+		const interactionEvent = shallowRef<{ x: number; y: number } | undefined>();
+		provide("CARD_HOST_INTERACTION_EVENT", interactionEvent);
```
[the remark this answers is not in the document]
OliverHappe: yes. will do! thanks for the reminder :)

src/components/feature-board/state/CardState.composable.ts
```diff
@@ -0,0 +1,97 @@
...
+			return;
+		}
+		cardState.card.height = newHeight;
+	};
+
+	const addElement = (type: ContentElementType) => {
+		if (cardState.card === undefined) {
+			return;
+		}
+		if (type === "text") {
```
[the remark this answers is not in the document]
OliverHappe: in generally prefer unions because they allow easier usage in templates. their advantage is obviously that they appear to be "less strict" or typesafe in typescript context. 

to use enums in a template you have it import the enum and return it in the setup function: 

```ts

//...
setup() {

 return {
    ContentElementType
 }
}
```

```html
 <template v-if="type === ContentElementType.Text">
  // compared to
 <template v-if="type === 'text'">
```

unfortunately we do not have consensus if we want to use enums consistently? I will raise this question in the arc meeting tomorrow :)

src/components/feature-board/content-elements/TextContentElement.vue
OliverHappe: you mean that the component has to implement an interface?

There is no way to do it without some trickery.

the solution i picked was giving all required props (atm, the element and isEditMode) as params to the elementStore. 

While the fields are used in the store it also helps with typing them correctly als element will have to be a known element type. :) 

unfortunately that still leaves the possibility to create an element without using useContentElementStore... but that would just be a bad decision we cannot really prevent.

If we want to be more verbose about props typing we might add another composable that only "defines" the props and call it `useExtendAnyContentElement` but so far i decided against it since i did not want to add boilerplate to elements. 

in general my main goal was to keep adding elements as easy as possible.

src/components/feature-board/types/Card.ts
OliverHappe: there will just be BoardCards that contain AnyContentElements.

Legacy stuff will be handled in sth like a LegacyLessonContentElement if we will handle it at all (i think there is no final decision yet)

src/components/feature-board/content-elements/TextContentElement.vue
christian-darsow: Yeap, it was about the need to implement an interface, actually in general
Totally agree, should not make it less easy or tricky :) And I like the approach. was just a question about if you have seen an approach to do that in vue somewhere in between

OliverHappe: so i saw this approach (composable consuming props to ensure specific prop names & types) in vue use and i think it is a really good compromise. 

see: https://vueuse.org/core/useVModel/

src/components/feature-board/state/CardState.composable.ts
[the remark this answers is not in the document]
OliverHappe: i think strings are easier to use as long as you have a decent IDE config giving you suggestions for the valid strings.

that does not mean that i would argue that importing things into the template is too much work :D

src/components/feature-board/state/ContentElementState.composable.ts
```diff
@@ -0,0 +1,56 @@
...
+import { AnyContentElement } from "../types/ContentElement";
+
+export const useContentElementState = <T extends AnyContentElement>(
+	props: {
+		element: T;
+		isEditMode: boolean;
+	},
+	options: { autoSaveDebounce?: number } = { autoSaveDebounce: 300 }
+) => {
+	useCardHostInteractionHandler(() => {
```
christian-darsow, line 13: this interactionHandler also handles mouse events for content components, right? or even better can be used on any component anywhere to check if e.g. was double-clicked? I just ask because I was confused at first sight what this host interaction got to do with element focus. Maybe renaming it to useCardInteractionHandler would make that clearer?

src/components/feature-board/content-elements/TextContentElement.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+	<div>
+		<TextContentElementDisplay
+			v-if="!isEditMode"
+			:value="modelValue.text"
+		></TextContentElementDisplay>
+		<TextContentElementEdit
```
christian-darsow, line 7: Well, its obviously smaller components this way to seperate into edit/view :) do you follow a rule of thumb why you decided so here ?

src/components/feature-board/state/ContentElementState.composable.ts
OliverHappe: while it is true that alle events that are triggered within the cards boundaries (so also on content-elements) the handler is specifically listening in the CardHost component. 

it captures and prevents some events to never reach underlying componens (esc keydown) while only catching others only when bubbling up. 

The name useCard**Host**InteractionHandler is supposed to refer to the origin of these interactions -> the CardHost component. 

when we add more interactions in the future (e.g. content-elements communicating with each other) we should add another handler to keep the features architecture composable and not couple it to tightly. 

tl;dr: the useCardHostInteractionHandler is specifically designed to handle events from the CardHost. not the whole card.

src/components/feature-board/content-elements/TextContentElement.vue
OliverHappe: yes. it might not be necesseray for a trivial element like this. but i mainly think that each element should be structured the same. one smart component (XxxContentElement) handling the business logic of how things are saved and how the element behaves (permissions etc will come later). one view component (XxxContentElementDisplay) to just render the content and another view component (XxxContentElementEdit) to show a form, handle validation etc. 

I hope that this pattern is adaptable to any complexity of ContentElements to keep them nicely organized :)

christian-darsow (approved):

PR #2465, opened 2023-03-27
BC-3008 - Delete a TaskCard

MartinSchuhmacher:
# Short Description
As a teacher i want to be able to delete a task.

BC-3008

## Changes
- implement option to delete beta task (and linked legacy task with it)
- add element - button at beginning of element list

## Checklist before merging


src/pages/TaskCard.page.vue
```diff
@@ -208,8 +248,10 @@ export default defineComponent({
 						title: taskCardData.courseName || "",
 					},
 				];
+				isDeletable.value = !!taskCardData.id;
```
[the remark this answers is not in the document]
MartinSchuhmacher: basically only if a taskCard does not have an ID, which in turn only happens if there is no taskCard at all. And then we also dont need the delete option.
That was my thought behind it :)

src/store/task-card.ts
```diff
@@ -105,6 +105,20 @@ export default class TaskCardModule extends VuexModule {
 		}
 	}
 
+	@Action
+	async deleteTaskCard(): Promise<void> {
```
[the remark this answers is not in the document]
MartinSchuhmacher: I can implement that if you think its better. To me I just thought that we do not use it, as we anyway have only the current taskCard in state so there can be no confusion or something. Was that the issue here?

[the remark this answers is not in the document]
christian-darsow: I would argue that these delete methods (like the fetch) in gerenal are more like static methods centralized in a store. as a method that can be used from a beta-task list (somewhere) . where we then just pass the id.

src/pages/TaskCard.page.vue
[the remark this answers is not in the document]
christian-darsow: I wonder if the whole form / component shouldnt behave different if there is no taskCard? kind of error handling if the id is eg wrong. I mean this check doesnt hurt, but maybe raises this question? what do you think?

src/store/task-card.ts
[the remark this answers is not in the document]
odalys-dataport: I agree with Christian :)

src/locales/de.json
```diff
@@ -797,6 +797,8 @@
 	"pages.taskCard.addElement": "Element hinzufügen",
 	"pages.taskCard.deleteElement.text": "Bist du dir sicher, dass du das Element löschen möchtest?",
 	"pages.taskCard.deleteElement.title": "Element löschen",
+	"pages.taskCard.deleteTaskCard.text": "Bist du dir sicher, dass du die Beta-Aufgabe löschen möchtest?",
```
odalys-dataport, line 800: I think it would be better if we'd mention the task title here like "Bist du dir sicher, dass du die Beta-Aufgabe "Titel Lorem ipsum" löschen möchtest?"

src/pages/TaskCard.page.vue
```diff
@@ -52,15 +56,34 @@
...
-					data-testid="save-btn"
+					v-if="isDeletable"
+					class="ml-auto"
+					color="secondary"
+					outlined
+					@click="openDeleteDialog()"
+					data-testid="delete-btn"
 				>
-					{{ $t("common.actions.save") }}
+					{{ $t("common.actions.remove") }}
```
odalys-dataport, line 66: I agree with Ankush here, that we should consistently use remove or delete

src/pages/TaskCard.page.vue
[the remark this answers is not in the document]
odalys-dataport: I think you're right, we're not checking if findTaskCard sets the data correctly or throws an error

src/pages/TaskCard.page.vue
```diff
@@ -344,6 +386,15 @@ export default defineComponent({
 			router.go(-1);
 		};
 
+		const deleteTaskCard = async () => {
+			if (hasErrors.value) {
```
odalys-dataport, line 390: Why are you checking for this. User should be able to delete an invalid task card as well I think. Or am I missunderstanding this?

odalys-dataport (changes requested):

src/store/task-card.ts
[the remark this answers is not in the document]
MartinSchuhmacher: if you both agree then I will just do so :D

src/pages/TaskCard.page.vue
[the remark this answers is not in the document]
MartinSchuhmacher: I will try to implement something that hopefully works as expected 👍

src/locales/de.json
MartinSchuhmacher: 👍

src/pages/TaskCard.page.vue
MartinSchuhmacher: 👍

src/pages/TaskCard.page.vue
MartinSchuhmacher: hmm...I had something in mind with the dateTimeInput...but yeah makes no sense to me neither anymore

src/store/task-card.unit.ts
```diff
@@ -265,7 +265,7 @@ describe("task-card store", () => {
 					);
 
 				taskCardModule.setTaskCardData(mockTaskCardData);
-				await taskCardModule.deleteTaskCard();
+				await taskCardModule.deleteTaskCard(mockTaskCardData.id);
```
christian-darsow, line 268: Just to clarify also in the test that the delete does not depend on the state I would suggest to remove these pieces, Line 267 etc

src/pages/TaskCard.page.vue
```diff
@@ -248,7 +251,9 @@ export default defineComponent({
 						title: taskCardData.courseName || "",
 					},
 				];
-				isDeletable.value = !!taskCardData.id;
+				if (taskCardData.id !== "") {
```
christian-darsow, line 254: Lets keep it like this for now I would say - but have a new ticket for handling unsuccessful beta-task fetching as it has nothing to do with the delete but will make this check unnecessary finally :)

MartinSchuhmacher: will be handled with error handling for id of task card (redirect to error page)

christian-darsow (approved):

PR #2472, opened 2023-03-31
BC-3509 - disable times in past

christian-darsow:
# Short Description

- Add option to restrict the time picker to not pick times in the past
- Adjust date- and time-validation to vuetify rules usage so that validate on submit gets triggered again
- Add loading state check on TaskCard.page

https://ticketsystem.dbildungscloud.de/browse/BC-3509

## Checklist before merging


src/pages/TaskCard.page.vue
```diff
@@ -236,7 +255,7 @@ export default defineComponent({
 
 				dueDate.value = endOfSchoolYear.toISOString();
 				initElements(initialCardElements);
-
+				isLoading.value = false;
```
odalys-dataport, line 258: shouldn't we rather compute a loading state out of the store loading states?

odalys-dataport (changes requested):

src/pages/TaskCard.page.vue
christian-darsow: Yeap, absolutely! Just forgot about that I wanted to use the new state you introduced recently. Will change that

src/components/date-time-picker/DatePicker.vue
```diff
@@ -130,51 +108,48 @@ export default defineComponent({
...
 
-		const onBlur = () => {
-			if (showDateDialog.value === false) {
-				validate(selectedDate.value);
-			}
-		};
-
-		const clearDate = () => {
-			selectedDate.value = "";
+		type Rule = (value: string | null) => boolean | string;
```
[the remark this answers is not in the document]
christian-darsow: Good idea!

src/pages/TaskCard.page.vue
```diff
@@ -76,6 +81,7 @@
 				has-buttons
 				confirm-btn-title-key="common.actions.remove"
 				@dialog-confirmed="deleteElement()"
+				:is-open="false"
```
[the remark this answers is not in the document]
christian-darsow: My IDE (and the component) tells me that this prop is required, I would like to keep it (and not get that error shown in IDE all the time)

src/components/date-time-picker/DatePicker.vue
```diff
@@ -18,21 +18,21 @@
...
 					filled
 					clearable
-					:error-messages="errors"
-					:error="hasErrors"
 					v-bind="attrs"
 					v-on="on"
+					:rules="rules"
+					validate-on-blur
+					autocomplete="off"
+					ref="inputfield"
```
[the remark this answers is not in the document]
christian-darsow: Yeap, change that

src/components/date-time-picker/DateTimePicker.vue
```diff
@@ -70,72 +74,74 @@ export default defineComponent({
...
+			return new Date(dateIsoString).toLocaleTimeString(locale, {
 				timeStyle: "short",
 				hourCycle: "h23",
 			});
 		};
 
-		if (props.dateTime !== "") {
-			setDateTime(props.dateTime);
-		}
+		const dateIsToday = (dateString: string) => {
```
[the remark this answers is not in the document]
christian-darsow: Thanks for the hint. Sure, lets use it

odalys-dataport (dismissed):

src/components/date-time-picker/DatePicker.vue
```diff
@@ -86,40 +92,16 @@ export default defineComponent({
...
-		const errors = ref<string[]>([]);
+		const inputField = ref<HTMLInputElement | null>(null);
 
 		const formattedDate = computed(() => {
-			return selectedDate.value
-				? dayjs(selectedDate.value).format(t("format.date"))
-				: selectedDate.value;
+			return model.value
+				? dayjs(model.value).format(t("format.date"))
+				: model.value;
```
MartinSchuhmacher, line 102: in case there is a value, it formats the value. otherwise it takes the value as it is.
the handling of the "else" case does not feel right to me

src/components/date-time-picker/DatePicker.vue
```diff
@@ -130,52 +112,72 @@ export default defineComponent({
...
+			}
+			return rules;
 		});
 
-		const resetErrors = () => {
-			errors.value = [];
+		const handleBlur = useDebounceFn(() => {
+			const date = model.value || "";
+			emit("input", date);
+		}, 200);
```
MartinSchuhmacher, line 133: why there is that arbitrary value of 200 set here? can that be based on some actual action being completed?

src/components/date-time-picker/DatePicker.vue
```diff
@@ -130,52 +112,72 @@ export default defineComponent({
...
+		const handleMenuToggle = () => {
+			if (showDateDialog.value) {
+				emit("valid");
 			}
-		});
+		};
+
+		const closeMenu = useDebounceFn(() => {
+			showDateDialog.value = false;
+		}, 50);
```
MartinSchuhmacher, line 152: same here as above

src/components/date-time-picker/DatePicker.vue
```diff
@@ -130,52 +112,72 @@ export default defineComponent({
...
 		};
 	},
 });
 </script>
+
+<style lang="scss" scoped>
+::v-deep {
+	.menu-open {
+		label {
+			transform: translateY(-6px) scale(0.75) !important;
```
MartinSchuhmacher, line 176: I'm no fan of that `important` property, please specify the class more if needed
also the specified values seem odd to me, in case of the scale maybe a proportional value can be set?

src/components/date-time-picker/DatePicker.vue
```diff
@@ -130,52 +112,72 @@ export default defineComponent({
...
+		}
+		.v-text-field__details {
+			display: none !important;
```
MartinSchuhmacher, line 179: same here as above with `important`

src/components/date-time-picker/DateTimePicker.unit.ts
```diff
@@ -1,32 +1,41 @@
...
-		},
-		propsData: props,
-		...options,
-	});
+type DateTimePickerProps = {
+	dateTime: string;
+	dateInputLabel?: string;
+	dateInputAriaLabel?: string;
+	minDate?: string;
+	maxDate?: string;
```
MartinSchuhmacher, line 11: as we work with TypeScript here, maybe we should work with Date types instead of just strings?

src/components/date-time-picker/DatePicker.vue
christian-darsow: do you mean it would be easier to read if the else part explicitly returns an empty string?

src/components/date-time-picker/DatePicker.vue
christian-darsow: Unfortunately there is no such other event/hook that could then trigger this function call, therefore it is debounced to have a smooth UI

src/components/date-time-picker/DatePicker.vue
christian-darsow: Good point, removed that by using vuetify vars here instead

src/components/date-time-picker/DatePicker.vue
christian-darsow: yeap, as above. Fixed

src/components/date-time-picker/DateTimePicker.unit.ts
christian-darsow: As talked about we take the date-handling, in detail the date-as-string handling to the next level for general discussion. for now we leave it as is in this PR

MartinSchuhmacher (approved):

PR #2503, opened 2023-04-24
BC-3860 - Fix code style duedate

MartinSchuhmacher:
# Short Description

in FE and BE the property for tasks and board tasks is defined as "duedate" and not (as it would be correct) "dueDate". This leads to development problems, as the issue can cause errors in case the dev assumes that it would be "dueDate" as in the database.

BC-3860
https://github.com/hpi-schul-cloud/schulcloud-server/pull/4080
## Changes

- `duedate` usage changed to `dueDate` (as in database property)

## Checklist before merging


src/store/types/room.ts
```diff
@@ -12,7 +12,7 @@ export type Task = {
 	name: string;
 	description?: string;
 	availableDate?: string;
-	duedate?: string;
+	dueDate?: string;
```
christian-darsow, line 15: Not related to this PR - but I really do not understand why we define the type "Task" here as well (also underneath in the next file).

christian-darsow (approved):

src/store/types/room.ts
MartinSchuhmacher: yeah, me neither

PR #2505, opened 2023-04-26
BC-3935 - Fix DateTimePicker bug

christian-darsow:
# Short Description

Fix DateTimePicker component that throws an error while trying to generate a Date object out of faulty time value

https://ticketsystem.dbildungscloud.de/browse/BC-3935

## Checklist before merging

### DIFF
--- a/src/components/date-time-picker/TimePicker.vue
+++ b/src/components/date-time-picker/TimePicker.vue
@@ -93,6 +93,8 @@ export default defineComponent({
 			toRef(props, "allowPast")
 		);
 
+		const timeRegex = /^([01][0-9]|2[0-3]):[0-5][0-9]$/g;
+
 		const requiredRule: ValidationRule = (value: string | null) => {
 			return value === "" || value === null
 				? t("components.timePicker.validation.required")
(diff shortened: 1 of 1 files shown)


MartinSchuhmacher (approved):

src/components/date-time-picker/TimePicker.vue
```diff
@@ -137,7 +138,9 @@ export default defineComponent({
 
 		const handleBlur = useDebounceFn(() => {
 			const time = model.value || "";
-			emit("input", time);
+			if (time.match(timeRegex)) {
```
odalys-dataport, line 141: I don't like that we have to check for the format here again.. What about the idea of only emitting input event when time value is valid?

odalys-dataport (changes requested):

src/components/date-time-picker/TimePicker.vue
christian-darsow: Well, I dont like that as well :) Unfortunetly, there is no event that gets triggered by vuetify after each validation - only when that error state changes (update:error). As by default that error state is false (Form valid) this does not suite to emit the input if no error occured before.
In the end this check was  introduced because the whole DateTimePicker turned falsy under the hood when Date changed to today and time is then in past - but: the TimePicker does not know that it is falsy when it never was in focus before. That whole validate-on-blur only kicks in when a field was focused - what makes sense... 
I wish I knew a better solution anyway - if you have another idea, I am more then happy

odalys-dataport (approved):

PR #2576, opened 2023-06-16
BC-4145 - Delete text element

christian-darsow:
# Short Description
Delete rich text element with keyboard

https://ticketsystem.dbildungscloud.de/browse/BC-4145

## Checklist before merging


src/components/feature-board/content-elements/RichTextContentElementEdit.vue
```diff
@@ -3,6 +3,7 @@
 		<ck-editor
 			v-model="modelValue"
 			@focus="onFocus"
+			@delete="onDelete"
```
MartinSchuhmacher, line 6: I was just asking myself if the naming should reflect the specific keyboard handling to not being confused with the "regular" deletion method. If that generalization is on purpose, then its fine :)

christian-darsow: I named it like that, because the CK itself (when used without vue etc) calls this event "delete". But you are right, this doesn't reflect at all that its "only" a keyboard event and might lead to confusion in our context. I rename that

MartinSchuhmacher (approved):

MartinSchuhmacher (approved):

PR #2577, opened 2023-06-19
Thr 20 create UI elements

marode-cap:
# Short Description

> ! This PR is not being merged into main, but into our feature branch !

This PR integrates the H5P editor into the vue client

https://github.com/hpi-schul-cloud/schulcloud-server/pull/4190
https://github.com/hpi-schul-cloud/schulcloud-client/pull/3211

## Changes
- Integrates H5P editor into vue frontend
- Adds a few new translations for en/de (We will collect all new keys and translate them at the end)
- Updates API docs

## Checklist before merging


src/pages/H5PEditor.page.vue
```diff
@@ -1,62 +1,185 @@
...
+			notifierModule.show({
+				text: t("common.validation.invalid"),
+				status: "error",
+				timeout: 10000,
+			});
+		}
+
+		async function onValidParams(event: ParamsValidEvent) {
+			try {
+				const response = await $axios.post<{
```
virgilchiriac, line 100: why axios? is there no store and then using the generated api, like other components?

src/layouts/loggedIn.layout.vue
```diff
@@ -36,13 +36,5 @@ export default {
...
 	width: inherit;
 	max-width: 100vw;
-
-	@include breakpoint(tablet) {
-		max-width: calc(100vw - var(--sidebar-width-tablet));
-	}
-
-	@include breakpoint(desktop) {
-		max-width: calc(100vw - var(--sidebar-width));
-	}
```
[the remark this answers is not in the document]
marode-cap: This should only affect pages that are displayed with `?inline=1` (So, the LernStore when opening it from a course topic).
The underlying layout already uses a grid where the sidebar takes up whatever it needs and the content takes up the rest, I don't see a reason to include redundant styles.

This fixes all the cases where a sidebar might not be present (e.g. when using `?inline=1`). In these cases the content is now actually centered

src/pages/H5PEditor.page.vue
```diff
@@ -1,62 +1,185 @@
...
-	setup(props) {
-		const loading = ref(props.loading);
-		let iframeSrc = ref(props.iframeSrc);
-		const scriptSrc = ref(props.scriptSrc);
+	setup() {
+		const i18n = inject<VueI18n | undefined>("i18n");
+
+		if (!i18n) {
+			throw new Error("Injection of dependencies failed");
+		}
```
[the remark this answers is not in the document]
marode-cap: Since we are currently developing on a feature-branch we do not have these changes yet and since we have multiple PRs open at the moment I'd like to avoid merging main into it. I've taken note of this and added it to our backlog and we will get to it at a later point

src/pages/H5PEditor.page.vue
marode-cap: There is no store because this is not part of the state.
But you are right on not using the generated API, I've updated the code to use it

src/layouts/loggedIn.layout.vue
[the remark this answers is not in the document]
christian-darsow: Hm. With that change e.g. on task overview page ( /tasks) the content on common tablet width 768px doesn't render as on main. It gets cutted on the right. To be honest, I also don't get the logic between that route param and the removed css style. Maybe I just don't understand the workflow though and it all turns out when the MVP is ready :)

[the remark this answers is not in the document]
marode-cap: The route param was introduced for the Lernstore as far as I know. It disables the sidebar, but this style assumes a sidebar will always be there.
But if this breaks existing pages, I'll revert it for now.

src/router/vue-client-route.js
```diff
@@ -38,6 +38,7 @@ const vueRoutes = [
 	`^/rooms-overview/?$`,
 	`^/rooms-list/?$`,
 	`^/h5p/player/${h5pId}/?$`,
+	`^/h5p/editor/?$`,
```
christian-darsow, line 41: what is that needed for? maybe that is not running on my local but I get a 404 error and also cant see anything related in routes.ts - sorry, I have no clue about the ticket you are working on and just try to figure out :)

marode-cap: This change allows access to the H5P editor page without an ID (When creating new content). It's the same page as the editor, which is why the route in routes.ts is the same (just with the ID being optional)
You probably get a 404 because the editor is a separate microservice on the backend.

src/pages/H5PEditor.page.vue
```diff
@@ -1,62 +1,209 @@
...
+		const router = useRouter();
+
+		const iframe = ref<HTMLIFrameElement>();
+
+		const contentId = route.params?.id;
+		const isInline = !!route.query?.inline;
+
+		const iframeSrc = `${window.location.origin}/api/v3/h5p-editor/edit/${
+			contentId ?? ""
+		}`;
```
christian-darsow, line 88: Last question: any chance to get that out of the way in this hardcoded style here? e.g. from that h5p editor api somehow?

christian-darsow (approved):

PR #2587, opened 2023-06-21
BC-4198 - Alert that students can see the board

muratmerdoglu-dp:
# Short Description
- The teachers should get an info message that the board is visible to all participants when they reach their column board.

https://ticketsystem.dbildungscloud.de/browse/BC-4198

## Checklist before merging


src/components/feature-board/board/Board.vue
```diff
@@ -150,6 +156,14 @@ export default defineComponent({
 		const onUpdateColumnTitle = async (columnId: string, newTitle: string) => {
 			if (hasEditPermission) await updateColumnTitle(columnId, newTitle);
 		};
+		onMounted(() => {
+			if (isTeacher) {
```
uidp, line 160: I'm wondering if it's really relevant that the user is a teacher, I think the really important information is if the user is able to edit the board - regardless of the role. What do you think?

src/components/feature-board/types/Board.ts
```diff
@@ -16,4 +16,5 @@ export type BoardPermissionsTypes = {
 	hasCreateColumnPermission?: boolean;
 	hasEditPermission?: boolean;
 	hasDeletePermission?: boolean;
+	isTeacher?: boolean;
```
uidp, line 19: Why are these optional? And can we maybe find another name for that type? I think it is a type (interface) that bundles a couple of board permission checks, right? So maybe something like `BoardPermissionChecks` would be more appropriate IMHO.

src/components/feature-board/board/Board.vue
muratmerdoglu-dp: We don't have any board-specific roles yet. We may use the `course_edit` permission, but admin users also have the  `course_edit` permission. That's why I've chosen if the user is a `teacher` or not.

muratmerdoglu-dp: But since the admin already can not reach the board we can also use `course_edit` permission as well. What is your suggestion?

uidp (approved):

PR #2593, opened 2023-06-22
BC-4143 - text by enter

christian-darsow:
# Short Description
Create a new text element as the first element and focus it when hitting enter key within title field

https://ticketsystem.dbildungscloud.de/browse/BC-4143

## Checklist before merging


odalys-dataport (dismissed):

src/components/feature-board/shared/BoardAnyTitleInput.vue
```diff
@@ -103,12 +111,23 @@ export default defineComponent({
 			}
 		});
 
+		const onEnter = ($event: KeyboardEvent) => {
+			if (props.scope !== "card") return;
+			if (!textarea.value) return;
+
+			if (textarea.value.value.length === textarea.value.selectionStart) {
```
MartinSchuhmacher, line 118: `textarea.value.value` sounds really weird code wise. can we set `textarea.value` into another variable and make it more readable? also for line 116 then?

MartinSchuhmacher (changes requested):

src/components/feature-board/shared/BoardAnyTitleInput.vue
christian-darsow: hm. the first `value` is for the reactive part. the second is then the property "value" of that reactive object ([HTMLTextAreaElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement)). So somewhere in the code I guess we will always have that .value.value Do you have an idea how to make it more readable though?

MartinSchuhmacher: if its like that then maybe something like `const reactiveTextarea = textarea.value` might fit as a first thought

christian-darsow: To be honest, I don't like this idea very much. This `.value` syntax tells me that something is reactive. I always watch out for this syntax to know whether something is reactive or not - even before reading a vars name - what do you think about an alternative: a computed prop thats name states that it presents the textareas value length?

MartinSchuhmacher: sounds good to me 👍

christian-darsow: I moved the length calculation to a seperat function. This still needs that weird looking `.value.value` but makes it more readable by the function name I hope

MartinSchuhmacher: I agree

PR #2638, opened 2023-07-07
BC-4085 - Add Edit Menu on Board Card

muratmerdoglu-dp:
# Short Description
- Users should be able to put a card into edit mode via the 3dot menu of the card

- https://ticketsystem.dbildungscloud.de/browse/BC-4085

## Checklist before merging


src/components/feature-board/card/CardTitle.unit.ts
```diff
@@ -11,7 +16,7 @@ describe(CardTitle.name, () => {
 		wrapper = shallowMount(CardTitle as MountOptions<Vue>, {
 			...createComponentMocks({}),
 			propsData: {
-				value: "props value",
+				...componentProps,
```
NFriedo, line 19: Is there a reason for this not to be inlined?

src/components/feature-board/shared/BoardAnyTitleInput.vue
```diff
@@ -69,11 +80,14 @@ export default defineComponent({
 			if (!textarea.value) return;
 			textarea.value.focus();
 		};
+		onMounted(() => {
+			if (props.isFocused) setFocusOnEdit();
```
NFriedo, line 84: Should we add a test for this?

src/components/feature-board/shared/InlineEditInteractionHandler.vue
```diff
@@ -31,8 +32,16 @@ export default defineComponent({
...
-			if (props.isEditMode) {
+
+		const checkEventTarget = (event: MouseEvent): boolean => {
+			if (!(event.target instanceof HTMLElement)) return false;
+
+			return event.target?.className?.includes("v-list-item");
+		};
+
+		const onClickOutside = (event: MouseEvent) => {
+			const isMenuEvent = checkEventTarget(event);
```
NFriedo, line 43: Do we have to check this? If we are in edit mode and click on edit again this will loose focus 🤔
Anyways I think it's very fragile to select the target by className.

src/components/feature-board/card/CardTitle.unit.ts
muratmerdoglu-dp: `isFocused` prop is already introduced on component before but is not passed in the test file, so I just thought to define the all props outside. There is no other reason :)

src/components/feature-board/shared/InlineEditInteractionHandler.vue
muratmerdoglu-dp: You are right it is odd to select by class name. It is complicated. The menus in the cardhost's behaviors are related the removing the `capturing: false`. In this removing case, every mouse clicks outside triggers the `end-edit-mode` emit even if the clicks are on the menu items. Considering both cases, there were no more options not to select by class name.

hoeppner-dataport (approved):

PR #2669, opened 2023-07-20
BC-4372 - Create Submission Element on Card

MartinSchuhmacher:
# Short Description

As a teacher, I want my students to be able to submit their solution to a task.
Implement empty submission element for now.

BC-4372
## Checklist before merging


src/components/feature-board-submission-element/SubmissionContentElement.vue
```diff
@@ -0,0 +1,110 @@
...
+				v-if="!isEditMode"
+				:dueDate="element.content.dueDate"
+			/>
+			<SubmissionContentElementEdit
+				v-if="isEditMode"
+				:dueDate="element.content.dueDate"
+				:isFirstElement="isFirstElement"
+				:isLastElement="isLastElement"
+				:hasMultipleElements="hasMultipleElements"
+				@move-down:element="onMoveFileEditDown"
```
christian-darsow, line 24: Is this naming on purpose? the "file" part? Also below and in the setup?

src/components/feature-board-submission-element/SubmissionContentElementDisplay.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+	<div class="grey lighten-4 px-4 py-3 rounded-t-sm">
+		<v-list-item
```
christian-darsow, line 3: Is there more to come (more items) or is there another reason why we use a list here? Just as i am curious :)

src/components/feature-board-submission-element/SubmissionContentElementEdit.vue
```diff
@@ -0,0 +1,85 @@
+<template>
+	<div class="grey lighten-4 px-4 py-3 rounded-t-sm">
+		<v-list-item
```
christian-darsow, line 3: Same as in the view: did we choose the list (the list-ite, without a list as wrapper) for some UX reasons? or maybe this came from main and "is like that" for all elemtens now?

src/components/feature-board-submission-element/SubmissionContentElement.vue
MartinSchuhmacher: I will check that again, might be a mistake on adapting the structure ^^

src/components/feature-board/shared/AddElementDialog.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<vCustomDialog
 		data-testid="element-type-selection"
-		:size="320"
+		:size="426"
```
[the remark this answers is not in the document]
MartinSchuhmacher: this is basically the minimum with needed to show all three items...I did not had any other option as the custom dialog needs to receive a number here and I could not just set it to "auto" or something...

src/components/feature-board-submission-element/SubmissionContentElementDisplay.vue
MartinSchuhmacher: basically, as you said below, this is only because it was used in the beginning and I just adapted it here ^^ but you are right, to me it also does not make sense to use the item without a proper list. or the list comes actually from using it in the content element list now that I think about it...

christian-darsow (approved):

PR #2725, opened 2023-08-09
BC-4737 - Remove users property from legacy task clean up task card in general

MartinSchuhmacher:
# Short Description

BC-4737
https://github.com/hpi-schul-cloud/schulcloud-server/pull/4316
https://github.com/hpi-schul-cloud/schulcloud-client/pull/3260
https://github.com/hpi-schul-cloud/dof_app_deploy/pull/602

## Changes
- delete complete task card module from code
- backend and frontend 
- delete mongo collections 
- delete seed data
- delete permissions with prefix TASK_CARD_*** from code, seed and production db
- delete implementation of course endpoint only used for task-card BC-3691
- delete implementation of task filter for assigned users to legacy task BC-3411
- delete implementation of task assignments for legacy task BC-2393

## Checklist before merging


odalys-dataport (changes requested):
Found a translation key that needs to be removed as well

src/components/templates/TasksDashboardMain.vue
```diff
@@ -254,39 +254,10 @@ export default {
 			};
 		},
 		fabItems() {
-			const actions = [];
 			if (
 				!this.isStudent &&
```
odalys-dataport, line 258: Just a thought I'm having: We do have the method "isTeacher" as well, maybe that' nicer to read. But also checking for the permission should suffice here actually, no?

MartinSchuhmacher: the condition for "at least not a student" was present in the past and reflects the current behaviour. the "new" permission check was actually implemented by me during the implementation of beta task and covers additionally to the existing behaviour as a second security gate. so I would like to keep it like that to not do an additional refactoring here tbh

odalys-dataport: Yeah I just mentioned it because this has been a topic multiple times (brought up by Cedric), that we should not check for roles but permissions instead and that there was a case where students where able to add tasks (on nbc I think) before this page was implemented. We do not have to change anything about this right now, I just wanted to mention it

odalys-dataport (dismissed):

christian-darsow (dismissed):

christian-darsow (approved):

PR #2742, opened 2023-08-17
BC-4853 - refactor common components

christian-darsow:
# Short Description
Refactor common components like CKEditor to also follow the new block structure

https://ticketsystem.dbildungscloud.de/browse/BC-4853

## Checklist before merging


src/components/feature-common-date-time-picker/index.ts
```diff
@@ -0,0 +1,3 @@
+import DateTimePickerVue from ".DateTimePicker.vue";
+
+export { DateTimePickerVue };
```
MartinSchuhmacher, line 3: why `DateTimePickerVue` and not just `DateTimePicker` ?

MartinSchuhmacher (commented):
In general I'm not sure about the `common` usage and if its needed. Why not just use `feature-date-time-picker`, `feature-editor` and `feature-render-html` ? just so I understand it better

src/components/feature-common-date-time-picker/index.ts
christian-darsow: That was not on purpose... fixed that typo

christian-darsow:
> In general I'm not sure about the `common` usage and if its needed. Why not just use `feature-date-time-picker`, `feature-editor` and `feature-render-html` ? just so I understand it better

Hm, well, like on the folder structure (before) I wanted to show, that all these three features belong somehow together and with that "common" indicate that they are somehow global and shared. In contrast  to e.g. all feature-board-* components. But as you ask now and I think again, that must really not be needed at all... and everything without something like "board" or "common" is just "global". ... I will change it and remove that common

MartinSchuhmacher (approved):

PR #2779, opened 2023-08-30
BC-4350 - mark task completed

christian-darsow:
# Short Description
Mark a task as completed or not-completed as a student

https://ticketsystem.dbildungscloud.de/browse/BC-4350
https://github.com/hpi-schul-cloud/schulcloud-server/pull/4380

## Checklist before merging


src/components/feature-board-submission-element/SubmissionContentElement.vue
```diff
@@ -97,13 +99,19 @@ export default defineComponent({
 			}
 		};
 
+		const updateCompleted = (completed: boolean) => {
```
MartinSchuhmacher, line 102: how about `updateCompletionState`?

src/components/feature-board-submission-element/SubmissionContentElementDisplay.vue
```diff
@@ -1,53 +1,95 @@
...
+		const isStudent = computed(() => {
+			return userRoles.value.includes("student");
+		});
+
+		const updateCompleted = (completed: boolean) => {
+			emit("update:completed", completed);
+		};
+
+		const editable = computed(() => {
+			return dayjs().isBefore(toRef(props, "dueDate").value);
```
MartinSchuhmacher, line 85: is this supposed to mean `isEditable` if its a state instead of just `editable`?
also the return line seems a bit crowded.
how about:
`const dueDate = toRef(props, "dueDate").value`
and then
`const today = dayjs()` or `... dayjs(new Date())`
together with
`return today.isBefore(dueDate)`
What do you think about it ?

src/components/feature-board-submission-element/SubmissionContentElementState.composable.unit.ts
```diff
@@ -0,0 +1,117 @@
...
+		mockedUseSubmissionItemApiCalls =
+			createMock<ReturnType<typeof useSubmissionItemApi>>();
+		mockedUseSubmissionItemApi.mockReturnValue(mockedUseSubmissionItemApiCalls);
+	});
+
+	afterEach(() => {
+		jest.resetAllMocks();
+	});
+
+	const setup = (contentElementId = "123123") => {
```
MartinSchuhmacher, line 35: what is the reason for defining the ID as a parameter? 
I was not able to figure out why this should be separately given to the setup function each time as an argument instead of just building it during runtime with the invoke of the setup function and a constant ID within that function.

src/components/feature-board-submission-element/SubmissionItemApi.composable.unit.ts
```diff
@@ -0,0 +1,64 @@
...
+			expect(
+				submissionItemApi.boardSubmissionControllerUpdateSubmissionItem
+			).toHaveBeenCalledWith(submissionItemId, {
+				completed: completed,
+			});
+		});
+	});
+
+	describe("fetch submissionItems", () => {
+		it("should call boardSubmissionControllerUpdateSubmissionItem api", async () => {
```
MartinSchuhmacher, line 54: guess this is not what it should call ^^

MartinSchuhmacher (changes requested):

src/components/feature-board-submission-element/SubmissionContentElementDisplay.vue
christian-darsow: I agree, even though not that hard to get, the other way is even easier to understand. I adjusted that as you proposed. For the other question I think we are fine without the "is" and "editable" means "this can be edited"

src/components/feature-board-submission-element/SubmissionItemApi.composable.unit.ts
christian-darsow: ooops - that was the wrong copy / paste :) fixed that

src/components/feature-board-submission-element/SubmissionContentElementState.composable.unit.ts
christian-darsow: I added it as a parameter as I needed that ID value also in the expect parts, eg. here https://github.com/hpi-schul-cloud/nuxt-client/pull/2779/files#diff-1c82236680ddcfd69ce3d9162f5077f3e755054ada37781b8a1a2ce718ced3d7R52   We could also create it inside the setup function and return it from there.. but that doesn't make it cleaner in my opinion

MartinSchuhmacher: yeah I agree...the last option also came into my mind but just produces more code on other places instead ^^

src/components/feature-board-submission-element/SubmissionContentElementDisplay.vue
MartinSchuhmacher: I thought more in perspective of general code style and remembering about if its being a state and the question "is it this or that" then it should be with the "is". But thats only my assumption so if you say its fine, then we are fine :)

MartinSchuhmacher (approved):

tests/test-utils/factory/submissionItemResponseFactory.ts
```diff
@@ -0,0 +1,12 @@
+import { SubmissionItemResponse } from "@/serverApi/v3";
+import { Factory } from "fishery";
+import { timestampsResponseFactory } from "@@/tests/test-utils/factory";
+import { userDataResponseFactory } from "@@/tests/test-utils/factory";
```
MartinSchuhmacher, line 4: kannst du das hier noch fixen? (also den Import zusammenfassen)

PR #2786, opened 2023-09-04
BC-4623-Creation time and update time of status messages

Michaellinaresxk:
# Short Description

https://ticketsystem.dbildungscloud.de/browse/BC-4623

## Changes
Added functinalities to update the date format to present the update and created according those requierments:

- From 1 to 59 minutes - display "updated [...] minutes ago | created [...] minutes ago". If value=1, the word "minute" is in singular.

- From 1 hour to 23h 59min - display "updated [...] hours ago | created [...] hours ago". If the value=1, the word "hour" is in singular.

- From 1 to 6d 23h 59min days - display "updated [...] days ago | created [...] days ago". If value=1, the word "day" is in singular.

- From 7 days and more - display the date of update in European format | date of creation in European format. For example updated 28.07.2023 | created 21.07.2023 or created 21.07.2023.

The update time is displayed first, then the creation time. The creation time is always displayed, the update time is displayed if there was an update.

## Screenshots of UI changes

<!
Created 1 day ago
created 7 days ago
created 1 minute ago

-->

## Checklist before merging


bn-pass (changes requested):

bn-pass (dismissed):

src/plugins/datetime.js
```diff
@@ -237,17 +237,46 @@ export const createInputDateTime = (date) => {
...
  * @param {boolean} isLocalTimeZone set true if input date is to be handled in local time zone instead of utc
  * @return {String} Date difference based on current timezone
  */
 export const fromNow = (date, isLocalTimeZone) => {
-	if (isLocalTimeZone == true) {
-		return dayjs(date).fromNow();
+	const time = isLocalTimeZone ? dayjs(date) : dayjs.tz(date, "UTC");
+	const current = isLocalTimeZone ? dayjs() : dayjs.utc();
+
+	const totalMinutesDiff = Math.abs(current.diff(time, "minute"));
```
virgilchiriac, line 258: days, minutes, hours, ago, etc this must also be translated
is there no library to handle this cases and support internationalization?

Michaellinaresxk: - dd.mm.yyyy it will be the new implementation about date time in european format.
 - I will deal with: days, minutes, hours, ago internationalization.

virgilchiriac (changes requested):
I never heard about european date format. At the same time, for example, we are working on a date picker, and display the date in vue, and everywhere else in the application (nuxt client and legacy too) we consistently use localised date formats.  We cannot / should not use diferent date formats.

virgilchiriac (commented):
FYI we are working on a date picker, and display the date in vue, where we used localised date formats.  We cannot / should not use diferent date formats - but this seems to be the only exception

src/locales/de.json
```diff
@@ -25,6 +25,11 @@
 	"common.labels.admin": "Admin(s)",
 	"common.labels.updateAt": "Aktualisiert:",
 	"common.labels.createAt": "Erstellt:",
+	"common.labels.minutes": "Vor Minuten",
+	"common.labels.hour": "Vor einer Stunde",
+	"common.labels.hours": "Vor Stunden",
+	"common.labels.day": "vor einem Tag",
+	"common.labels.days": "vor Tagen",
```
virgilchiriac, line 32: cannot test it, but how would this be be rendered as "vor 5 Tagen"? 
5 vor Tagen is not correct!
Also, sometimes Uppercase, other times lowercase...

virgilchiriac (changes requested):

src/plugins/datetime.js
odalys-dataport: We do not want to use the european format (which would be YYYY-MM-DD), but using specific formats for each language. You can find these date formats in the translation files under "date.format".

odalys-dataport: Ok, I clarified this with UX. For now we want to prioritise consistency over localisation. So we're going with the DD.MM.YYYY format.

Sorry for the hassle!

src/components/topbar/StatusAlerts.vue
```diff
@@ -34,10 +34,23 @@
...
+							$t(formatDate(item.timestamp).key)
+						}}
+						|
+					</template>
 					{{ $t("common.labels.createAt") }}
-					{{ getCreatedDate(item.createdAt) }}
+					{{
+						formatDate(item.createdAt).value
+							? formatDate(item.createdAt).value +
+							  " " +
```
odalys-dataport, line 50: Please move expressions like this into the script

src/locales/de.json
odalys-dataport: Agreed, you can use parameters in translations to fix this

src/locales/de.json
```diff
@@ -25,6 +25,11 @@
 	"common.labels.admin": "Admin(s)",
 	"common.labels.updateAt": "Aktualisiert:",
 	"common.labels.createAt": "Erstellt:",
+	"common.labels.minutes": "Vor Minuten",
+	"common.labels.hour": "Vor einer Stunde",
```
odalys-dataport, line 29: Please be more specific with the translation keys and change it so something like "hoursAgo", but as Virgil mentioned we should be able to remove these and use dayjs translations instead

src/plugins/datetime.js
```diff
@@ -237,19 +237,55 @@ export const createInputDateTime = (date) => {
...
+	const totalDaysDiff = Math.abs(current.diff(time, "day"));
+
+	if (totalDaysDiff < MAX_DAYS_BEFORE_SHOWING_FULL_DATE) {
+		return {
+			key: totalDaysDiff === 1 ? "common.labels.day" : "common.labels.days",
+			value: totalDaysDiff,
+		};
+	}
+
+	return {
```
odalys-dataport, line 284: This is a pretty big change, have you checked that all occurrences where the "fromNow" function is used are still working?

odalys-dataport (changes requested):

src/plugins/datetime.js
Michaellinaresxk: Thank you for your clarification.

odalys-dataport (approved):

odalys-dataport (approved):

src/plugins/datetime.js
```diff
@@ -252,6 +252,29 @@ export const fromNow = (date, isLocalTimeZone) => {
 	return fromUTC(date).fromNow();
 };
 
+export const formatDateForAlerts = (date, isLocalTimeZone = false) => {
+	const time = isLocalTimeZone ? dayjs(date) : dayjs.tz(date, "UTC");
+	const current = isLocalTimeZone ? dayjs() : dayjs.utc();
+
+	setDayjsLocale();
```
bn-pass, line 259: @Michaellinaresxk is there any reason to call this function on every `formatDateForAlerts()` function call? Wouldn't it be sufficient to just call it once for every time the script is loaded (as it looks like it configures the dayjs in its global scope, I don't see any relation to this function params)?

PR #2808, opened 2023-09-18
N21-1206 show classes and goups

MBergCap:
# Short Description
adds a page for class overview

https://ticketsystem.dbildungscloud.de/browse/N21-1206
https://github.com/hpi-schul-cloud/schulcloud-server/pull/4406
https://github.com/hpi-schul-cloud/schulcloud-client/pull/3305
https://github.com/hpi-schul-cloud/dof_app_deploy/pull/643

## Checklist before merging


IgorCapCoder (commented):
missing unit tests ;)

src/store/group.ts
```diff
@@ -0,0 +1,142 @@
...
+@Module({
+	name: "groupModule",
+	namespaced: true,
+	stateFactory: true,
+})
+export default class GroupModule extends VuexModule {
+	private classes: ClassInfo[] = [];
+
+	private loading = false;
+	private error: object | null = null;
```
arnegns, line 23: Please use BusinessError. Take a look into external tools store

src/store/group.ts
```diff
@@ -0,0 +1,142 @@
...
+
+	private pagination: Pagination = {
```
arnegns, line 25: Why you save pagination in the store? We dont get these from the server api. We only need it in the page / component. Please move it

src/store/group.ts
```diff
@@ -0,0 +1,142 @@
...
+		skip: 0,
+		total: 30,
+	};
+
+	private sortBy = "name";
+	private sortOrder: SortOrder = SortOrder.ASC;
+	private page = 1;
+
+	private get groupApi(): GroupApiInterface {
+		return GroupApiFactory(undefined, "/v3", $axios);
```
arnegns, line 36: There was no decision for the new structure (see data-board) until now. We should use composables to access the api instead of vuex stores because they are more like the pinia stores which could be used in the future.

So for now okay, but please keep in mind to use the new structure especially api access mechanism

src/store/group.ts
```diff
@@ -0,0 +1,142 @@
...
+	@Mutation
+	setPage(page: number): void {
+		this.page = page;
+	}
+
+	@Action
+	async loadClassesForSchool(): Promise<void> {
+		this.setLoading(true);
+		try {
+			const sortBy =
```
arnegns, line 111: When you move the pagination and sorting you should use them as parameters here

src/store/group.ts
```diff
@@ -0,0 +1,142 @@
...
+			);
+
+			this.setPagination({
+				limit: response.data.limit,
+				skip: response.data.skip,
+				total: response.data.total,
+			});
+			this.setClasses(mappedClasses);
+		} catch (error) {
+			this.setError(error as Error);
```
arnegns, line 138: Please use the error mechanism like in externaltools

src/store/types/sort-order.enum.ts
```diff
@@ -0,0 +1,4 @@
+export enum SortOrder {
```
arnegns, line 1: Please use a new folder "group" for the types for this store

src/pages/administration/ClassOverview.page.vue
```diff
@@ -0,0 +1,149 @@
...
+
+		const classes: ComputedRef<ClassInfo[]> = computed(
+			() => groupModule.getClasses
+		);
+
+		const pagination: ComputedRef<Pagination> = computed(
+			() => groupModule.getPagination
+		);
+
+		const sortBy: ComputedRef<string> = computed(() => groupModule.getSortBy);
```
arnegns, line 79: Please verify this and ask bb.. Should sortBy and pagination in the store? So when the user leaves the page and open it up again, should the sorting and pagination be the same?

arnegns (dismissed):

src/pages/administration/ClassOverview.page.vue
```diff
@@ -0,0 +1,149 @@
...
+			:headers="headers"
+			:items="classes"
+			:items-per-page.sync="pagination.limit"
+			:server-items-length="pagination.total"
+			:sort-by="sortBy"
+			:sort-Order="sortOrder"
+			:page="page"
+			:footer-props="footerProps"
+			data-testid="admin-class-table"
+			class="custom-table elevation-1"
```
MarvinOehlerkingCap, line 18: is there a requirement to change the color or did you copy it from somewhere?

src/pages/administration/ClassOverview.page.vue
```diff
@@ -0,0 +1,149 @@
...
+			groupModule.setPagination({ ...pagination.value, skip });
+			await groupModule.loadClassesForSchool();
+		};
+		const onUpdateItemsPerPage = async (itemsPerPage: number) => {
+			groupModule.setPagination({ ...pagination.value, limit: itemsPerPage });
+			await groupModule.loadClassesForSchool();
+		};
+
+		onMounted(() => {
+			groupModule.loadClassesForSchool();
```
MarvinOehlerkingCap, line 124: I would recommend to await this, even if the linter says that onMounted does not expect a promise

src/store/types/sort-order.enum.ts
MarvinOehlerkingCap: sort order looks like a very general type that can be used everywhere, not only for groups

src/store/group.ts
MarvinOehlerkingCap: Compared to the external tools store, I would recommend `BusinessError | null` as the type to better distinguish between having an error and not having one.

src/store/group.ts
```diff
@@ -0,0 +1,142 @@
...
+					: undefined;
+
+			const response: AxiosResponse<ClassInfoSearchListResponse> =
+				await this.groupApi.groupControllerFindClassesForSchool(
+					this.pagination.skip,
+					this.pagination.limit,
+					this.getSortOrder,
+					sortBy
+				);
+			const mappedClasses: ClassInfo[] = response.data.data.map(
```
MarvinOehlerkingCap, line 123: please export this to a separate mapper (see context external tools store)

src/store/group.unit.ts
```diff
@@ -0,0 +1,220 @@
...
+
+				return {
+					response,
+					sortBy,
+					sortOrder,
+					pagination,
+				};
+			};
+
+			it("should update the classes", async () => {
```
MarvinOehlerkingCap, line 186: you are missing a test for expecting that the store state has changed

src/utils/sidebar-base-items.ts
```diff
@@ -177,6 +177,13 @@ const getSidebarItems = (
 				testId: "Klassen",
 				activeForUrls: ["^/administration/classes($|/.*)"],
 			},
+			{
```
MarvinOehlerkingCap, line 180: you are missing the feature flag

MarvinOehlerkingCap (changes requested):

src/pages/administration/ClassOverview.page.vue
MBergCap: this is the color of the table header from the legacy client. There was no specific requirement. But this is how it looks like in the old client

MarvinOehlerkingCap: we should use one style in the vue client and not copy legacy styles imo

src/utils/sidebar-base-items.ts
MBergCap: a function for the feature has been added

src/store/group.unit.ts
MBergCap: fixed

src/store/group.ts
MBergCap: fixed

src/pages/administration/ClassOverview.page.vue
MBergCap: fixed

src/pages/administration/ClassOverview.page.vue
MBergCap: custom-table class removed

IgorCapCoder (approved):

src/pages/administration/ClassOverview.page.vue
MBergCap: After consultation with bb, the sorting and pagination can be saved like this

src/store/group.ts
MBergCap: After consultation with bb, the sorting and pagination can be saved like this

PR #2906, opened 2023-11-10
BC-5739 - TimePicker Select Highlighting

odalys-dataport:
# Short Description

https://ticketsystem.dbildungscloud.de/browse/BC-5739

Translation File changes for EN, ES and UK are aligned with changes in DE. They show changes in the whole file, because formatting was messed up in a prior PR.

## Checklist before merging


src/locales/en.json
```diff
@@ -141,12 +141,12 @@
 	"common.words.languages.en": "English",
 	"common.words.languages.es": "Spanish",
 	"common.words.languages.uk": "Ukrainian",
-	"components.datePicker.messages.future": "Please specify a date and time in the future.",
+	"components.datePicker.messages.dateInPast": "The date and time are in the past.",
+	"components.datePicker.messages.dateRequired": "Time not possible without date.",
```
MartinSchuhmacher, line 145: I find the wording of "Time not possible without date" a bit unclear. Because "a time" is possible without a date, even though I know what its supposed to imply.

I would suggest something like `Time specification not possible without date` or sth similar

odalys-dataport: True, I adjusted the wording

MartinSchuhmacher (approved):

src/components/ui-date-time-picker/TimePicker.vue
```diff
@@ -126,6 +128,18 @@ export default defineComponent({
 			showTimeDialog.value = false;
 		}, 50);
 
+		const scrollToSelectedItem = useDebounceFn(() => {
+			if (!showTimeDialog.value) return;
+
+			const menu = document.getElementsByClassName(
```
MartinSchuhmacher, line 134: could we get that as a unique ID instead of a class name?

odalys-dataport: I would have loved to do that, but for some reason you can't set an id to the menu. Maybe it works with using a ref, I will try that out

src/components/ui-date-time-picker/TimePickerState.composable.ts
christian-darsow, line 1: Cant we kick this file completely? No need to have it in main, if not needed at all, I would suggest.

odalys-dataport: I'm leaving it in for now, because I still wanna test if this works in Vue3. If it doesn't, we can still delete it at a later point.

christian-darsow (approved):

PR #2941, opened 2023-12-05
BC-5933 - update typescript eslint

odalys-dataport:
# Short Description

https://ticketsystem.dbildungscloud.de/browse/BC-5933

Rules are disabled where it's not sure, if removing those lines would break something.

## Checklist before merging


src/pages/files/file-table-utils.composable.ts
```diff
@@ -102,7 +102,6 @@ export function useFileTableUtils(
 		const paramsArray: string[] = route.params.catchAll
 			.split("/")
 			.filter((element: string) => element !== "");
-		const teamsPath: string = route.path.replace("/cfiles/teams", "");
```
arnegns, line 105: Why you removed teamsPath?

arnegns (approved):

src/router/guards/multi.guard.unit.ts
```diff
@@ -138,8 +138,7 @@ describe("Multiguard", () => {
 
 			const multiguard = Multiguard([
 				(to, from, next) => next(expected),
-				(to, from, next) =>
-					fail("Second guard called when it shouldn't have been"),
+				() => fail("Second guard called when it shouldn't have been"),
```
christian-darsow, line 141: As above... we are changing the signature now "only" to satisfy the linter. from reading test first, I might think I can call `Multiguard` like that. but I can't. personally I would prefer to have an ignore here as well, and not change the signature

src/store/types/collaborative-file.ts
```diff
@@ -1,7 +1,7 @@
 export enum CollaborativeFileType {
 	FAVORITES = "favorites",
 	FILE = "draft",
-	PDF = "draft",
+	PDF = "pdf",
```
christian-darsow, line 4: I guess this was just a fix, you now did in your PR :)

christian-darsow (approved):

PR #2943, opened 2023-12-06
BC-5881 - replace tiptap

christian-darsow:
# Short Description
Replace tiptap in news section with custom CK5

https://ticketsystem.dbildungscloud.de/browse/BC-5881

## Checklist before merging


src/components/feature-editor/config.ts
```diff
@@ -0,0 +1,155 @@
...
+	"math",
+	"horizontalLine",
+	"|",
+	"blockQuote",
+	"insertTable",
+	"specialCharacters",
+	"removeFormat",
+];
+
+const newsToolbar = [
```
virgilchiriac, line 42: in BE we decided to avoid having input formats specific for an entity such as news
so we have CK5 (used in board), and CK5 simple (used in news). Besides Ck4

src/components/feature-editor/config.ts
```diff
@@ -0,0 +1,155 @@
...
+	"Essentials",
+	"Heading",
+	"Image",
+	"ImageInsertViaUrl",
+	"Italic",
+	"List",
+	"Paragraph",
+	"RemoveFormat",
+	"Strikethrough",
+];
```
virgilchiriac, line 95: see above...

src/components/feature-editor/config.ts
christian-darsow: ok, I didnt know that. is this good? what happens when we update to CK6 in FE...?

src/components/feature-editor/CKEditor.vue
```diff
@@ -71,99 +80,31 @@ export default defineComponent({
...
-						{
-							model: "heading2",
-							view: "h5",
-							title: "Heading 2",
-							class: "ck-heading_heading2",
-						},
-					],
-				},
+				plugins: plugins[props.mode],
+				heading: headings[props.mode],
```
uidp, line 107: Just a thought: Maybe it would also be nice to have an actual `config` object per mode and be able to get that by the mode key, e.g.
```typescript
const editorConfigs = {
  news: {
    plugins: [],
    headings: [],
    // ...
  }, // ...

const config = {
  ...editorConfigs[props.mode],
  // ...
}
```
I'm not sure if that is really necessary though. The current solution maybe fine as it is.

src/components/feature-editor/config.ts
uidp: I hope I understand this correctly, maybe we shouldn't name the modes after entities but after something more general "mode-like"?

uidp (approved):

PR #2951, opened 2023-12-11
N21-1583 copying ctl tools info

mrikallab:
# Short Description

This PR adjust information text in ShareModal, CopyModal and ImportModal, when feature for ctl tools tab is enabled.

https://ticketsystem.dbildungscloud.de/browse/N21-1583

## Checklist before merging


src/components/copy-result-modal/CopyResultModal.vue
```diff
@@ -154,6 +163,11 @@ export default {
 				: "";
 			return `${courseFilesText} ${fileErrorText}`.trim();
 		},
+		externalToolsInfoText() {
+			return this.hasFeatureCtlsToolsenabled
```
IgorCapCoder, line 167: you already test for the feature flag in the "isShow" property which hangs in an v-if statement in the template. So without the feature flag the whole modal would not be shown. Is there a reason why you would like to do it twice?

src/components/share/ImportModal.vue
```diff
@@ -21,7 +21,17 @@
...
 						<v-icon color="info">{{ mdiInformation }}</v-icon>
 					</div>
-					<div>
+					<div
+						v-if="ctlToolsEnabled"
+						data-testid="import-modal-external-tools-info"
+					>
+						{{
+							$t(
+								`components.molecules.import.courses.options.ctlTools.infoText`
```
IgorCapCoder, line 30: it seems like you could avoid the v-if all together if you would set $(parentType) to "course" and change the language key according to that

src/components/copy-result-modal/CopyResultModal.unit.ts
```diff
@@ -127,6 +135,22 @@ describe("@/components/copy-result-modal/CopyResultModal", () => {
 			);
 		});
 
+		it("should render ctl tools info if root item is a Course and has no failed file ", () => {
+			const copyResultItems = mockResultItems([]);
+			envConfigModule.setEnvs({ FEATURE_CTL_TOOLS_TAB_ENABLED: true } as Envs);
```
IgorCapCoder, line 140: what do you think about a test with feature = false

src/components/share/ImportModal.unit.ts
```diff
@@ -122,4 +126,26 @@ describe("@components/share/ImportModal", () => {
 		const emitted = wrapper.emitted("import");
 		expect(emitted).toBeUndefined();
 	});
+
+	describe("when ctl tools are enabled", () => {
```
IgorCapCoder, line 130: with feature = false you would expect $t(`components.molecules.import.${parentType}.options.infoText`). How do you feel about testing that?

src/components/share/ShareModal.vue
```diff
@@ -26,6 +26,16 @@
 							{{ $t(`components.molecules.share.${type}.options.infoText`) }}
 							<br />
 							{{ $t("components.molecules.copyResult.courseFiles.info") }}
+							<div
+								v-if="ctlToolsEnabled"
+								data-testid="share-modal-external-tools-info"
+							>
+								{{
+									$t(
+										`components.molecules.share.courses.options.ctlTools.infotext`
```
IgorCapCoder, line 35: see above: is it possible to use the already provided structure?

IgorCapCoder (changes requested):

src/components/copy-result-modal/CopyResultModal.unit.ts
mrikallab: The test of the deactivated feature is above.

src/components/share/ImportModal.vue
mrikallab: I can not avoid the v-if, need to check on the feature flag

IgorCapCoder (dismissed):

IgorCapCoder (approved):

PR #3012, opened 2024-01-17
BC-6187 - Vue3 upgrade: fix components/share tests

NFriedo:
# Short Description

Fixes components/share tests. Also fixes a bug in the SelectCourseModal.vue, which lead to an error when importing tasks or lessons.

BC-6187

## Checklist before merging

### DIFF
--- a/src/components/share/ImportFlow.unit.ts
+++ b/src/components/share/ImportFlow.unit.ts
@@ -25,6 +25,7 @@ import {
 	createTestingI18n,
 	createTestingVuetify,
 } from "@@/tests/test-utils/setup";
+import vueDompurifyHTMLPlugin from "vue-dompurify-html";
 
 describe("@components/share/ImportFlow", () => {
 	let copyModuleMock: CopyModule;
@@ -47,14 +48,17 @@ describe("@components/share/ImportFlow", () => {
 
 		const wrapper = mount(ImportFlow, {
 			global: {
-				plugins: [createTestingVuetify(), createTestingI18n()],
+				plugins: [
+					createTestingVuetify(),
+					createTestingI18n(),
+					vueDompurifyHTMLPlugin,
+				],
 				provide: {
 					copyModule: copyModuleMock,
 					loadingStateModule: loadingStateModuleMock,
 					[NOTIFIER_MODULE_KEY.valueOf()]: notifierModule,
 					[ENV_CONFIG_MODULE_KEY.valueOf()]: envConfigModuleMock,
 				},
-				stubs: ["RenderHTML"],
 			},
 			props: {
 				token,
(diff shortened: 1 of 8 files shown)


src/components/share/ImportFlow.unit.ts
```diff
@@ -40,36 +39,35 @@ describe("@components/share/ImportFlow", () => {
...
+		const wrapper = mount(ImportFlow, {
+			global: {
+				plugins: [createTestingVuetify(), createTestingI18n()],
+				provide: {
+					copyModule: copyModuleMock,
+					loadingStateModule: loadingStateModuleMock,
+					[NOTIFIER_MODULE_KEY.valueOf()]: notifierModule,
+					[ENV_CONFIG_MODULE_KEY.valueOf()]: envConfigModuleMock,
+				},
+				stubs: ["RenderHTML"],
```
uidp, line 57: Alternatively, when we want to check the html output we can use `plugins: [vueDompurifyHTMLPlugin]`. Maybe you already knew...

src/components/share/ImportFlow.unit.ts
```diff
@@ -84,27 +82,28 @@ describe("@components/share/ImportFlow", () => {
 
 	describe("token is provided", () => {
 		it("should render with props", () => {
-			const wrapper = mountComponent();
+			const wrapper = setup();
```
uidp, line 85: I think this is not 100% correct because we are returning `{ wrappper }` from `setup()`

src/components/share/ShareModalOptionsForm.unit.ts
```diff
@@ -1,42 +1,48 @@
...
+	it("should emit event on changes of isSchoolInternal checkbox", async () => {
+		const { wrapper } = setup();
 
-		const switchIsSchoolInternal = wrapper.find(
+		const checkboxIsSchoolInternal = wrapper.findComponent(
 			'[data-testid="isSchoolInternal"]'
 		);
-		await switchIsSchoolInternal.setChecked(false);
+
+		await checkboxIsSchoolInternal.setValue(false);
```
uidp, line 34: Does that work on a component instance? I thought we'd have to get the HTML input like:
```
const checkboxIsSchoolInternal = wrapper.find(
	'[data-testid="isSchoolInternal"]'
).get('input[type="checkbox"]');
```
If it works thanks for finding out :-)

muratmerdoglu-dp (approved):

src/components/share/ImportFlow.unit.ts
NFriedo: I didn't know. Thats really helpful, thank you!

src/components/share/ImportFlow.unit.ts
NFriedo: You are right, I missed this one and will fix it

src/components/share/ShareModalOptionsForm.unit.ts
NFriedo: Yes this is working. I tried it in different tests and it worked there :)

PR #3037, opened 2024-01-23
BC-6192 - fix date time picker tests

odalys-dataport:
# Short Description

https://ticketsystem.dbildungscloud.de/browse/BC-6192

## Checklist before merging

### DIFF
--- a/src/components/ui-date-time-picker/DatePicker.unit.ts
+++ b/src/components/ui-date-time-picker/DatePicker.unit.ts
@@ -1,4 +1,3 @@
-import { nextTick } from "vue";
 import { ComponentMountingOptions, mount } from "@vue/test-utils";
 import {
 	createTestingI18n,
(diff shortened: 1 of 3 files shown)


src/components/ui-date-time-picker/DatePicker.unit.ts
```diff
@@ -1,52 +1,34 @@
...
-				[I18N_KEY.valueOf()]: i18nMock,
+	const mountComponent = (
+		options: ComponentMountingOptions<typeof DatePicker> = {}
+	) => {
+		return mount(DatePicker, {
+			global: {
+				plugins: [createTestingVuetify(), createTestingI18n()],
 			},
-			attachTo: "#root",
+			...options,
```
MartinSchuhmacher, line 16: I don't think putting every other property next to `props` into `options` is a good way to handle it. That way something general like `mocks` would appear as it also needs to be set with each instanciated wrapper. 
I would prefer putting it like `props` here in the mount, giving that as an attribute to the `mountComponent` and then calling it for each test with the parameter object `props` as needed.

uidp: Well, it depends on what you want to achieve. Sometimes you want to have more control over the whole options object. Although a general pattern would be nice. Currently we have different approaches depending on the specific tests.

uidp (approved):

PR #3068, opened 2024-02-01
BC-6362 - content element bar fix

hoeppner-dataport:
# Short Description
Fix usage and internal structure of content-element-bar.

BC-6362

## Checklist before merging


src/components/ui-board/LineClamp.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<div ref="textElement" class="line-clamp">
 		<v-tooltip
-			top
+			location="top"
 			start
```
NFriedo, line 5: I think we can remove the `start` prop in line 5 when we want to have the `location="top" :)
It is ignored anyways because in vuetify 3 it would be `location="start"` but seems like we want the top location :D

src/components/feature-board/board/BoardColumn.vue
```diff
@@ -262,6 +263,7 @@ export default defineComponent({
 .sortable-drag-ghost .v-card {
 	opacity: 0.6;
 	background-color: rgba(var(--v-theme-secondary-lighten-1));
+	width: 350px;
```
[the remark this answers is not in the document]
hoeppner-dataport: replaced :)

src/components/ui-board/LineClamp.vue
hoeppner-dataport: removed ✅

src/components/ui-board/content-element/ContentElementBar.vue
```diff
@@ -1,112 +1,106 @@
...
+				:class="{
+					'mr-10': hasSlotContent($slots.menu) && !$slots.display,
+				}"
 			>
 				<div v-if="$slots.logo" class="logo-container mr-2">
 					<slot name="logo" />
 				</div>
 
-				<ContentElementTitleIcon v-if="icon" :icon="icon" class="mr-2" />
+				<v-icon v-if="icon" color="black" size="20" class="mr-2">{{
```
NFriedo, line 29: Since vuetify 3 we could also write:
```html
<v-icon :icon="icon" />
```
but it also works like this :D

src/components/ui-board/content-element/ContentElementBar.vue
```diff
@@ -1,112 +1,106 @@
...
-				icon ||
-				$slots.logo ||
-				$slots.title ||
-				$slots.menu ||
-				$slots.element ||
-				$slots.subtitle ||
-				$slots.description
+				$slots.title || $slots.element || $slots.subtitle || $slots.description
 			"
+			:class="dynamicClasses"
```
NFriedo, line 15: Do we need a computed property here or would it be better to inline it?
```suggestion
			:class="{ 'bg-grey-lighten-4': props.hasGreyBackground }"
```

src/components/ui-board/content-element/ContentElementBar.vue
```diff
@@ -1,112 +1,106 @@
...
-<style scoped type="text/scss">
-.font-narrow {
+<style scoped lang="scss">
+.content-element-title {
 	font-family: var(--font-accent) !important;
+	font-size: 1rem !important;
+	font-weight: 700;
+
+	:deep(a) {
+		color: var(--v-black-base) !important;
```
NFriedo, line 88: I was just wondering if we want to use our theme black color:
`color: rgba(var(--v-theme-black));`

But it seems like it has the same value like the --v-black-base, so I am not sure what to use :D

src/components/ui-board/content-element/ContentElementBar.vue
hoeppner-dataport: changed that to get everything in one line :+1:

src/components/ui-board/content-element/ContentElementBar.vue
hoeppner-dataport: the computed does not make sense any longer ✅

src/components/ui-board/content-element/ContentElementBar.vue
hoeppner-dataport: replaced white v-theme-black ✅

PR #3083, opened 2024-02-08
BC-6523 - news cancel error

virgilchiriac:
# Short Description
BC-6523

## Checklist before merging

### DIFF
--- a/src/pages/NewsCreate.page.vue
+++ b/src/pages/NewsCreate.page.vue
@@ -77,8 +77,9 @@ export default {
 			}
 		},
 		async cancelHandler() {
+			console.log("cancelHandler");
 			this.$router.push({
-				name: "news",
+				name: "news-cancel",
 			});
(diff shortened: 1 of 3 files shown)


src/pages/NewsCreate.page.vue
```diff
@@ -77,8 +77,9 @@ export default {
 			}
 		},
 		async cancelHandler() {
+			console.log("cancelHandler");
 			this.$router.push({
-				name: "news",
+				name: "news-cancel",
```
virgilchiriac, line 82: this is a bit hacky, but wondering, why was it working before, because the news path itself is in legacy-client

odalys-dataport: I guess it wasn't actually working before, because on canceling you would be redirected to /dashboard

odalys-dataport (approved):

PR #3159, opened 2024-03-20
BC-6794 Add check if column board has drawing as its child

blazejpass:
# Short Description
For copy course functionality if there is whiteboard in column board we need to display a message about not copying the content of drawing.

https://ticketsystem.dbildungscloud.de/browse/BC-6794
https://github.com/hpi-schul-cloud/schulcloud-server/pull/4850

## Checklist before merging


src/store/room.unit.ts
```diff
@@ -349,6 +349,46 @@ describe("room module", () => {
 			});
 		});
 
+		describe("fetchHasDrawingChild", () => {
+			it("should call api to get info if has drawing child", async () => {
+				const mockApi = {
```
wiaderwek, line 354: move to setup function, add additional description for those it with description like "when room has drawing child"

src/store/room.unit.ts
```diff
@@ -349,6 +349,46 @@ describe("room module", () => {
...
+
+				expect(mockApi.boardControllerHasDrawingChild).toHaveBeenCalledTimes(1);
+				expect(mockApi.boardControllerHasDrawingChild).toHaveBeenCalledWith(
+					"id"
+				);
+
+				spy.mockRestore();
+			});
+
+			it("should catch error in catch block", async () => {
```
wiaderwek, line 373: same here

src/store/room.unit.ts
```diff
@@ -630,6 +670,16 @@ describe("room module", () => {
 				expect(roomModule.getPermissionData).toStrictEqual(expectedPermissions);
 			});
 		});
+
+		describe("setHasDrawingChild", () => {
+			it("should set the state", () => {
+				const roomModule = new RoomModule({});
```
wiaderwek, line 676: move to setup function

wiaderwek (changes requested):

src/store/room.unit.ts
blazejpass: done

src/store/room.unit.ts
blazejpass: done

src/store/room.unit.ts
blazejpass: done

wiaderwek (approved):

src/components/copy-result-modal/CopyResultModal.vue
```diff
@@ -151,6 +156,9 @@ export default {
 				CopyApiResponseTypeEnum.LessonContentNexboard
 			);
 		},
+		hasDrawingElement() {
+			return this.isCourse && roomModule.getHasDrawingChild;
```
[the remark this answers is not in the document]
blazejpass: I'm not sure, because request is providing with info about only new copied course and result failed copied elements.
It is done in copy.ts -> copy()->setCopyResult() and setCopyResultFailedItems()

wiaderwek (approved):

src/store/room.ts
```diff
@@ -346,6 +363,10 @@ export default class RoomModule extends VuexModule {
 		return this.roomData.roomId;
 	}
 
+	get getHasDrawingChild(): boolean {
```
CeEv, line 366: I am unsure about this change. Drawing or not sound not like what we want on this level. I see that also code from task and lesson is placed in this store. But its feel for me wrong.

muratmerdoglu-dp: I think there are three options:

1- Current implementation. At room-details level there is no info if the course has a white-board element or not, the client has to make an extra call to the server again if the course has.

2- At the copy-service level(backend). The backend copy service should send the proper info about the course that already has the element. So that there is no need for extra calling at the client level.

3- Only showing the info in the UI to the users that something like 'the white-board element which is probably included in this course was not copied'.

blazejpass: For second option we would have to add property with just information if course copy has drawing child.
image

Metauriel: the copy algorithm in the backend builds a tree of all things in the course that are copied. However, in the mapping to the response object, the tree is currently discarded for courses (but I think not for Tasks or Lessons).

It would be possible to change that by adding the children to the response mapping. Then in the frontend you could do a quick scan of the tree to see if there is an element of the relevant type

wiaderwek (dismissed):

davwas (approved):

wiaderwek (dismissed):

davwas (dismissed):

src/store/copy.unit.ts
```diff
@@ -191,6 +224,32 @@ describe("copy module", () => {
 			});
 		});
 
+		describe("checkDrawingChildren", () => {
+			describe("when drawing element is not among types in passed property", () => {
```
muratmerdoglu-dp, line 228: What it should do `"when drawing element is not among types in passed property"`.
I think using `it('should do ....)` pattern will be clear and more readable.

src/store/copy.unit.ts
```diff
@@ -191,6 +224,32 @@ describe("copy module", () => {
...
+				const payload: CopyApiResponseElementsTypesEnum[] = [
+					CopyApiResponseElementsTypesEnum.Board,
+					CopyApiResponseElementsTypesEnum.Card,
+					CopyApiResponseElementsTypesEnum.Column,
+				];
+
+				copyModule.checkDrawingChildren(payload);
+				expect(copyModule.getHasDrawingChild).toBe(false);
+			});
+			describe("when drawing element is among types in passed property", () => {
```
muratmerdoglu-dp, line 239: Same above

muratmerdoglu-dp (changes requested):

muratmerdoglu-dp (approved):

PR #3193, opened 2024-04-10
BC- 7030 - new sidebar

odalys-dataport:
# Short Description

Adds browser tooltips to sidebar items

https://github.com/hpi-schul-cloud/schulcloud-client/pull/3439
https://ticketsystem.dbildungscloud.de/browse/BC-7030

## Checklist before merging


src/modules/ui/layout/SidebarCategoryItem.vue
```diff
@@ -10,7 +10,7 @@
 				<template #prepend>
 					<v-icon :icon="item.icon" class="mr-2" />
 				</template>
-				<v-list-item-title>
+				<v-list-item-title :title="$t(item.title)">
```
MartinSchuhmacher, line 13: script setup uses only `t(..)` instead of `$t(...)` doesn't it ?

src/modules/ui/layout/SidebarItem.vue
```diff
@@ -12,7 +12,7 @@
 		<template #prepend>
 			<v-icon v-if="hasIcon(item)" :icon="item.icon" class="mr-2" />
 		</template>
-		<v-list-item-title>
+		<v-list-item-title :title="$t(item.title)">
```
MartinSchuhmacher, line 15: see above

src/modules/ui/layout/SidebarCategoryItem.vue
odalys-dataport: We can use both. By using $t here I don't need to inject it just to use it in the template. If I'd use the t function in the script as well, I would use t in the template too, but I don't think it matters much. Mayne @uidp can share his thoughts on this.

MartinSchuhmacher: fine for me :)

MartinSchuhmacher (approved):

MartinSchuhmacher (approved):

MartinSchuhmacher (approved):

PR #3258, opened 2024-05-24
N21-1971 cancel migration wizard

mrikallab:
# Short Description
This PR will introduce cancelation button and modal on migration wizard to cancel the whole migration process.

- https://github.com/hpi-schul-cloud/schulcloud-server/pull/5026
- https://ticketsystem.dbildungscloud.de/browse/N21-1971

## Checklist before merging

### DIFF
--- a/src/pages/administration/Migration.page.vue
+++ b/src/pages/administration/Migration.page.vue
@@ -74,6 +74,16 @@
 						{{ t("pages.administration.migration.step5") }}
 					</VStepperItem>
 				</VStepperHeader>
+				<vCustomDialog
+					:is-open="isCancelDialogOpen"
+					has-buttons
+					@dialog-canceled="isCancelDialogOpen = !isCancelDialogOpen"
+					@dialog-confirmed="confirmCancelMigration"
+					data-test-id="cancel-migration-dialog"
+				>
+					<template #title> ASDFewre</template>
+					<template #content>fdgdfgfdg</template>
+				</vCustomDialog>
 			</VStepper>
 		</template>
 
@@ -160,19 +170,29 @@
 
 					<VStepperWindowItem :value="2" data-testid="migration_importUsers">
 						<ImportUsers />
-						<div class="text-right">
-							<VBtn @click="migrationStep = 1">
-								{{ t("pages.administration.migration.back") }}
-							</VBtn>
-							<VBtn
-								class="ml-2"
-								id="migration_importUsers_next"
-								color="primary"
-								:disabled="!canPerformMigration"
-								@click="migrationStep = 3"
-							>
-								{{ t("pages.administration.migration.next") }}
-							</VBtn>
+						<div>
+							<div class="float-left">
+								<VBtn
+									@click="cancelMigration"
+									data-test-id="importUsers-cancel-migration-btn"
+								>
+									{{ t("common.actions.cancel") }}
+								</VBtn>
+							</div>
+							<div class="float-right">
+								<VBtn @click="migrationStep = 1">
+									{{ t("pages.administration.migration.back") }}
+								</VBtn>
+								<VBtn
+									class="ml-2"
+									id="migration_importUsers_next"
+									color="primary"
+									:disabled="!canPerformMigration"
+									@click="migrationStep = 3"
+								>
+									{{ t("pages.administration.migration.next") }}
+								</VBtn>
+							</div>
 						</div>
 					</VStepperWindowItem>
 
@@ -214,7 +234,15 @@
 											/>
 										</VRow>
 									</VCardText>
-									<div class="text-right">
+									<div class="float-left">
+										<VBtn
+											@click="cancelMigration"
+											data-test-id="summary-cancel-migration-btn"
+										>
+											{{ t("common.actions.cancel") }}
+										</VBtn>
+									</div>
+									<div class="float-right">
 										<VBtn :disabled="isLoading" @click="migrationStep = 2"
 											>{{ t("pages.administration.migration.back") }}
 										</VBtn>
@@ -485,6 +513,7 @@ import {
 } from "vue";
 import { useI18n } from "vue-i18n";
 import { useRouter } from "vue-router";
+import VCustomDialog from "../../components/organisms/vCustomDialog.vue";
 
 const { t } = useI18n();
 
@@ -499,6 +528,8 @@ const errorTimeout: Ref<number> = ref(7500);
 const isLoading: Ref<boolean> = ref(false);
 
 const checkTotal: Ref<NodeJS.Timeout | undefined> = ref(undefined);
+
+const isCancelDialogOpen: Ref<boolean> = ref(false);
 
 const isMigrationNotStarted = computed(() => {
 	return school.value.inUserMigration === undefined;
@@ -701,6 +732,14 @@ const nextStep = () => {
 	}
 
 	migrationStep.value = nextStep;
+};
+
+const cancelMigration = () => {
+	isCancelDialogOpen.value = true;
+};
+
+const confirmCancelMigration = () => {
+	return console.log("CONFIRMED");
 };
 
 watch(migrationStep, async (val) => {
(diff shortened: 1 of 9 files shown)


src/pages/administration/Migration.page.vue
```diff
@@ -74,6 +74,28 @@
 						{{ t("pages.administration.migration.step5") }}
 					</VStepperItem>
 				</VStepperHeader>
+				<vCustomDialog
+					:is-open="isCancelDialogOpen"
```
MarvinOehlerkingCap, line 78: `:is-open="isCancelDialogOpen"` + `@update:is-open="isCancelDialogOpen = $event"` = `v-model:is-open="isCancelDialogOpen"`

MarvinOehlerkingCap: du darfst auch gerne das `isDialogOpen` im vCustomDialog zu einem
```
const isOpen: ModelRef<boolean> = defineModel("isOpen", {
	type: Boolean,
	required: true,
});
```
ändern.
(isOpen und update:isOpen muss dann aus den props und emits raus)

src/pages/administration/Migration.page.vue
```diff
@@ -485,6 +525,8 @@ import {
 } from "vue";
 import { useI18n } from "vue-i18n";
 import { useRouter } from "vue-router";
+import VCustomDialog from "../../components/organisms/vCustomDialog.vue";
+import router from "../../router";
```
MarvinOehlerkingCap, line 529: please use the useRouter composable from "vue-router";

src/pages/administration/Migration.page.vue
```diff
@@ -485,6 +525,8 @@ import {
 } from "vue";
 import { useI18n } from "vue-i18n";
 import { useRouter } from "vue-router";
+import VCustomDialog from "../../components/organisms/vCustomDialog.vue";
```
MarvinOehlerkingCap, line 528: `@/components/organisms/vCustomDialog.vue`

src/pages/administration/Migration.page.vue
```diff
@@ -214,7 +246,15 @@
 											/>
 										</VRow>
 									</VCardText>
-									<div class="text-right">
+									<div class="float-left">
```
MarvinOehlerkingCap, line 249: please use flex over float, since flex is responsive

src/pages/administration/Migration.page.vue
```diff
@@ -160,19 +182,29 @@
...
-								class="ml-2"
-								id="migration_importUsers_next"
-								color="primary"
-								:disabled="!canPerformMigration"
-								@click="migrationStep = 3"
-							>
-								{{ t("pages.administration.migration.next") }}
-							</VBtn>
+						<div>
+							<div class="float-left">
```
MarvinOehlerkingCap, line 186: please use flex over float, since flex is responsive

src/pages/administration/Migration.page.vue
```diff
@@ -558,6 +602,11 @@ const sourceSystemName = computed(() => {
 	}
 });
 
+const schoolSettingsPage: Breadcrumb = {
```
MarvinOehlerkingCap, line 605: I don't understand why you create a `Breadcrumb` object when you only need the `to` attribute

src/pages/administration/migration.unit.ts
```diff
@@ -38,6 +39,9 @@ const getWrapper = (
 			provide: {
 				[THEME_KEY.valueOf()]: $theme,
 			},
+			mocks: {
```
MarvinOehlerkingCap, line 42: I don't think you need to add this since you never use $theme in the template

src/pages/administration/Migration.page.vue
```diff
@@ -160,19 +182,29 @@
...
-								:disabled="!canPerformMigration"
-								@click="migrationStep = 3"
-							>
-								{{ t("pages.administration.migration.next") }}
-							</VBtn>
+						<div>
+							<div class="float-left">
+								<VBtn
+									@click="cancelMigration()"
+									data-testid="importUsers-cancel-migration-btn"
```
MarvinOehlerkingCap, line 189: camelCase or kebab-case, but not both.

src/pages/administration/Migration.page.vue
```diff
@@ -74,6 +74,28 @@
 						{{ t("pages.administration.migration.step5") }}
 					</VStepperItem>
 				</VStepperHeader>
+				<vCustomDialog
+					:is-open="isCancelDialogOpen"
+					has-buttons
```
MarvinOehlerkingCap, line 79: please also add the available buttons explicitly (for safety)

MarvinOehlerkingCap (changes requested):

src/pages/administration/migration.unit.ts
mrikallab: $theme is the instance name and is used in every step in the template.. if I delete the mock, then all tests are failing, because we need to read the name of the instance...

MarvinOehlerkingCap: The theme is provided by `const theme = injectStrict(THEME_KEY);` which is mocked by the provide above. There is no use of $theme from the global vue instance in the component. When it does not work without, then there is something else not working and we can look over it together.

src/components/organisms/vCustomDialog.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<v-dialog
 		ref="vDialog"
-		v-model="isDialogOpen"
```
MarvinOehlerkingCap, line 4: This should stay a `v-model`.
A v-model is a two-way binding that allows a parent component to change the property of a child (like a normal prop) and the child to also change the value and notify the parent, while only using one single reference.
By using `:model-value` you make this a one-way binding, where only the parent can enter a value, but all changes from the child are discarded, since the `update:model-value` event is not used.

src/components/organisms/vCustomDialog.vue
```diff
@@ -81,14 +81,10 @@
 </template>
 
 <script setup lang="ts">
-import { computed, PropType } from "vue";
+import { ModelRef, PropType } from "vue";
+import { size } from "lodash";
 
 const props = defineProps({
-	isOpen: { type: Boolean, required: true },
-	size: {
```
MarvinOehlerkingCap, line 88: you also removed size and replaced it with a random import. I think this is just slip-up.

src/pages/administration/Migration.page.vue
```diff
@@ -160,19 +183,29 @@
...
-								class="ml-2"
-								id="migration_importUsers_next"
-								color="primary"
-								:disabled="!canPerformMigration"
-								@click="migrationStep = 3"
-							>
-								{{ t("pages.administration.migration.next") }}
-							</VBtn>
+						<div class="flex-container">
+							<div class="">
```
MarvinOehlerkingCap, line 187: if you don't use a class then there is no need for `class=""`

src/pages/administration/Migration.page.vue
```diff
@@ -750,4 +817,9 @@ iframe.full {
 	overflow: scroll;
 	border: none;
 }
+
+.flex-container {
```
MarvinOehlerkingCap, line 821: please use the classes that are already provided by vuetify before defining custom ones. https://vuetifyjs.com/en/styles/flex/

src/pages/administration/Migration.page.vue
```diff
@@ -214,7 +247,15 @@
 											/>
 										</VRow>
 									</VCardText>
-									<div class="text-right">
+									<div class="float-left">
```
MarvinOehlerkingCap, line 250: you are still using floats instead of flex

src/pages/administration/Migration.page.vue
```diff
@@ -485,9 +526,12 @@ import {
 } from "vue";
 import { useI18n } from "vue-i18n";
 import { useRouter } from "vue-router";
+import VCustomDialog from "@/components/organisms/vCustomDialog.vue";
 
 const { t } = useI18n();
 
+const router = useRouter();
```
MarvinOehlerkingCap, line 533: you can remove the line in the `onMounted` where a second router is defined

MarvinOehlerkingCap (changes requested):

MarvinOehlerkingCap (approved):

MarvinOehlerkingCap (approved):

PR #3297, opened 2024-06-21
BC-7096 - Upgrade frontend libraries after Vue3 upgrade

uidp:
# Short Description

After the upgrade to Vue 3 we want to upgrade the frontend dependencies to the newest versions, most importantly:

- Typescript to v5
- Jest to v29
- Webpack and webpack-plugin-vuetify

The frontend is currently built using Webpack, and the configuration is auto-generated by Vue-CLI which is now deprecated and prevents us from upgrading other libraries.

- BC-7096
- BC-4475

## Checklist before merging


package.json
```diff
@@ -4,12 +4,12 @@
 	"author": "Dataport Bildungscloud",
 	"private": true,
 	"scripts": {
-		"serve": "vue-cli-service serve",
-		"build": "vue-cli-service build --mode production",
+		"serve": "NODE_ENV=development webpack serve --config config/webpack/webpack.dev.js",
+		"build": "NODE_ENV=production webpack --config config/webpack/webpack.prod.js",
 		"test": "npm run test:unit",
-		"test:unit": "vue-cli-service test:unit",
+		"test:unit": "npx jest",
```
odalys-dataport, line 10: Maybe we can remove `npm run test:unit`and only have `npm run test` instead?

uidp: The task is used by the GitHub test workflow. If we can change that to just `npm test` this might be possible. Not entirely sure though.
https://github.com/hpi-schul-cloud/nuxt-client/blob/5bb2d777fac31690e15c224c01e905bbe3feee6f/.github/workflows/test.yml#L25

MartinSchuhmacher: And then using `test:unit:ci` instead for the GitHub flow? I would think that having `npm run test` would be nice and then using `test:unit:ci` for the GitHub flow to have distinct separation for both

src/components/templates/RoomDashboard.unit.ts
```diff
@@ -487,7 +487,7 @@ describe("@/components/templates/RoomDashboard.vue", () => {
 			await nextTick();
 			expect(deleteTaskMock).not.toHaveBeenCalled();
 			expect(deleteLessonMock).toHaveBeenCalled();
-			expect(fetchContentMock).not.toHaveBeenCalled();
+			expect(fetchContentMock).toHaveBeenCalled();
```
virgilchiriac, line 490: I find this change strange...

src/main.ts
```diff
@@ -191,5 +191,9 @@ app.use(VueDOMPurifyHTML, {
 	);
 	app.provide(THEME_KEY.valueOf(), themeConfig);
 
+	// FIXME wait for a small period of time to make sure vue-router is initialized
+	// otherwise it is trying to match the '/' route and not the provided web history route
+	await delay(500);
```
virgilchiriac, line 196: should try to find a solution before merge (otherwise, a followup ticket),  because I can imagine, can lead to unexpected behaviour for some devices / low internet bandwidth etc

src/modules/data/board/ContentElementState.composable.ts
```diff
@@ -20,7 +20,7 @@ export const useContentElementState = <T extends AnyContentElement>(
 	const isLoading = ref<boolean>(false);
 
 	watchDebounced<T["content"]>(
-		modelValue.value,
+		modelValue,
```
virgilchiriac, line 23: can this make elements not to be updated?

package.json
uidp: Yes, agreed. I also think we need both.

src/modules/data/board/ContentElementState.composable.ts
uidp: Thanks, has been fixed.

src/components/templates/RoomDashboard.unit.ts
uidp: Actually `fetchContent()` is called after deleing any item. https://github.com/hpi-schul-cloud/nuxt-client/blob/24c72197df8d327e5933422d028093363b50b48d/src/components/templates/RoomDashboard.vue#L367
So this is expected. Maybe the test setup was not correct in the first place.

src/main.ts
uidp: I removed the delay. Application still initialized properly. Maybe we only needed the startup delay at an intermediate stage.

src/modules/data/board/ContentElementState.composable.ts
```diff
@@ -13,7 +13,7 @@ export const useContentElementState = <T extends AnyContentElement>(
 	const cardStore = useCardStore();
 	const _elementRef: Ref<T> = toRef(props, "element");
 
-	const modelValue = ref<T["content"]>(unref<T>(_elementRef).content);
+	const modelValue = ref(_elementRef.value.content);
```
MartinSchuhmacher, line 16: whats the reason for putting `_` in front of the variable instead of just `elementRef`?
I just see a maybe occuring typo incoming if somebody doesn't take that into consideration as we do not use `_` that often in code in my opinion.

src/components/lern-store/ContentEduSharingFooter.unit.ts
```diff
@@ -21,7 +21,7 @@ describe("@/components/molecules/ContentEduSharingFooter", () => {
 		);
 
 		expect(wrapper.find(".edusharing-footer__logo").attributes("src")).toBe(
-			wrapper.vm.img
+			"Mocked media file"
```
MartinSchuhmacher, line 24: hard coded necessary?

src/modules/feature/board-drawing-element/DrawingContentElement.unit.ts
```diff
@@ -82,7 +82,7 @@ describe("DrawingContentElement", () => {
 
 		describe("when arrow key up is pressed", () => {
 			describe("when component is in edit-mode", () => {
-				it('should NOT emit "move-keyboard:edit"', async () => {
+				it('should emit "move-keyboard:edit"', async () => {
```
MartinSchuhmacher, line 85: that change irritates me, if I am in edit mode, I dont want the component to move, right?

src/modules/data/board/ContentElementState.composable.ts
uidp: I agree. The code was there before. So I also wasn't sure about the intention of the original author.

src/components/lern-store/ContentEduSharingFooter.unit.ts
uidp: We can do
```typescript
import mediaFileMock from "@@/tests/test-utils/mediaFileMock.js";
```
and use that instead. I updated the code.

src/modules/feature/board-drawing-element/DrawingContentElement.unit.ts
uidp: Well that is only the text of the `describe` reflecting what the test does. So maybe the test was wrong?

MartinSchuhmacher (approved):

PR #3303, opened 2024-06-26
BC-7445 - make text inputs on the board smooth

hoeppner-dataport:
# Short Description
When entering texts into text-fields for e.g. Alt-Texts or descriptions in elements, sometimes some letters get lost. Reason for this is updates from the server coming in - that were triggered by the user (him|her)self.
Solution: While the user is editing a textfield no update from the server is seen as more up-to-date than the current state of the input-field the user has... and by that gets ignored.

BC-7445

## Checklist before merging


NFriedo (approved):
Code was mixed with BC-7588. so here are just 2 files left :)

PR #3314, opened 2024-07-03
BC-7331 - Detect Internet and Socket Connection 

muratmerdoglu-dp:
# Short Description
- Users should be informed if the connection (internet or websocket issue) is interrupted

- https://ticketsystem.dbildungscloud.de/browse/BC-7331

## Checklist before merging


src/modules/data/board/boardActions/connections.composable.unit.ts
```diff
@@ -0,0 +1,166 @@
...
+};
+
+jest.mock("vue-i18n", () => ({
+	useI18n: () => ({
+		t: mocks.i18n.t,
+	}),
+}));
+
+jest.mock("../Board.store", () => ({
+	useBoardStore: () => ({
```
NFriedo, line 33: Wouldn't it be better to use: 

- `setActivePinia(createTestingPinia());`
- `const boardStore = mockedPiniaStoreTyping(useBoardStore);`

for mocking our pinia board store? 🤔 
Then we could do boardStore.fetchBoardRequest and it would be type safe :)

src/modules/data/board/boardInactivity.composable.ts
```diff
@@ -0,0 +1,45 @@
+import { useDocumentVisibility } from "@vueuse/core";
+import { useTimeoutFn } from "@vueuse/shared";
+import { nextTick, ref, watch } from "vue";
+import { useBoardStore, useCardStore } from "@data-board";
+
+export const connectionOptions = {
+	isTimeoutReached: false,
+	MAX_TIMEOUT_FOR_INACTIVITY: 0,
```
NFriedo, line 8: If this is a constant shouldn't we take it out of this object?
Or, if this is a variable value shouldn't it be camelCase?

src/modules/data/board/boardInactivity.composable.unit.ts
NFriedo, line 1: When running the tests there is this reference error:
`ReferenceError: setImmediate is not defined`

We need to mock `socket.ts` in this test 🤔  (jest.mock("./socket/socket" [...], e.g. like in `boardSocketApi.composable.unit.ts`)

I also think we should put lines 53-57 (and then also the mocked socket thing) in a beforeEach(), so that each test starts with a clean setup 🧹

hoeppner-dataport (approved):

PR #3324, opened 2024-07-11
BC-7442 - bugfix birthdate

EzzatOmar:
# Short Description

Render error in firefox if date is less then unix time ~ 1970.

https://ticketsystem.dbildungscloud.de/browse/BC-7442

## Changes

Add a new Date wrap in the fromUTC function.

## Checklist before merging


src/plugins/datetime.js
```diff
@@ -100,7 +100,8 @@ const initDefaultTimezone = (app, store) => {
  * @return {dayjs} Date object based on current timezone
  */
 export const fromUTC = (date) => {
-	return dayjs.tz(date, "UTC");
+	// Date object is needed for firefox bug if date is before 1970
+	return dayjs.tz(new Date(date), "UTC");
```
MartinSchuhmacher, line 104:
```suggestion
        const dateObject = new Date(date);
	return dayjs.tz(dateObject, "UTC");
```

wolfganggreschus: Maybe something like this:
...
const utcDate = dayjs.tz(dateObject, "UTC");
return utcDate;

MartinSchuhmacher (approved):

PR #3331, opened 2024-07-18
N21-2076 adds empty state for media shelf

arnegns:
# Short Description
Adds empty state to media-shelf

- https://ticketsystem.dbildungscloud.de/browse/N21-2076

## Checklist before merging


src/modules/feature/media-shelf/data/mediaBoardState.composable.unit.ts
```diff
@@ -439,6 +441,14 @@ describe("mediaBoardState.composable", () => {
 				);
 			});
 
+			it("should handle board operation loading state", async () => {
```
MarvinOehlerkingCap, line 444: I don't understand what these tests are supposed to do. It says "should handle board operation loading state" but it only checks that `isLoading` is false after running the function.

src/modules/page/media-shelf/MediaShelf.page.vue
```diff
@@ -32,11 +32,34 @@
 				/>
 			</VBtnToggle>
 		</div>
-		<MediaBoard
-			v-if="mediaBoard && availableMediaLine"
-			:board="mediaBoard"
-			:available-media-line="availableMediaLine"
-		/>
+		<template v-if="isLoading && !mediaBoard && !availableMediaLine">
```
MarvinOehlerkingCap, line 35: `isLoading && (!mediaBoard || !availableMediaLine)`
If it is loading and either mediaBoard or availableMediaLine are undefined

src/modules/page/media-shelf/MediaShelf.page.vue
```diff
@@ -61,8 +85,25 @@ const {
...
+	isLoading,
+	isBoardOperationLoading,
 } = useSharedMediaBoardState();
 
+const isEmptyState: ComputedRef<boolean> = computed(() => {
+	const isMediaBoardEmpty = mediaBoard.value?.lines.every(
+		(line) => line.elements.length === 0
+	);
+
+	const isAvailableMediaLineEmpty = !availableMediaLine.value?.elements.length;
```
MarvinOehlerkingCap, line 97: isAvailableMediaLineEmpty returns true if availableMediaLine is undefined and isMediaBoardEmpty returns false if mediaBoard is undefined. That does not seem consistent.

src/modules/page/media-shelf/MediaShelf.page.unit.ts
```diff
@@ -99,6 +122,21 @@ describe("MediaShelfPage", () => {
 
 			expect(board.isVisible()).toEqual(true);
 		});
+
+		it("should display empty state when there are no media elements", async () => {
```
MarvinOehlerkingCap, line 126: this is a new "when"-scenario and should be in a describe with its own setup

src/modules/page/media-shelf/MediaShelf.page.unit.ts
```diff
@@ -65,6 +65,29 @@ describe("MediaShelfPage", () => {
 		jest.resetAllMocks();
 	});
 
+	describe("when the page is loading", () => {
+		const setup = async () => {
+			const { wrapper } = getWrapper();
+
+			useSharedMediaBoardStateMock.mediaBoard.value = undefined;
```
MarvinOehlerkingCap, line 72: you did not set the useSharedMediaBoardStateMock to isLoading=true, but it still displays the skeleton loader?

MarvinOehlerkingCap: I think you need to assign a ref to the createMock() in the beforeEach with
```
createMock<ReturnType<typeof useSharedMediaBoardState>>({
	isLoading: ref(false),
	isBoardOperationLoading: ref(false),
	mediaBoard: ref(mediaBoardResponseFactory.build()),
	availableMediaLine: ref(mediaAvailableLineResponseFactory.build());
})
```

MarvinOehlerkingCap: Also setting up your mocks (`useSharedMediaBoardStateMock.mediaBoard.value = undefined;`) before creating the wrapper might save you from using `await flushPromises();`

src/modules/feature/media-shelf/data/mediaBoardState.composable.ts
```diff
@@ -378,8 +388,9 @@ const useMediaBoardState = () => {
 			].elements.findIndex((element) => element.id === elementId);
 
 			mediaBoard.value.lines[lineIndex].elements.splice(elementIndex, 1);
-
+			isBoardOperationLoading.value = true;
```
MarvinOehlerkingCap, line 391: I think this is a philosophical question, but is only the api call to delete the element the BoardOperationLoading or is the whole try block the BoardOperation, that is Loading? Since the `splice` on the elements array is the actual remove operation on the data structure, that is displayed and used for the calculation of isEmptyState, then should it not be included in the isBoardOperationLoading? I think this would be a problem on a multithreaded system, but luckily JavaScript is not.

MarvinOehlerkingCap (changes requested):

MarvinOehlerkingCap (approved):

PR #3356, opened 2024-08-13
BC-7810 - Board: Mention element type in delete dialog

NFriedo:
# Short Description
As a user with edit rights for the board I want that in the delete dialog the element type is mentioned so that I can be sure which element I am deleting.

BC-7810

## Checklist before merging


src/modules/ui/board/BoardMenu.vue
```diff
@@ -58,8 +48,37 @@ const props = defineProps({
...
+	column: "components.board.menu.column",
+	card: "components.board.menu.card",
+	collaborativeTextEditorElement:
+		"components.board.menu.collaborativeTextEditorElement",
+	drawingElement: "components.board.menu.drawingElement",
+	externalToolElement: "components.board.menu.externalToolElement",
+	fileElement: "components.board.menu.fileElement",
+	linkElement: "components.board.menu.linkElement",
+	submissionElement: "components.board.menu.submissionElement",
+};
```
muratmerdoglu-dp, line 64: Because this constant is not used somewhere else, it can be moved in `boardMenuAriaLabel` scope (line 66).

src/modules/ui/board/BoardMenu.vue
```diff
@@ -58,8 +48,37 @@ const props = defineProps({
...
+
+const BOARD_SCOPES_WITH_BACKGROUND: Array<BoardMenuScope> = [
+	"card",
+	"collaborativeTextEditorElement",
+	"drawingElement",
+	"externalToolElement",
+	"fileElement",
+	"linkElement",
+	"submissionElement",
+];
```
muratmerdoglu-dp, line 78: Using capital letters looks as if a global variable for me, and it also can be moved the scope where it is used (line 80)

src/modules/ui/board/BoardMenu.vue
NFriedo: I would rather not move to the computed scope since that would be re-evaluated whenever the scope value changes, which means we would always rebuild this static record. 
I know that the performance impact is not huge but I would prefer this for readability. 
Do you have a strong opinion on this?

src/modules/ui/board/BoardMenu.vue
NFriedo: I will change the capital letters :) 
Regarding moving it to the scope: same question as above ⬆️

muratmerdoglu-dp (approved):

PR #3361, opened 2024-08-15
BC-7664 - rename room stores and room pages

wolfganggreschus:
# Short Description

The most necessary part of the old room namespace is renamed to give the new room concept the opportunity to be properly implied

BC-7664
## Checklist before merging


src/router/routes.ts
```diff
@@ -231,33 +231,47 @@ export const routes: Readonly<RouteRecordRaw[]> = [
...
 	},
-	{
-		path: `/rooms/:id(${REGEX_ID})/board`,
-		component: async () => (await import("@page-board")).ColumnBoardPage,
-		name: "rooms-board",
-		props: (route: RouteLocationNormalized) => ({
-			boardId: route.params.id,
-		}),
-	},
+	// TODO Is this redirect really necessary? How likely is it that there are bookmarks?
```
MartinSchuhmacher, line 239: imo we should get rid of that extra route

src/router/routes.ts
```diff
@@ -231,33 +231,47 @@ export const routes: Readonly<RouteRecordRaw[]> = [
 		beforeEnter: createPermissionGuard(["news_edit"]),
 	},
 	{
-		path: `/rooms/:id(${REGEX_ID})`,
-		component: () => import("@/pages/rooms/RoomDetails.page.vue"),
-		name: "rooms-id",
-	},
-	{
+		// TODO BC-7822, BC-7823 target this route at new rooms page
```
MartinSchuhmacher, line 234: I would add to the comment "... and remove the redirect with the new component used" 
Just to be safe.

src/router/routes.ts
uidp: Seems that there's still a reference in the legacy client:
https://github.com/hpi-schul-cloud/schulcloud-client/blob/90b9a5986160f04aabc510d2daa36bcfb5ceabc5/views/courses/edit-course.hbs#L50
That has to be updated then as well. Plus we have to remove the Ingress-entry from dof_app_deploy:
https://github.com/hpi-schul-cloud/dof_app_deploy/blob/8a3ad21db336c396ca6dec3c49821f09ce36b1cf/ansible/group_vars/all/x_ingress.yml#L90-L93

src/router/routes.ts
uidp: Updated comment

src/router/routes.ts
MartinSchuhmacher: then maybe we can add a comment here instead with something like "used in legacy and therefore necessary"?

uidp (dismissed):

src/router/routes.ts
uidp: Added the comment in line 234.

MartinSchuhmacher (approved):

src/router/routes.ts
uidp: I created a ticket (BC-7877) and updated the comment.

PR #3362, opened 2024-08-15
BC-7256 - Adaptive preview for file (only pdf with preview) and link element

NFriedo:
# Short Description
Add adaptive previews for link and pdf for the list board.
The preview is placed next to the content for >600px viewport. 

BC-7256

## Checklist before merging


src/composables/mimeType.composable.ts
```diff
@@ -0,0 +1,26 @@
...
+			mimeType === "application/ogg"
+		);
+	});
+
+	return {
+		isPdfMimeType,
+		isAudioMimeType,
+		isVideoMimeType,
+	};
+};
```
muratmerdoglu-dp, line 26: I think this file should not be a composable file because there is no need to make it stateful. The return values of `isPdfMimeType, isAudioMimeType, isVideoMimeType` should also be not reactive (computed, ref, etc..). So, I think the function should be moved to the utils file (fileHelper).

src/modules/feature/board-file-element/content/FileContent.vue
```diff
@@ -1,76 +1,138 @@
...
-	},
+const onUpdateText = useDebounceFn((value: string) => {
+	emit("update:alternativeText", value);
+}, 600);
+
+const onAddAlert = (alert: FileAlert) => {
+	emit("add:alert", alert);
+};
+
+const mimeType = computed(() => props.fileProperties.mimeType);
```
muratmerdoglu-dp, line 94: An alternative way to prevent losing reactivity of props is using `toRef` or `toRefs` instead of using `computed`.
E.g.: `const { mimeType } = toRefs(props).fileProperties.value;`

src/modules/feature/board-file-element/content/display/FileDisplay.vue
```diff
@@ -1,133 +1,66 @@
...
-	</div>
+	<PdfDisplay
+		v-if="isPdfMimeType && fileProperties.previewUrl"
+		:src="fileProperties.url"
+		:preview-src="fileProperties.previewUrl"
+		:name="fileProperties.name"
+		:is-edit-mode="isEditMode"
+		:element="fileProperties.element"
+	>
+		<slot />
```
muratmerdoglu-dp, line 10: Are these slots used somewhere?

src/modules/feature/board-file-element/content/display/FileDisplay.vue
```diff
@@ -1,133 +1,66 @@
...
-			);
-		});
+	isEditMode: { type: Boolean, required: true },
+});
+const emit = defineEmits(["video-error", "add:alert"]);
 
-		const onAddAlert = (alert: FileAlert) => {
-			emit("add:alert", alert);
-		};
+const mimeType = computed(() => props.fileProperties.mimeType);
```
muratmerdoglu-dp, line 58: Same as above.

src/modules/feature/board-file-element/content/inputs/FileInputs.vue
```diff
@@ -6,46 +6,34 @@
...
 
-		const hasPdfMimeType = computed(() =>
-			isPdfMimeType(props.fileProperties.mimeType)
-		);
+const onUpdateText = (value: string) => emit("update:alternativeText", value);
 
-		return { onUpdateText, onUpdateCaption, hasPdfMimeType };
-	},
-});
+const mimeType = computed(() => props.fileProperties.mimeType);
```
muratmerdoglu-dp, line 37: Same as above

muratmerdoglu-dp (changes requested):

src/composables/mimeType.composable.ts
NFriedo: moved back to file helper

src/modules/feature/board-file-element/content/FileContent.vue
NFriedo: Thank you for the hint! 
Removed because mimeType moved back to file helper

src/modules/feature/board-file-element/content/display/FileDisplay.vue
NFriedo: I investigated the slots here. This is used for the 3 dot menu on the preview images. 
 So implementation is quite confusing but we still need them.

muratmerdoglu-dp (dismissed):

muratmerdoglu-dp (approved):

PR #3368, opened 2024-08-20
N21-2138 migration wizard clear all auto-matches

GordonNicholasCap:
# Short Description

- https://ticketsystem.dbildungscloud.de/browse/N21-2138
- https://github.com/hpi-schul-cloud/schulcloud-server/pull/5193

## Checklist before merging


src/pages/administration/Migration.page.vue
```diff
@@ -74,28 +74,58 @@
...
+							{{
+								t(
+									"components.administration.adminMigrationSection.migrationWizardCancelDialog.Description"
+								)
+							}}
+						</template>
+					</VCustomDialog>
+				</div>
+
+				<div data-testid="clear-auto-matches-dialog-wrapper">
```
GordonNicholasCap, line 103: Each VCustomDialog is put into its own parent div. This is a workaround for testing teleporting component. In the tests, the wrapper can find both VCustomDialog's. However, the inner contents (html, text) and the html attributes of the found components are "teleported" away. Without the "data-testid" attribute, it is hard to find out if the component is for "cancel" or "clear all auto matches". Instead of the correct contents, the found components contain the contents and html attributes of their parent. Here is where these "parent divs" help with the problem. Since, the found components show the "data-testid" of the parent, we can use that to know which dialog component it is.

alweber-cap (approved):

src/pages/administration/Migration.page.vue
MarvinOehlerkingCap: I think it is better to use the ref-attribute, since it is only virtual and you only need this for unit tests, so rendering an actual tag just to make it testable feels wrong. 
```
Template:
<VCustomDialog ref="dialog123">

Unit-Test:
const dialog = wrapper.findComponent({ ref: "dialog123" });
```

PR #3371, opened 2024-08-21
BC-7812 - Implement timeout to show error notifications

muratmerdoglu-dp:
# Short Description

https://ticketsystem.dbildungscloud.de/browse/BC-7812

## Checklist before merging


src/modules/data/board/socket/socket.ts
```diff
@@ -26,24 +25,26 @@ export const useSocketConnection = (dispatch: (action: Action) => void) => {
...
-
-				if (!(boardStore.board && cardStore.cards)) return;
-
-				await boardStore.reloadBoard();
-				await cardStore.fetchCardRequest({
-					cardIds: Object.keys(cardStore.cards),
-				});
-				await nextTick();
+			if (isInitialConnection.value) return;
+			if (timeoutFn.isPending.value) {
```
hoeppner-dataport, line 29: Is it possible that the code gets to this line without timeoutFn being defined... is a questionmark or a check against undefined maybe needed here?

src/modules/data/board/socket/socket.unit.ts
```diff
@@ -107,20 +136,69 @@ describe("socket.ts", () => {
 			);
 		});
 
+		it("should not show connection restored notification when 'isInitialConnection' is true", () => {
```
hoeppner-dataport, line 139: The when part of this text should be a wrapping "describe" and should not use variable names:
```
describe("when the client connects for the first time", () => {
  it("should not show 'connection restored'-notification", () => {
```
same changes suggested for the following it(s).

src/modules/data/board/socket/socket.unit.ts
```diff
@@ -107,20 +136,69 @@ describe("socket.ts", () => {
...
+			getOrInitialiseBoardStore();
+			isInitialConnection.value = true;
+			jest.clearAllMocks();
```
hoeppner-dataport, line 142: jest.clearAllMocks should be placed in an afterAll()

src/modules/data/board/socket/socket.unit.ts
```diff
@@ -107,20 +136,69 @@ describe("socket.ts", () => {
...
+
+			eventCallbacks.connect();
+
+			expect(mockBoardNotifierCalls.showSuccess).not.toHaveBeenCalled();
+			expect(boardStore.reloadBoard).not.toHaveBeenCalled();
+			expect(cardStore.fetchCardRequest).not.toHaveBeenCalled();
+		});
+
+		it("should show connection restored notification when 'isInitialConnection' is false", () => {
+			getOrInitialiseBoardStore();
```
hoeppner-dataport, line 155: Maybe a setup-function can be useful here.

hoeppner-dataport (changes requested):
Only some small suggestions on what can be improved.

src/modules/data/board/socket/socket.unit.ts
```diff
@@ -59,9 +84,6 @@ describe("socket.ts", () => {
 
 		mockBoardNotifierCalls = createMock<ReturnType<typeof useBoardNotifier>>();
 		mockUseBoardNotifier.mockReturnValue(mockBoardNotifierCalls);
-	});
-
-	beforeEach(() => {
```
NFriedo, line 64: Is there a reason why you removed the beforeEach() here? 
I think it would be better to have it there, to keep the tests independent from each other 🤔

src/modules/data/board/socket/socket.unit.ts
```diff
@@ -26,15 +26,40 @@ const mockUseBoardNotifier = jest.mocked(useBoardNotifier);
...
+	});
+};
+
 describe("socket.ts", () => {
 	let dispatchMock: jest.Mock;
 	let mockSocket: Partial<socketModule.Socket>;
 	let timeoutResponseMock: { emitWithAck: jest.Mock };
 	let mockBoardNotifierCalls: DeepMocked<ReturnType<typeof useBoardNotifier>>;
-	// We need to set these two things in the outmost describe level since the socket event handlers that set and use these
-	// values are created only once when the module is loaded and initially used.
```
NFriedo, line 35: Why did you remove this comment?

src/modules/data/board/socket/socket.unit.ts
muratmerdoglu-dp: `mockSocket.connected = true;` can be moved into `beforeEach()`, it make sense. But assigning `dispatchMock = jest.fn();` in every iteration is not necessary.

src/modules/data/board/socket/socket.unit.ts
muratmerdoglu-dp: It's unclear "which two things", that's why these comments were removed.

NFriedo: I think it was about the following two lines. So maybe it would be better two write "We need to set the following two lines[...]" instead of removing it completely? Or do you think the info is not necessary?

NFriedo (approved):

PR #3374, opened 2024-08-23
BC-7904 - adap for FE the BE rooms endpoints renaming

virgilchiriac:
# Short Description

BC-7904
https://github.com/hpi-schul-cloud/schulcloud-server/pull/5203

## Checklist before merging


src/modules/data/board/BoardApi.composable.ts
```diff
@@ -28,7 +28,7 @@ export const useBoardApi = () => {
 	const cardsApi = BoardCardApiFactory(undefined, "/v3", $axios);
 	const elementApi = BoardElementApiFactory(undefined, "/v3", $axios);
 
-	const roomApi = RoomsApiFactory(undefined, "/v3", $axios);
+	const roomApi = CourseRoomsApiFactory(undefined, "/v3", $axios);
```
MartinSchuhmacher, line 31: same here as in the server PR the distinction between roomApi and courseRoomApi might be important

src/modules/data/board/BoardApi.composable.ts
```diff
@@ -196,7 +196,9 @@ export const useBoardApi = () => {
 			return undefined;
 		}
 		const context = contextResponse.data;
-		const roomResponse = await roomApi.roomsControllerGetRoomBoard(context.id);
+		const roomResponse = await roomApi.courseRoomsControllerGetRoomBoard(
```
MartinSchuhmacher, line 199: same here as above

src/modules/data/board/BoardApi.composable.ts
virgilchiriac: this is a very good point, because unlike the old room this code is not deprecated.

In the case of the board, basically we need some info from the (old) room (name and id - I assume is used for breadcrumbs). 

Once the new room will implemented, the board can belong to new room or to the old room. 
Depends on future implementation, this context would to be provided by the backend, and maybe will be ok to keep calling it room (old or new, FE does not have to care at this point).

MartinSchuhmacher (approved):

PR #3387, opened 2024-09-02
BC-7964 - Implement room conditional loading for room details page

uidp:
# Short Description

We want to be able to load both - rooms and course rooms - on the same page (room details). This requires to implement a mechanism which does the following:

- try to load a room by id
- check if it exists
- if yes: render room details
- otherwise try to load course with the same id
- if found: render course room details
- otherwise handle not found

BC-7964

## Checklist before merging


src/modules/feature/room/RoomGrid.vue
```diff
@@ -73,3 +73,4 @@ onMounted(() => {
 	}
 }
 </style>
+../../data/room/Rooms.state../../data/room/Rooms.state
```
odalys-dataport, line 76: Some auto import or rename must have gone wrong here

src/modules/data/room/RoomDetails.state.ts
```diff
@@ -0,0 +1,25 @@
...
+	const isLoading = ref(true);
+
+	const fetchRoom = async (id: string) => {
+		await delay(100);
+		// TODO call API
+		room.value = roomsData.find((r) => r.id === id);
+		isLoading.value = false;
+	};
+
+	const isRoom = computed(() => room.value != null);
```
odalys-dataport, line 17: Why are we not using `!==`?

src/modules/page/room/RoomDetails.page.vue
```diff
@@ -0,0 +1,32 @@
...
+
+<script setup lang="ts">
+import CourseRoomDetailsPage from "@/pages/course-rooms/CourseRoomDetails.page.vue";
+import { RoomDetails } from "@feature-room";
+import { useRoomDetailsState } from "@data-room";
+import { ref, watch } from "vue";
+import { useRoute } from "vue-router";
+
+const route = useRoute();
+const isRoom = ref(true);
```
odalys-dataport, line 21: Can't we use the computed property from `RoomDetails.state` here?

odalys-dataport (changes requested):

src/modules/feature/room/RoomGrid.vue
uidp: Updated

src/modules/data/room/RoomDetails.state.ts
uidp: This is because `value != null` includes `undefined` as well and is a general way to check for existence. Agreed, we could also have used `value !== undefined` because of the exact type of the property but I think the above is ok as an existence check.

src/modules/page/room/RoomDetails.page.vue
uidp: Well, the computed doesn't work here. But I moved the `isRoom` ref to the `RoomDetails.state` and removed the (unused) computed property.

odalys-dataport (dismissed):

src/modules/page/room/RoomDetails.page.vue
```diff
@@ -12,18 +12,26 @@
...
+const envConfigModule = injectStrict(ENV_CONFIG_MODULE_KEY);
+
 const route = useRoute();
 const { fetchRoom, isLoading, isRoom, room } = useRoomDetailsState();
 
 watch(
 	() => route.params.id,
 	async () => {
-		await fetchRoom(route.params.id as string);
+		if (envConfigModule.getEnv["FEATURE_ROOMS_ENABLED"]) {
```
odalys-dataport, line 29: Shouldn't that rather be a route guard?

MartinSchuhmacher: if the same route is used, I would asume it makes more sense here

src/modules/feature/room/RoomDetails.vue
```diff
@@ -0,0 +1,42 @@
...
+const { t } = useI18n();
+
+const props = defineProps({
+	room: { type: Object as PropType<Room | undefined> },
+});
+
+const room = computed(() => props.room);
+
+const breadcrumbs: ComputedRef<Breadcrumb[]> = computed(() => {
+	if (room.value != null) {
```
MartinSchuhmacher, line 28: `!==` ?

src/modules/page/room/RoomDetails.page.vue
odalys-dataport: Ah okay, I thought they can also redirect and such

src/modules/feature/room/RoomDetails.vue
uidp: We already had that comment by @odalys-dataport in a previous commit. When we check `value != null` then it includes the check for `undefined` because `(undefined == null) === true`. So IMHO this is a nice way to check for existence. Agreed we could use `value !== undefined`. That would be even more exact. But I like `!= null` as a general existence check.

odalys-dataport: Uwe already explained this in a prior comment

odalys-dataport (approved):

PR #3390, opened 2024-09-02
BC-7985 - Adapt sidebar active selection to rooms changes

uidp:
# Short Description
In order to distinguish between the room details and course room details in the sidebar selection we have to add some specific logic.

BC-7985

## Checklist before merging


src/modules/ui/layout/sidebar/SidebarItem.vue
```diff
@@ -34,19 +38,39 @@ const props = defineProps({
...
-	const isRoomsPageActive = route.path === "/rooms";
-	if (roomsSidebarItem) {
-		return isRoomsPageActive;
+	// RoomDetails, CourseRoomDetails
+	if (route.name === "rooms-id") {
+		if (isRoom.value === true) {
+			return props.item.to === "/rooms";
+		} else {
+			return (
+				props.item.to === "/rooms/courses-overview" && isLoading.value === false
```
MartinSchuhmacher, line 55: why seting `isLoading.value === false` when in the room details store in `fetchRoom` it is anyway set?

MartinSchuhmacher (commented):
als some tests seem to be failing in the CI

src/modules/page/room/RoomDetails.page.vue
```diff
@@ -13,26 +13,34 @@
...
 		} else {
-			isLoading.value = false;
-			isRoom.value = false;
+			deactivateRoom();
 		}
 	},
 	{ immediate: true }
 );
+
+const isRoom = computed(() => roomVariant.value === RoomVariant.ROOM);
```
MartinSchuhmacher, line 41: not sure how I feel about roomVariant and RoomVariant working together here. Can we find a more distinctive naming?

MartinSchuhmacher (approved):

src/modules/page/room/RoomDetails.page.vue
uidp: Sure, we can talk about that. I just wanted to avoid having two boolean flags.

src/pages/Home.page.vue
```diff
@@ -0,0 +1 @@
+<template>&nbsp;</template>
```
odalys-dataport, line 1: What's this about? Can this be removed again?

uidp: I added the root route to prevent Vue warnings about missing route for path `""`. This page is actually never rendered. The `&nbsp;` is there because we cannot have empty `<template>` tags.

MartinSchuhmacher: lets talk tomorrow about the whole idea for the routing so we can all have a similar understanding with the new approach :)

PR #3395, opened 2024-09-04
BC-8017 - Enable Cache Busting via Content Hash in Frontend

uidp:
# Short Description
In order to make sure that frontend application code is reloaded in the user's browser when it has been changed we want to enable Cache Busting.
Webpack builds can be configured to write a content hash in the generated file names during build. When the conten has been updated the hash will change and the browser is forced to reload the file.

https://webpack.js.org/guides/caching/

BC-8017

## Checklist before merging

### DIFF
--- a/config/webpack/webpack.common.js
+++ b/config/webpack/webpack.common.js
@@ -26,7 +26,7 @@ module.exports = {
 
 	// Output
 	output: {
-		filename: "_nuxt/js/[name].js",
+		filename: "_nuxt/js/[name]-[contenthash].js",
 		path: path.resolve(__base, "dist"),
 		publicPath: "/",
 		chunkFilename: "_nuxt/js/[name].js",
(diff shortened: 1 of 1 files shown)


config/webpack/webpack.common.js
```diff
@@ -25,13 +25,16 @@ module.exports = {
...
-		publicPath: "/",
-		chunkFilename: "_nuxt/js/[name].js",
-		clean: true,
-	},
+	output:
+		process.env.NODE_ENV === "development"
+			? {}
+			: {
+					filename: "_nuxt/js/[name]-[contenthash:6].js",
+					chunkFilename: "_nuxt/js/[name]-[contenthash:6].js",
```
MartinSchuhmacher, line 33: Why specifically length 6 instead of default?

Loki-Afro (approved):

config/webpack/webpack.common.js
Loki-Afro: suggest another number, not security relevant ;)

PR #3411, opened 2024-09-27
BC-8147 - Upgrade to ESLint 9

NFriedo:
# Short Description
In Eslint 9, there is a new default config format system, the flat config. Its also possible to use the new format in Eslint v8 (current: eslintrc format).
We want try to use the flat config and see if there are any problems.

This PR will upgrade ESLint 8 to ESLint 9.

BC-8147

## Changes

- change eslintrc format to flat config
- add `configType: "flat"` option to `ESLintWebpackPlugin` in `webpack.common.js` 
    - ⚠️ Attention: as long as we use eslint 8, we need to add `eslintPath: "eslint/use-at-your-own-risk"`, this is experimental
    - eslint 9 support is already there for the eslint webpack plugin
- just `@vue/eslint-config-typescript` is not supporting the new format yet: 
    - so we need to use the compatibility utilities `FlatCompat` from `@eslint/eslintrc`
    - they already working on v14, where they add support for flat config
- we don't need to add our custom eslint plugin to the package.json anymore
- remove .eslint.ignore (no longer supported), moved to `ignores` config object in `eslint.config.js`

Commits 7.10.24:
- upgrade `@vue/eslint-config-typescript` to version 14
- upgrade to `eslint 9.12.0`
- set `"@typescript-eslint/no-unused-vars": "warn"`, because there are many eslint errors in catch block
    - need to decide if want to use default option `caughtErrors: all` for 'catch' blocks
- disable `"no-unused-vars` for  @typescript-eslint/no-unused-vars 
- set `"@typescript-eslint/no-empty-object-type": ["error", { allowInterfaces: "with-single-extends" },],`	
- set `"@typescript-eslint/no-require-imports": "off"` for now
- remove `eslintPath: "eslint/use-at-your-own-risk"` because this was just needed vor eslint 8
- fix eslint errors

## Checklist before merging


eslint.config.js
```diff
@@ -0,0 +1,142 @@
...
+		ignores: [
+			".vscode/*",
+			"node_modules/*",
+			"**/dist/*",
+			"src/serverApi/*",
+			"src/fileStorageApi/*",
+			"src/h5pEditorApi/*",
+		],
+		rules: {
+			"schulcloud/material-icon-imports": "error",
```
hoeppner-dataport, line 45: Suggestion: let's sort all the configuration keys on all levels alphabetically.

NFriedo: Good idea, i will do that 👍

eslint.config.js
```diff
@@ -0,0 +1,142 @@
...
+			"node_modules/**",
+			"**/dist/**",
+			"src/serverApi/**",
+			"src/fileStorageApi/**",
+			"src/h5pEditorApi/**",
+		],
+	},
+	{
+		languageOptions: {
+			ecmaVersion: 2020,
```
uidp, line 38: Can we even set `latest` here (the default)?
https://eslint.org/docs/latest/use/configure/language-options
Or do we have restrictions regarding the "old" js-code?

eslint.config.js
```diff
@@ -0,0 +1,142 @@
...
+		},
+	},
+	{
+		files: ["**/*.unit.{j,t}s?(x)", "tests/**"],
+		languageOptions: {
+			globals: {
+				...globals.jest,
+				mount: false,
+				shallowMount: false,
+				fail: false,
```
uidp, line 138: What does `[key]: false` mean? Is it to disable the respective globals? And shouldn't that then be `"off"` instead of `true`? https://eslint.org/docs/latest/use/configure/language-options#using-configuration-files

uidp (dismissed):

eslint.config.js
NFriedo: Yes I think this would be even better! I will adjust it, I already tried it out. 
I just copied this from our old eslint config before 😅

eslint.config.js
NFriedo: Good catch! Its deprecated to use `false` now, seems like false was replaced with `"readonly"`.

When using "off" , the warning  `'mount' is not defined  no-undef` appears. 
So I think the right choice is "`readonly"`

uidp (approved):

uidp (approved):

PR #3419, opened 2024-10-08
N21-2167 preferred tools for boards

IgorCapCoder:
# Short Description
- add preferred tool icons to addElement Dialog
- create external tool element directly after clicking icon in addElement when no custom paramter is needed
- open configuration dialog after clicking icon in addElement when at least 1 custom parameter is needed

- https://ticketsystem.dbildungscloud.de/browse/N21-2167
- https://github.com/hpi-schul-cloud/dof_app_deploy/pull/985

## Checklist before merging


src/modules/data/board/cardActions/cardRestApi.composable.unit.ts
```diff
@@ -1,36 +1,57 @@
...
 import { useCardRestApi } from "./cardRestApi.composable";
-import { Router, useRouter } from "vue-router";
 
 jest.mock("@/components/error-handling/ErrorHandler.composable");
 const mockedUseErrorHandler = jest.mocked(useErrorHandler);
 
 jest.mock("../BoardApi.composable");
 const mockedUseBoardApi = jest.mocked(useBoardApi);
 
+jest.mock("@data-external-tool/ContextExternalToolApi.composable");
```
GordonNicholasCap, line 52: The "unit test" github action could not resolve this path, locally it works fine, maybe due to the path for "@data-external-tool" alias not being correctly set. "@data-external-tool" can be replaced by with "@/modules/..." or maybe you can add the alias to `tsconfig.json`, push and see if that fixes this alias problem.

src/modules/data/board/Card.store.ts
```diff
@@ -220,7 +235,16 @@ export const useCardStore = defineStore("cardStore", () => {
...
+	const loadPreferredTools = async (contextType: ToolContextType) => {
+		preferredTools.value = await restApi.getPreferredTools(contextType);
+	};
+
+	const getPreferredTools = (): PreferredToolResponse[] | undefined => {
+		return preferredTools.value;
+	};
+
 	return {
+		createPreferredElement,
```
GordonNicholasCap, line 247: Not a bug here, but in `/board/boardInactivity.composable.unit.ts` there is a type error for the card store. I do not really understand why... the changes you did should not cause this error. This could be a bug from the "mockPiniaStoreTyping" function. The only fix I can see here is to set the type to "any" and suppress any warnings.

GordonNicholasCap (changes requested):
The changes are many, but it looks fine. There are some errors that need fixing.

src/modules/ui/extended-icon-btn/ExtendedIconBtn.unit.ts
```diff
@@ -38,6 +39,9 @@ describe("@ui-extended-icon-btn/ExtendedIconBtn.vue", () => {
 		expect(wrapper.html()).toContain("example-label");
 	});
 
+	// TODO N21-2167 find tooltip
```
GordonNicholasCap, line 42: Reminder for commented code.

I did abit of research on this: Since the tooltip is a teleporting component, this is tricky. I cannot find the tooltip via the "wrapper.findComponent" way (usually this works also for teleporting comp.), but I can find it in the DOM itself via "document.body.innerHTML". There you can see the html body of the tooltip.

But, the html of the tooltip is always there, even if the "mouseover" event is NOT triggered. I guess this is a feature of tooltip in V-Btn and it is something out of our control. So, this makes it hard to test a tooltip for a V-Btn. You can either not test the tooltip (it cannot be properly tested) or just test that the html of the tooltip is there.

src/modules/ui/extended-icon-btn/ExtendedIconBtn.unit.ts
```diff
@@ -47,4 +51,21 @@ describe("@ui-extended-icon-btn/ExtendedIconBtn.vue", () => {
 
 		expect(wrapper.emitted("click")).toBeDefined();
 	});
+
+	// TODO N21-2167 find tooltip
```
GordonNicholasCap, line 55: Reminder for commented code

src/modules/feature/board/shared/AddElementDialog.composable.unit.ts
```diff
@@ -472,11 +499,42 @@ describe("ElementTypeSelection Composable", () => {
...
 
-				const action = elementTypeOptions.value[4].action;
+				const action = elementTypeOptions.value[5].action;
 				action();
 
 				expect(closeDialogMock).toBeCalledTimes(1);
 			});
 		});
+
+		// TODO N21-2167 fix Cannot read properties of undefined (reading 'showCustomNotifier')
```
GordonNicholasCap, line 509: Reminder for commented code

GordonNicholasCap (dismissed):

PR #3420, opened 2024-10-09
BC-8211 - Cancel Confirmation on Room Edit/Create

odalys-dataport:
# Short Description

Ask user to confirm that they want to cancel editing or creating a room. Do not open dialog, when the form doesn't show any changes.

BC-8211

## Checklist before merging


src/modules/feature/room/RoomForm.vue
```diff
@@ -76,14 +84,22 @@ const props = defineProps({
 const emit = defineEmits(["save", "cancel"]);
 const { t } = useI18n();
 
+const { askConfirmation } = useConfirmationDialog();
+
 const roomData = computed(() => props.room);
 
+const onUpdateColor = () => {
+	v$.value.$touch();
```
wolfganggreschus, line 92: Even it's possible a variable should be declared before using it. First time I saw it I thought it's a global variable.

wolfganggreschus, line 92: Even though this didn't happen in this PR, naming a variable "v$" is very cryptic and hard to read for people unfamiliar with the code. Suggestion: 
`const validationState = useVuelidate(rules, { roomData }, { $lazy: true, $autoDirty: true });`

src/modules/feature/room/RoomForm.vue
```diff
@@ -106,7 +122,16 @@ const onSave = async () => {
 	emit("save", roomData.value);
 };
 
-const onCancel = () => {
-	emit("cancel");
+const onCancel = async () => {
+	if (!v$.value.$anyDirty) emit("cancel");
```
wolfganggreschus, line 126: It's a little bit difficult to understand. Suggestion:
`const noChangesMade = !v$.value.$anyDirty;
if (!v$.value.$anyDirty) emit("cancel");`

src/modules/feature/room/RoomColorPicker/RoomColorPickerSwatch.unit.ts
```diff
@@ -5,7 +5,7 @@ import {
 import { mount } from "@vue/test-utils";
 import { ComponentProps } from "vue-component-type-helpers";
 import RoomColorPickerSwatch from "./RoomColorPickerSwatch.vue";
-import { RoomColorEnum } from "./types";
+import { RoomColor } from "@/serverApi/v3";
```
uidp, line 8: The type aliases were actually intended to be able to decouple from the api types. Why why are they removed?

src/locales/de.ts
```diff
@@ -156,10 +156,10 @@ export default {
 	"common.words.classes": "Klassen",
 	"common.words.color": "Farbe",
 	"common.words.color.blue": "Blau",
-	"common.words.color.blueGrey": "Blaugrau",
+	"common.words.color.blue-grey": "Blaugrau",
```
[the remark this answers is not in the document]
odalys-dataport: Yes, I did this to be able to reuse the color names from the enum. There are some other occurrences where we use kebab case instead as well, but I'm not sure if I am happy with this solution either.

src/modules/feature/room/RoomColorPicker/RoomColorPickerSwatch.unit.ts
odalys-dataport: As it's not possible to assign enums to a different variable name, we'll keep the import from api here.

src/modules/feature/room/RoomForm.vue
odalys-dataport: done :)

odalys-dataport: It is a little confusing at first glance, but we decided to stick with `v$` as it's the vuelidate convention.

src/modules/feature/room/RoomForm.unit.ts
```diff
@@ -0,0 +1,136 @@
...
+};
+
+describe("@feature-room/RoomForm", () => {
+	const setup = (props: ComponentProps<typeof RoomForm>) => {
+		const wrapper = mount(RoomForm, {
+			global: {
+				plugins: [createTestingVuetify(), createTestingI18n()],
+			},
+			props,
+			attachTo: document.body,
```
uidp, line 31: Do we need this?

src/modules/feature/room/RoomForm.unit.ts
```diff
@@ -0,0 +1,136 @@
...
+	it("should not save invalid room", async () => {
+		const { wrapper } = setup({ room: invalidMockRoom });
+
+		// validation needs to be triggered manually due to code structure
+		await wrapper.vm.v$.$validate();
+
+		await wrapper.find("[type='submit']").trigger("click");
+		await nextTick();
+
+		expect(wrapper.vm.v$.$invalid).toEqual(true);
```
uidp, line 46: Can we find a better way to watch the behavior of the component other than looking at the internal state? (other than `wrapper.vm`)

uidp (approved):

src/modules/feature/room/RoomForm.unit.ts
MartinSchuhmacher: yes, otherwise the state was not observable in the test

PR #3436, opened 2024-11-04
BC-8187 - Create and use Boards in Rooms

uidp:
# Short Description

As a teacher and student I want to be able to use a new single or multicolumn board within a room.

BC-8187

## Checklist before merging


src/modules/page/room/RoomDetails.page.vue
```diff
@@ -128,6 +159,73 @@ const breadcrumbs: ComputedRef<Breadcrumb[]> = computed(() => {
...
+
+const boardLayoutDialogIsOpen = ref(false);
+
+const createBoard = async (layout: BoardLayout) => {
+	if (!room.value) return;
+
+	const boardApi = BoardApiFactory(undefined, "/v3", $axios);
+
+	const params: CreateBoardBodyParams = {
+		title: "Raum-Board", // t("pages.room.boardCard.label.courseBoard"),
```
MartinSchuhmacher, line 174: ?

src/modules/data/room/RoomDetails.store.ts
```diff
@@ -14,13 +14,17 @@ export const useRoomDetailsStore = defineStore("roomDetailsStore", () => {
...
 
 	const fetchRoom = async (id: string) => {
 		try {
 			room.value = (await roomApi.roomControllerGetRoomDetails(id)).data;
 			roomVariant.value = RoomVariant.ROOM;
+			roomBoards.value = (
+				await roomApi.roomControllerGetRoomBoards(id)
+			).data.data;
 		} catch (error) {
 			const responseError = mapAxiosErrorToResponseError(error);
```
odalys-dataport, line 29: We should handle an error here properly by displaying an alert and staying on the current page.

src/modules/feature/room/BoardGrid.vue
```diff
@@ -0,0 +1,25 @@
+<template>
+	<v-row class="mt-4" data-testid="board-grid">
```
odalys-dataport, line 2: I don't think we need a data-testid for the grid, but maybe we ask Ankush

src/modules/feature/room/BoardTile.vue
```diff
@@ -0,0 +1,89 @@
...
+		v-if="board"
+		class="board-tile pa-4"
+		:class="{ 'board-hidden': isHidden }"
+		hover
+		role="link"
+		:data-testid="`board-tile-${props.index}`"
+		@click="openBoard"
+		@keydown.enter.self="openBoard"
+	>
+		<div
```
odalys-dataport, line 12: Why are we not using `VCardTitle` and `VCardSubtitle` here?

src/modules/feature/room/BoardTile.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+	<VCard
+		v-if="board"
```
odalys-dataport, line 3: I feel like we should move this if to the parent component. Inside the BoardTile component the board data should be present otherwise there is no need to render the BoardTile component

src/modules/feature/room/BoardTile.vue
```diff
@@ -0,0 +1,89 @@
...
+		class="board-tile pa-4"
+		:class="{ 'board-hidden': isHidden }"
+		hover
+		role="link"
```
odalys-dataport, line 7: Knowing there is gonna be a board, we can then use the `to` prop, which will declare the card as a link automatically, which should either add the role link automatically as well or make it obsolete. semantic html > setting roles manually

src/modules/feature/room/BoardTile.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+	<VCard
+		v-if="board"
+		class="board-tile pa-4"
+		:class="{ 'board-hidden': isHidden }"
```
odalys-dataport, line 5: The naming of the class and the variable is a little confusing in my opinion as the card is not actually hidden it just gets a different look.

src/modules/feature/room/BoardTile.vue
```diff
@@ -0,0 +1,89 @@
...
+const isListBoard = computed(() => {
+	return board.value?.layout === BoardLayout.List;
+});
+
+const isHidden = computed(() => {
+	return board.value?.isVisible === false;
+});
+
+const subtitleIcon = computed(() => {
+	const icon = isListBoard.value ? "$gridOutline" : mdiViewDashboardOutline;
```
odalys-dataport, line 55: Why are we using the `"$gridOutline"` here instead of importing the icon?

src/modules/feature/room/RoomDetails.vue
```diff
@@ -1,20 +1,23 @@
 <template>
 	<template v-if="room">
 		<div>
```
odalys-dataport, line 3: Is this div still necessary?

src/modules/page/room/RoomDetails.page.vue
```diff
@@ -2,7 +2,12 @@
 	<div v-if="isLoading" />
```
odalys-dataport, line 2: Maybe it's time we show a proper loading state?

src/modules/page/room/RoomDetails.page.vue
```diff
@@ -128,6 +159,73 @@ const breadcrumbs: ComputedRef<Breadcrumb[]> = computed(() => {
...
+		layout,
+	};
+	const boardId = (await boardApi.boardControllerCreateBoard(params)).data.id;
+
+	router.push(`/boards/${boardId}`);
+};
+
+const fabItems = computed(() => {
+	const actions = [];
+	// TODO refine permissions
```
odalys-dataport, line 186: Can this be done now or is there a ticket for this?

src/modules/page/room/RoomDetails.page.vue
```diff
@@ -128,6 +159,73 @@ const breadcrumbs: ComputedRef<Breadcrumb[]> = computed(() => {
...
+		parentId: room.value.id,
+		parentType: BoardParentType.Room,
+		layout,
+	};
+	const boardId = (await boardApi.boardControllerCreateBoard(params)).data.id;
+
+	router.push(`/boards/${boardId}`);
+};
+
+const fabItems = computed(() => {
```
odalys-dataport, line 184: I feel like this code (defining fabItems) is incredibly hard to read

src/modules/page/room/RoomDetails.page.vue
```diff
@@ -128,6 +159,73 @@ const breadcrumbs: ComputedRef<Breadcrumb[]> = computed(() => {
...
+		actions: actions,
+	};
+
+	return items;
+});
+
+const fabItemClickHandler = (event: string) => {
+	if (event === "board-type-dialog-open") {
+		boardLayoutDialogIsOpen.value = true;
+	} else if (event === "board-create") {
```
odalys-dataport, line 224: Maybe that's just me, but I prefer two ifs over else if

odalys-dataport (changes requested):

src/modules/data/room/RoomDetails.store.ts
```diff
@@ -14,13 +14,17 @@ export const useRoomDetailsStore = defineStore("roomDetailsStore", () => {
...
 
 	const roomApi = RoomApiFactory(undefined, "/v3", $axios);
 
 	const fetchRoom = async (id: string) => {
 		try {
 			room.value = (await roomApi.roomControllerGetRoomDetails(id)).data;
 			roomVariant.value = RoomVariant.ROOM;
+			roomBoards.value = (
+				await roomApi.roomControllerGetRoomBoards(id)
+			).data.data;
```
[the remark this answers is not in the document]
uidp: Wir haben das bereits an verschiedenen Stellen. Und zwar immer bei Pagination-Responses. Die haben ein `data` für die axios Response Daten und ein `data` für die Elemente. Man könnte evtl. darüber nachdenken, die Pagination Response so zu ändern, dass der Property Name für die "Items" anders als `data` heißt.

src/modules/page/room/RoomDetails.page.vue
uidp: Fixed that.

src/modules/data/board/BoardPageInformation.composable.ts
```diff
@@ -11,43 +11,62 @@ const useBoardPageInformation = () => {
...
 		const courseNameForPageTitle = courseName ? ", " + courseName : "";
+
 		return buildPageTitle(
 			`${t("pages.room.boardCard.label.courseBoard")}${courseNameForPageTitle}`
 		);
-	};
+	});
+
+	const roomId = computed(() => contextInfo.value?.id);
+	const contextType = computed(() => contextInfo.value?.type);
```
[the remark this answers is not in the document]
uidp: Ok, I changed `contextInfo` to `boardContext`. Since the composable is used in a couple of places changing `contextType` would require some refactoring or at least lead to some confusion if we would only rename it locally.  So I would rather leave it that way (for now). Actually `roomId` isn't precise eather because there are potentially other board contexts than room (aka course).

src/modules/data/room/RoomDetails.store.ts
uidp: But that would be very unlikely because the room should be already found by the previous API request. So I would see that as an unexpected error.

src/modules/feature/room/BoardTile.vue
uidp: We can actually. Was just a first prototype. So I didn't use the components. I updated the code.

src/modules/feature/room/BoardTile.vue
uidp: I think it can even be left out entirely because it should always be defined.

src/modules/feature/room/BoardTile.vue
uidp: Right. Works even with the enter key.

src/modules/feature/room/BoardTile.vue
uidp: Semantically, I think it should be something like `isDraft`. But then it's quite confusing when the API property is called `isVisible`. Generally I would prefer the `isPublished` / `isDraft` approach. But we decided in favor of `isVisible` 🤨

src/modules/data/room/RoomDetails.store.ts
odalys-dataport: Unless the error is impossible we should still handle it though no?

src/modules/feature/room/BoardTile.vue
uidp: The icon `mdiViewGridOutline` is different and looks more like `mdiViewDashboardOutline`. That has to be fixed in the final version.

src/modules/feature/room/BoardTile.vue
odalys-dataport: Yea, it's a little confusing, because it's visible for the teacher but not for the student in the future. Within this component I'd still prefer to move away from `isHidden` and `board-hidden`.

src/modules/feature/room/RoomDetails.vue
uidp: Moved the check to `RoomDetailsPage` and add `required: true` to the property declaration.

src/modules/page/room/RoomDetails.page.vue
uidp: We're still at the prototype stage. I suggest to add that later. Maybe we come up with something better than the old static skeleton?

src/modules/page/room/RoomDetails.page.vue
uidp: Well, that's a way to aggregate the fab items based on the permissions of the user. Maybe we can think about refactoring that somehow.

src/modules/page/room/RoomDetails.page.vue
odalys-dataport: Let's create a ticket for that then, so it's not forgotten.

src/modules/feature/room/BoardTile.vue
odalys-dataport: Please remove `role="link"` as well :)

src/modules/feature/room/RoomDetails.vue
```diff
@@ -1,20 +1,20 @@
 <template>
-	<template v-if="room">
-		<div>
-			<!-- content goes here-->
-		</div>
-	</template>
+	<div>
```
odalys-dataport, line 2: Is this surrounding div necessary?

odalys-dataport (changes requested):

odalys-dataport (approved):

PR #3446, opened 2024-11-11
N21-2149 merlin bibliothek element

MBergCap:
# Short Description

- https://ticketsystem.dbildungscloud.de/browse/N21-2149
- https://github.com/hpi-schul-cloud/schulcloud-server/pull/5323
- https://github.com/hpi-schul-cloud/superhero-dashboard/pull/268

## Checklist before merging


src/modules/feature/board-external-tool-element/ExternalToolElement.vue
```diff
@@ -138,6 +135,14 @@ const hasLinkedTool: ComputedRef<boolean> = computed(
 	() => !!modelValue.value.contextExternalToolId
 );
 
+const isDeepLinkingTool: ComputedRef<boolean | undefined> = computed(
+	() => toolLaunchRequest.value?.isDeepLink
+);
+
+const showTool: ComputedRef<boolean> = computed(() =>
+	isDeepLinkingTool.value ? false : hasLinkedTool.value
```
GordonNicholasCap, line 143: I found out that when the merlin tool should not be shown (i.e. not in edit mode or user not allowed to edit), the merlin tool appears briefly after the page loads, which seems weird. I assume it is most probably because the value of `isDeepLinkingTool` remains `undefined` shortly until the `toolLauchRequest` finishes its request.

Can you add an extra condition for when it is undefined, return false for `showTool` and check if this problem is still there?

GordonNicholasCap (changes requested):

GordonNicholasCap (approved):

PR #3450, opened 2024-11-13
BC-8342 - Change naming from participant(s) to member(s)

muratmerdoglu-dp:
# Short Description

- https://ticketsystem.dbildungscloud.de/browse/BC-8342

## Checklist before merging


src/modules/feature/room/RoomMembers/AddMembers.vue
NFriedo, line 1: Not really related to the ticket, but we can remove the `color="primary"` from the autocompletes because we added it in the src/plugins/vuetify.ts to use the primary color. 
I also think we don't need "bg-color" there, could you check it again? :)

NFriedo (approved):

PR #3500, opened 2025-01-16
BC-8731 - Fix outline and contrast issue for boards as draft in rooms

f-g-s:
- fix color contrast issue for boards in draft status to 80% opacity
- darken outline of draft for better visibility
- delete outline in cards (board)
- adapt outline for elements

### DIFF
--- a/src/modules/feature/room/BoardTile.vue
+++ b/src/modules/feature/room/BoardTile.vue
@@ -1,6 +1,6 @@
 <template>
 	<VCard
-		:class="isDraft ? 'opacity-70' : 'bg-surface-light'"
+		:class="isDraft ? 'opacity-80' : 'bg-surface-light'"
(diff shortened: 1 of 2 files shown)


src/modules/feature/room/BoardGrid.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<v-row class="mt-4" data-testid="board-grid">
+	<v-row class="mt-4 board-grid" data-testid="board-grid">
```
virgilchiriac, line 2: class is not used anywhere, but ok

virgilchiriac (approved):

src/modules/feature/room/BoardGrid.vue
f-g-s: is used in _cards.scss for 25% opacity outline (only for room and course) - opacity stays at 12% for all other places eg in boards

PR #3504, opened 2025-01-17
BC-7631 - Keyboard does not open when tapping on column or board title 

NFriedo:
# Short Description

BC-7631

## Changes
Change `readonly` to `disabled` in `BoardAnyTitleInput.vue` because it seems like readonly does not trigger the keyboard as the input is not editable (readonly: true - can be focused but not edited).
This is just a quick fix. The Board inputs should be reworked with a11y in mind anyways (inputs should not have role heading)

## Checklist before merging


hoeppner-dataport (approved):
👍

PR #3516, opened 2025-01-22
BC-8578 - Remove Visibility Alert From Board

odalys-dataport:
# Short Description

Removes alert from board about its visibility and adds a tooltip to the "draft-chip".

BC-8578

## Screenshots of UI changes

Screenshot 2025-01-22 at 14 21 19

## Checklist before merging


src/modules/feature/board/board/BoardHeader.vue
```diff
@@ -26,14 +26,23 @@
...
+							v-bind="props"
+							size="small"
+							class="align-self-center cursor-default"
+							data-testid="board-draft-chip"
+						>
+							{{ t("common.words.draft") }}
+						</VChip>
+					</template>
+				</VTooltip>
+			</template>
```
uidp, line 45: Maybe it's a good idea to put that in a seperate component? Event if it's only used ones.

src/modules/feature/board/board/Board.vue
```diff
@@ -328,25 +326,9 @@ onUnmounted(() => {
 	resetNotifierModule();
 });
 
-const setAlert = useDebounceFn(() => {
```
MartinSchuhmacher, line 331: Remove import of useDebounceFn above

src/modules/feature/board/board/BoardHeader.vue
MartinSchuhmacher: Would also suggest that, imo could easily be adjusted to find usage somewhere else, but maybe topic for a follow up refactoring?!

MartinSchuhmacher (changes requested):

src/components/molecules/Alert.vue
```diff
@@ -7,7 +7,6 @@
 		closable
 		max-width="400"
 		min-width="200"
-		:close-label="$t('common.labels.close')"
```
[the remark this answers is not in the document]
odalys-dataport: Because we can now use vuetify translations for it. I checked and it's the exact same translations.

src/modules/feature/board/board/BoardDraftChip.unit.ts
```diff
@@ -0,0 +1,22 @@
+import {
+	createTestingI18n,
+	createTestingVuetify,
+} from "@@/tests/test-utils/setup";
+import BoardDraftChip from "./BoardDraftChip.vue";
+
+describe("@feature-board/BoardDraftChip", () => {
```
MartinSchuhmacher, line 7: what do you think about two additional tests for text in `VChip` and `VTooltip`?

MartinSchuhmacher (changes requested):
+ this one still needs to be removed:
https://github.com/hpi-schul-cloud/nuxt-client/pull/3516#discussion_r1925889090

src/modules/feature/board/board/BoardDraftChip.unit.ts
odalys-dataport: Imo this falls under the category "testing the framework" as there is no logic or changing texts in this component.

MartinSchuhmacher (approved):

src/modules/feature/board/board/BoardDraftChip.unit.ts
uidp: Hm, maybe it would make sense to check if the correct text is rendered inside the component. Something like:
```typescript
it("should render the text message", () => {
	const wrapper = setup();

	expect(wrapper.text()).toContain("common.words.draft");
});
```

odalys-dataport: Can do. Not sure how easy it is to test for the tooltip text though

uidp: Right. I confused the two components, sorry.

uidp (approved):

PR #3583, opened 2025-03-04
BC-8259 Use inline version of ckeditor on boards

dyedwiper:
# Short Description

## Checklist before merging


src/modules/feature/board-text-element/RichTextContentElementEdit.vue
```diff
@@ -1,26 +1,29 @@
...
+		<InlineEditor
 			v-model="modelValue"
 			:autofocus="autofocus"
 			:placeholder="$t('components.cardElement.richTextElement.placeholder')"
-			type="balloon"
-			mode="simple"
+			type="inline"
+			mode="regular"
+			:viewport-offset-top="ckeditorViewportOffsetTop"
 			@update:value="onUpdateValue"
```
uidp, line 10: As a side note: I don't understand how the value update is working here: Why do we need `v-model` + local ref (`modelValue`) and `onUpdateValue`? Shouldn't be a `computed` get / set enought? like:
```typescript
			const value = computed({
				get() {
					return props.value;
				},

				set value(newValue: string) {
					emit("update:value", newValue);
				},
			});
	```

File: src/modules/feature/board-text-element/RichTextContentElementEdit.vue [hunk 2]
```diff
@@ -30,11 +33,62 @@ export default defineComponent({
...
+				const marginBottom =
+					columnHeaderStyle.getPropertyValue("margin-bottom");
+
+				offset =
+					staticOffset + height + parseInt(marginTop) + parseInt(marginBottom);
+			}
+
+			return offset;
+		});
+
```
uidp, line 91: Can we maybe extract the whole computation of CKEditor top viewport offset to a composable. The `BOARD_IS_LIST_LAYOUT` injection could then also be moved there.

src/modules/feature/board/board/BoardColumnHeader.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-	<div class="d-flex flex-wrap mb-4 mt-2" ref="columnHeader">
+	<div
+		id="boardColumnHeader"
```
uidp, line 3: Shouldn't that be a class instead of an id? Looks like the `BoardColumnHeader` is not unique - see RichTextContentElementEdit.vue:69

src/modules/feature/editor/EditorConfig.composable.ts
```diff
@@ -0,0 +1,277 @@
+import { Editor } from "@ckeditor/ckeditor5-core";
```
uidp, line 1: Looks like the package was removed from package.json. Looks like this leads to "Cannot find module '@ckeditor/ckeditor5-core'..."

src/modules/feature/editor/EditorConfig.composable.ts
```diff
@@ -0,0 +1,277 @@
...
+			(evt: CKEditorEventInfo, data: CKEditorKeystrokeInfo) => {
+				if (
+					data.domEvent.key === "Backspace" ||
+					data.domEvent.key === "Delete"
+				) {
+					callback();
+				}
+			}
+		);
+	};
```
uidp, line 263: Shouldn't that be implemented as an event instead of a callback? Or is there a particular reaon for the callback here?

src/modules/feature/board/board/BoardColumnHeader.vue
dyedwiper: You're right. This should be a class and probably written in kebab-case also.

src/modules/feature/board/board/BoardColumn.vue
```diff
@@ -129,6 +129,43 @@ export default defineComponent({
...
+			const marginTop = columnHeaderStyle.getPropertyValue("margin-top");
+			const marginBottom = columnHeaderStyle.getPropertyValue("margin-bottom");
+
+			const offset =
+				staticOffset + height + parseInt(marginTop) + parseInt(marginBottom);
+
+			return offset;
+		});
+
+		provide("ckeditorViewportOffsetTop", ckeditorViewportOffsetTop);
```
uidp, line 167: Shouldn't that be provided by a symbol to be type-safe?

src/modules/feature/board-text-element/RichTextContentElementEdit.vue
```diff
@@ -41,53 +40,10 @@ export default defineComponent({
...
-				const marginBottom =
-					columnHeaderStyle.getPropertyValue("margin-bottom");
-
-				offset =
-					staticOffset + height + parseInt(marginTop) + parseInt(marginBottom);
-			}
-
-			return offset;
-		});
+		const ckeditorViewportOffsetTop = inject(
```
uidp, line 43: Shouldn't that be `injectStrict` to be more type-safe?

src/modules/feature/board/board/BoardColumn.vue
```diff
@@ -129,6 +129,43 @@ export default defineComponent({
 	setup(props, { emit }) {
 		const boardStore = useBoardStore();
 		const reactiveIndex = toRef(props, "index");
+
+		const ckeditorViewportOffsetTop = computed(() => {
```
uidp, line 133: Can we maybe extract that into its own composable?

src/modules/feature/board-text-element/RichTextContentElementEdit.vue
MartinSchuhmacher: done

src/modules/feature/board/board/BoardColumn.vue
MartinSchuhmacher: done

src/modules/data/editor/ViewportOffsetCalculation.composable.ts
```diff
@@ -0,0 +1,33 @@
+export function ckeditorViewportOffsetTop(): number {
```
uidp, line 1: Is this really a composable? I think it would be better if the composable function returned an object containing a ref or computed property with value `viewportOffsetTop` and would be named someting like `useviewportOffsetTop` or `useviewportOffsetTopCalculation`, e.g.
```typescript
import { computed } from "vue";

function calculateViewportOffsetTop(): number {
    // ...
 }
 
 export const useviewportOffsetTop = () => {
	const offset = computed(() => calculateViewportOffsetTop());

	return {
		offset,
	};
};
```

src/modules/feature/board-text-element/RichTextContentElement.vue
```diff
@@ -10,6 +10,7 @@
 			class="rich_text"
 			:autofocus="autofocus"
 			:value="modelValue.text"
+			:columnIndex="columnIndex"
```
odalys-dataport, line 13: This is not being used in `RichTextContentElementEdit`, can it be removed?

src/modules/feature/board/board/BoardColumnHeader.vue
odalys-dataport: Has this been taken care of. If not, @MartinSchuhmacher do you wanna take this?

src/modules/feature/board-text-element/RichTextContentElement.vue
odalys-dataport: Used for indexed data-testids now

MartinSchuhmacher (approved):

PR #3587, opened 2025-03-05
BC-8012 - Fix no-console-log linter errors

wolfganggreschus:
# Short Description
A very simple logger is to be built to intercept the existing console.logs.

BC-8012
## Checklist before merging


src/modules/util/logger/logger.ts
uidp, line 1: I really don't like how this is implemented as a class with class methods. Can we just implement that as an object of a specific type `Logger`? e.g.
```typescript
export type Logger = {
	info(message: unknown, ...optionalParams: unknown[]): void;
	warn(message: unknown, ...optionalParams: unknown[]): void;
	error(message: unknown, ...optionalParams: unknown[]): void;
	log(message: unknown, ...optionalParams: unknown[]): void;
}

const logger: Logger = {
	// ...
};

export { logger };
```

wolfganggreschus: "I really don't like..."?

Loki-Afro:
everything that you do with process.env.* has no effect on some env other than your local one, this env var is not as everything gets packed into js beforehand and simply delivered to the browser, thus there are no env vars

wolfganggreschus:
> everything that you do with process.env.* has no effect on some env other than your local one, this env var is not as everything gets packed into js beforehand and simply delivered to the browser, thus there are no env vars

Of course, what you say makes sense. Thank you.

uidp (approved):

PR #3588, opened 2025-03-06
BC-8013 - remove vue3-mq dependency

odalys-dataport:
# Short Description

The vue3-mq package manages media queries. It causes a typescript error as the package is written in pure javascript and it is not needed anymore as we can handle media queries with Vuetfiy. Additionally removing this package and replacing it with Vuetify functionality will also make the media breakpoints in those few components where mq is used more consistent with the rest of our application.

BC-8013
https://github.com/hpi-schul-cloud/schulcloud-server/pull/5562

## Checklist before merging


src/components/molecules/AlertContainer.unit.ts
```diff
@@ -41,52 +36,67 @@ describe("AlertContainer", () => {
 		expect(alertComponent.exists()).toBe(false);
 	});
 
-	it("should hand over notification to Alert", () => {
-		const data: AlertPayload = {
+	it("should display Alert with correct data", () => {
+		const alertPayload: AlertPayload = {
```
wolfganggreschus, line 40: You could avoid the explicit naming of the type if you made the AlertStatus type an enum. Then the assignment of the variable would look like this:
`	it("should display Alert with correct data", () => {
		const alertPayload = {
			text: "hello world",
			status: AlertStatusEnum.Success,
			autoClose: true,
			timeout: 5000,
		};

And the wrapper would accept the variable.
But that might be too much refactoring.

src/pages/course-rooms/CourseRoomList.page.unit.ts
```diff
@@ -83,6 +80,12 @@ const mockData = [
 ];
 
 describe("@/pages/CourseRoomListPage", () => {
+	const setup = () => {
+		const wrapper = getWrapper();
+
+		return { wrapper };
+	};
+
 	beforeEach(() => {
```
wolfganggreschus, line 89: Just a question, and it wasn't part of the ticket, but: is it against a convention if you move the block in beforeEach into the setup function, since you call the setup function every time?
It would work and I find it simpler.

src/components/molecules/AlertContainer.unit.ts
uidp: Hm, but then you wouldn't have any type safety on the assignment of the literal object. Specifying the type ensures that we have an object of the exact type on the right side of the assignment (which would be otherwise very loosely typed). This prevents further ambiguities when using the object (e.g. with optional properties).

src/components/molecules/AlertContainer.unit.ts
```diff
@@ -41,52 +36,67 @@ describe("AlertContainer", () => {
...
-		const { wrapper } = getWrapper("desktop", [dataOne, dataTwo]);
+		const { wrapper } = getWrapper(data);
 
 		const notificationData = wrapper.findAllComponents(Alert);
 		expect(notificationData.length).toEqual(2);
 	});
 
 	it("should set mobile position-class as default", () => {
-		const { wrapper } = getWrapper("mobile");
+		Object.defineProperty(window, "innerWidth", {
```
uidp, line 76: Looks like be that we need this more often. Maybe we should generalize this somewhere in the future e.g. in the test setup or in a test helper.

src/pages/course-rooms/CourseRoomOverview.page.unit.js
```diff
@@ -217,7 +206,7 @@ describe("@/pages/CourseRoomOverview.page", () => {
 	});
 
 	it("should display 2 avatars component in 'mobile' device", async () => {
-		const wrapper = getWrapper("mobile");
+		const wrapper = getWrapper();
```
uidp, line 209: Looks like we now we haven't any means of setting up the device for the test anymore right? Or did I miss something?

MartinSchuhmacher (approved):

PR #3590, opened 2025-03-06
BC-7477 Fix double scroll bar on board

dyedwiper:
# Short Description

https://ticketsystem.dbildungscloud.de/browse/BC-7477

## Checklist before merging


src/modules/feature/board/board/BoardColumn.vue
```diff
@@ -353,51 +340,69 @@ export default defineComponent({
...
+		border-radius: 5px;
+	}
+
+	.column-drag-handle:hover .scrollable::-webkit-scrollbar-thumb {
+		background-color: rgba(var(--v-theme-on-surface), 0.6);
+		border-radius: 5px;
+	}
+
+	.scrollable::-webkit-scrollbar-thumb:hover {
+		background: rgba(var(--v-theme-on-surface), 0.8) !important;
```
MartinSchuhmacher, line 394: can we adjust specificity here to avoid using `!important` ?

src/modules/feature/board/board/BoardColumn.vue
```diff
@@ -353,51 +340,69 @@ export default defineComponent({
...
+	See: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_box_model/Mastering_margin_collapsing and https://stackoverflow.com/a/19719427/11854580 */
+	overflow: hidden;
 }
-.column-drag-handle:hover .scrollable-column::-webkit-scrollbar-thumb {
-	background-color: rgba(var(--v-theme-on-surface), 0.6);
-	border-radius: 5px;
+
+@supports selector(::-webkit-scrollbar) {
+	.scrollable::-webkit-scrollbar {
+		width: 6px;
```
MartinSchuhmacher, line 375: any option to use vars here? (also for remaining places here)

src/modules/feature/board/board/BoardColumn.vue
```diff
@@ -353,51 +340,69 @@ export default defineComponent({
...
+	transition: box-shadow 150ms all;
 }
 
-/* Track */
-.scrollable-column::-webkit-scrollbar-track {
-	background: white;
-	border: none;
+.multi-column-board-column {
+	/* Subtracted are the heights of schulcloud-header, board-header and scrollbar. */
+	height: calc(100vh - 64px - 92.75px - 10px);
```
MartinSchuhmacher, line 361: I would prefer that we do not use hand written values here because we would run into issues, once the values for the mentioned components change

src/modules/feature/board/board/BoardColumn.vue
```diff
@@ -353,51 +340,69 @@ export default defineComponent({
...
+	width: 400px;
 }
 
-/* Handle */
-.scrollable-column::-webkit-scrollbar-thumb {
-	background-color: transparent;
-	border-radius: 5px;
+.expanded-sortable {
+	/* Subtracted is the height of the add-button. */
+	min-height: calc(100% - 64px);
```
MartinSchuhmacher, line 367: same here as above

src/modules/feature/board/board/BoardColumnGhost.vue
```diff
@@ -83,25 +83,17 @@ const title = computed(() =>
...
- * Without this rule cards have to be placed closely below the last card in a column to be added.
-*/
-.column-container {
-	min-height: 70vh;
-	height: 100%;
-	padding-bottom: 50px;
+/* The height is set here to extend the droppable area of the ghost-column. */
+/* Subtracted is the height of the ghost-column-header. */
+.sortable-container {
+	height: calc(100% - 77px);
```
MartinSchuhmacher, line 89: same here as in `.multi-column-board-column`

src/modules/feature/board/board/BoardColumn.vue
dyedwiper: done that

src/styles/css-variables/_sizes.scss
```diff
@@ -12,4 +12,5 @@
 	--sidebar-sub-item-height: 40px;
 	--topbar-height: 50px;
 	--content-max-width-vw: 100vw;
+	--appbar-height: 64px;
```
odalys-dataport, line 15: Can we do the same for the other values again (92.75px, 77px and 10px)?

I also think we should name this `topbar-height` as our component is called topbar and we have no control over vuetify name changes. 

It would also be great to then put these in a new file as all other variables in this file are old and will be cleaned up within EPIC-70. Maybe even a more telling name than just `sizes`.

odalys-dataport (changes requested):

src/styles/css-variables/_sizes.scss
dyedwiper: `topbar-height` is already in use, see line 13. I could rename this to `legacy-topbar-height`. Do you have a suggestion for a filename?

src/modules/feature/board/board/BoardColumn.vue
```diff
@@ -291,13 +282,10 @@ export default defineComponent({
 			return props.column.cards[index];
 		};
 
-		const sortableClasses = computed(() => {
+		const scrollableClasses = computed(() => {
 			const classes = [];
 			if (!props.isListBoard) {
-				classes.push("scrollable-column");
-				if (isDragging.value) {
```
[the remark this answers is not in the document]
dyedwiper: These `scrollableClasses` are now for the `div` surrounding `Sortable`and `BoardAddCardButton`. The dropping area must be set on the `Sortable` though. I put the logic in the template, see line 42.

src/modules/feature/board/board/Board.vue
```diff
@@ -454,3 +453,39 @@ const onSelectBoardLayout = async (layout: BoardLayout) => {
...
+<style scoped>
+.board {
+	padding: 0 var(--space-lg);
+}
+
+.column-board {
+	overflow-x: auto;
+}
+
+@supports selector(::-webkit-scrollbar) {
```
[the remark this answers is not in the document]
dyedwiper: General: https://developer.mozilla.org/en-US/docs/Web/CSS/@supports
Specific for scrollbar: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar#adding_a_fallback_for_scrollbar_styles

src/components/share/ShareModalResult.vue
```diff
@@ -123,7 +123,7 @@ const onShowQrCode = () => {
 }
 
 .button-max-width {
-	max-width: calc(var(--topbar-height) * 2);
+	max-width: calc(var(--legacy-topbar-height) * 2);
```
odalys-dataport, line 126: The class `button-max-width` is not being used in this component or its tests, it can be removed. This is faulty anyways as this modal is only accessible when you're logged in afaik and the `legacy-topbar-height` is only used with the `loggedOut` Layout.

src/modules/feature/board/board/Board.vue
```diff
@@ -454,3 +452,45 @@ const onSelectBoardLayout = async (layout: BoardLayout) => {
 	});
 };
 </script>
+
+<style>
+:root {
+	--board-header-height: 66px;
```
odalys-dataport, line 458: I think we should define this in a scss file as well instead of here

src/styles/css-variables/_sizes.scss
```diff
@@ -10,6 +10,8 @@
 	--sidebar-width-tablet: 60px;
 	--sidebar-item-height: 60px;
 	--sidebar-sub-item-height: 40px;
-	--topbar-height: 50px;
+	--legacy-topbar-height: 50px;
 	--content-max-width-vw: 100vw;
+	--topbar-height: 64px;
+	--breadcrumbs-height: 22px;
```
odalys-dataport, line 16: I think I mentioned this before, please add these three new variables (topbar, board-header and breadcrumbs height) within a new file that has a more specific name than just `sizes`. Maybe something like `_layout-sizes` idk

odalys-dataport (changes requested):

src/styles/css-variables/_sizes.scss
dyedwiper: Yes and I asked for your suggestion for a filename above. `_layout-sizes` is fine for me.

src/components/share/ShareModalResult.vue
dyedwiper: I didn't check for that. The other classes in this style block are unused too. So I removed the whole block.

src/modules/feature/board/board/BoardColumn.vue
dyedwiper: I introduced CSS variables for everything but the scrollbar height. To conveniently compute the scrollbar height, we have to first make all scrollbars consistent. So I'll leave it hardcoded for now.

src/styles/css-variables/_sizes.scss
odalys-dataport: Oh I missed that then, sorry! Thank you! :)

odalys-dataport (approved):

PR #3602, opened 2025-03-13
BC-8009 - eslint deep imports

hoeppner-dataport:
# Short Description
Fix ESlint rules:
- @typescript-eslint/no-restricted-imports
- incorrect undef lint-warnings for Clipboard and NodeJS
- fix deep-imports

BC-8009

## Checklist before merging


eslint.config.js
```diff
@@ -44,6 +44,8 @@ module.exports = defineConfigWithVueTs([
 			ecmaVersion: "latest",
 			globals: {
 				...globals.node,
+				NodeJS: true,
+				Clipboard: true,
```
odalys-dataport, line 48: Oohh nice! Can we do this for `Jest` as well? We're importing jest in some places even though it's not necessary (e.g.` __mocks__/maska.ts`)

eslint.config.js
```diff
@@ -151,4 +153,10 @@ module.exports = defineConfigWithVueTs([
 			"schulcloud/material-icon-imports": "off",
 		},
 	},
+	{
+		files: ["**/*.{ts,vue}"],
+		rules: {
+			"@typescript-eslint/no-require-imports": "error",
```
odalys-dataport, line 159: Could set the rule to "error" as default and turn it off as an exception?

eslint.config.js
NFriedo: We are doing this already but just for test files and files under test directory:

```ts
files: ["**/*.unit.{j,t}s?(x)", "tests/**"],
languageOptions: {
    globals: {
	...globals.jest,
```

Maybe this example file you mentioned should be moved to the test directory?

odalys-dataport: Like I said, there are more places where this is done. I looked them up by searching for `import { Jest`

Other examples:
- AddElementDialogMock.ts
- sharedElementTypeSelectionMock.ts
- ...

These are placed in our feature folder like `/modules/feature/<feature-name>/test-utils`. So I guess either we globally import Jest as well or move all these files to our test folder. We do not have to do this within this PR of course. We can take it into our FE meetings

NFriedo: Interesting, I always thought that we put everything test related in the `tests` directory. 
Maybe it's a good idea to talk about this in the FE meeting, as you suggested.

I actually liked it to have the jest globals just for the test files, but maybe thats just personal preference :)

NFriedo (approved):

eslint.config.js
hoeppner-dataport: I added the **/test-utils/** and the __mocks__ as patterns for test-related files, so that the profit on the "...globals.jest" that Nicole mentioned => avoiding to import it also for all other files.

Removed the fake-imports `import { Jest as jest } from "@jest/environment";` from several files.

PR #3614, opened 2025-03-21
BC-5670 - Prevent Switching View Mode of Card

muratmerdoglu-dp:
# Short Description

- https://ticketsystem.dbildungscloud.de/browse/BC-5670

## Checklist before merging


NFriedo:
When I delete a link element that is already created (entered a link), the card switches to view mode and does not stay in edit mode.
When I delete it when I didn't enter something (is being created), Card stays in edit mode.

src/modules/feature/board/shared/InlineEditInteractionHandler.vue
```diff
@@ -50,7 +58,12 @@ export default defineComponent({
 			)
 				return true;
 
-			return !isListItem(event.target) && !isDatePicker(event.target);
+			return (
+				event.target &&
```
NFriedo, line 62: Maybe we can improve readability here and assign `event.target` to a variable.

And maybe we could also use an extra var to store the "disallowedConditions", so it would be like:

```ts
const isAllowedTarget = (event: MouseEvent): boolean => {
    const target = event.target;
    if (!(target instanceof HTMLElement) && !(target instanceof SVGElement))
	return true;

    const disallowedConditions = [isListItem, isDatePicker, isLinkElement];
    return disallowedConditions.every((fn) => !fn(target));
};
```

src/modules/data/board/Card.store.unit.ts
```diff
@@ -729,6 +729,7 @@ describe("CardStore", () => {
 						cardStore.cards[cardId].elements[3].id
 					);
 					expect(setEditModeId).toHaveBeenCalled();
+					expect(setEditModeId).toHaveBeenCalledWith(cardId);
```
hoeppner-dataport, line 732: The line above can be removed.

src/modules/feature/board-file-element/content/footer/download/FileDownload.vue
```diff
@@ -13,29 +13,22 @@
...
-			mdiTrayArrowDown,
-		};
-	},
+const props = defineProps({
+	fileName: { type: String, required: true },
+	isDownloadAllowed: { type: Boolean, required: true },
+	url: { type: String, required: true },
 });
+
+const testId = ref("board-file-element-edit-menu-download");
```
hoeppner-dataport, line 27: Why did you introduce this ref? The value is never changed - not even in tests.

PR #3619, opened 2025-03-24
BC-9297 Scoring long board titles

f-g-s:
# Short Description

Long board title out of viewport, leads to horizontal scroll bar. 
Added max width and ellipsis for board-title.

https://ticketsystem.dbildungscloud.de/browse/BC-9297

## Screenshots of UI changes

image
image

## Checklist before merging


src/modules/feature/board/board/BoardHeader.vue
```diff
@@ -214,4 +216,12 @@ watchEffect(() => {
 	font-family: var(--font-accent);
 	letter-spacing: $field-letter-spacing;
 }
+
+.board-header--item {
```
dyedwiper, line 220:
```suggestion
.input-container {
```
I don't think you need to include "board-header" in the name because the styles are scoped to this component.

src/modules/feature/board/board/BoardHeader.vue
```diff
@@ -214,4 +216,12 @@ watchEffect(() => {
...
+	overflow: hidden;
+}
+
+.board-header--input {
```
dyedwiper, line 224:
```suggestion
.input {
```

src/modules/feature/board/board/BoardHeader.vue
```diff
@@ -214,4 +216,12 @@ watchEffect(() => {
...
+	max-width: calc(100% + 1.5rem);
```
dyedwiper, line 225: Why `+ 1.5em`?

f-g-s: As many characters of the board name as possible should be visible so that users know which board they are on. Scoring results in fewer characters being visible, while still leaving some space available until the chip or menu. This is used by overriding the otherwise valid padding by +1.5rem.

We can delete this if it is to hacky.

src/modules/feature/board/board/BoardHeader.vue
```diff
@@ -1,30 +1,32 @@
...
+					<BoardAnyTitleInput
+						class="ml-n4"
+						ref="boardHeader"
+						scope="board"
+						:value="boardTitle"
+						data-testid="board-title"
+						:isEditMode="isEditMode"
+						:isFocused="isFocusedById"
+						:maxLength="100"
+						:style="{ width: `${fieldWidth + 1}px` }"
```
dyedwiper, line 22: I just tried around with this `+ 1` and I don't see any difference. What is it for again?

f-g-s: Without the +1, Chrome displays the title with dots, even if there's enough space available on the right. Dots should only be displayed if the title goes to the right end of the viewport.

f-g-s: image
Without the +1

dyedwiper: Okay, I see. I moved the calculation into the script and added a comment.

src/modules/feature/board/board/BoardHeader.vue
```diff
@@ -2,28 +2,26 @@
...
-					:is-edit-mode="isEditMode"
-					:is-focused="isFocusedById"
-					:max-length="100"
-					:style="{ width: `${fieldWidth}px` }"
-					@update:value="updateBoardTitle"
-					@blur="onBoardTitleBlur"
-				/>
-				<span ref="inputWidthCalcSpan" class="input-width-calc-span" />
-			</div>
+			<BoardAnyTitleInput
```
uidp, line 12: I suppose it is safe to remove the outer `<div ref="boardHeader">` and set the `ref` on the `<BoardAnyTitleInput>` itself right?

uidp (approved):

PR #3639, opened 2025-04-09
BC-9107 - refactor the action menu of the room-members-page

hoeppner-dataport:
# Short Description
For reusing the actionMenu (used for batch actions) on other room-membership related functionalities, it needs to be refactored.

BC-9107

## Checklist before merging

### DIFF
--- a/src/components/templates/DefaultWireframe.vue
+++ b/src/components/templates/DefaultWireframe.vue
@@ -10,7 +10,7 @@
 			aria-live="assertive"
 			class="d-sr-only"
 		/>
-		<div class="wireframe-header sticky" :class="{ fixed: fixedHeader }">
+		<div class="wireframe-header sticky">
 			<Breadcrumbs v-if="breadcrumbs.length" :breadcrumbs="breadcrumbs" />
 			<div v-else class="breadcrumbs-placeholder" />
 			<slot name="header">
@@ -102,9 +102,6 @@ const props = defineProps({
 	dataTestid: {
 		type: String as PropType<string | null>,
 		default: null,
-	},
-	fixedHeader: {
-		type: Boolean,
 	},
 	mainWithoutPadding: {
 		type: Boolean,
@@ -212,6 +209,7 @@ $fab-wrapper-height: 80px;
 	align-items: center;
 	justify-content: flex-end;
 	height: $fab-wrapper-height;
+	margin-top: -#{$fab-wrapper-height}; // stylelint-disable-line sh-waqar/declaration-use-variable
 	margin-top: -#{$fab-wrapper-height};
 	pointer-events: none;
 
--- a/src/components/templates/DefaultWireframe.vue
+++ b/src/components/templates/DefaultWireframe.vue
@@ -209,7 +209,7 @@ $fab-wrapper-height: 80px;
 	align-items: center;
 	justify-content: flex-end;
 	height: $fab-wrapper-height;
-	margin-top: -#{$fab-wrapper-height}; // stylelint-disable-line sh-waqar/declaration-use-variable
+	margin-top: -#{$fab-wrapper-height};
 	pointer-events: none;
 
 	* {
--- a/src/components/templates/DefaultWireframe.vue
+++ b/src/components/templates/DefaultWireframe.vue
@@ -210,7 +210,6 @@ $fab-wrapper-height: 80px;
 	justify-content: flex-end;
 	height: $fab-wrapper-height;
 	margin-top: -#{$fab-wrapper-height}; // stylelint-disable-line sh-waqar/declaration-use-variable
-	margin-top: -#{$fab-wrapper-height};
 	pointer-events: none;
 
 	* {
--- a/src/modules/feature/room/RoomMembers/MembersTable.vue
+++ b/src/modules/feature/room/RoomMembers/MembersTable.vue
@@ -1,11 +1,8 @@
 <template>
 	<div
 		class="d-flex justify-space-between align-center ga-2 mb-2 table-title-header"
-		:class="{
-			'fixed-position': fixedPosition.enabled,
-			'flex-column': isExtraSmallDisplay,
-		}"
-		:style="{ top: `${fixedPosition.positionTop}px` }"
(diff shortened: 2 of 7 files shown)


src/modules/feature/room/RoomMembers/MembersTable.vue
```diff
@@ -1,22 +1,26 @@
...
 			class="multi-action-menu"
 			:class="{ 'order-2': isExtraSmallDisplay }"
 			:selected-ids="selectedUserIds"
-			:is-visible-change-role-button="isVisibleChangeRoleButton"
-			@remove:selected="onRemoveMembers"
 			@reset:selected="onResetSelectedMembers"
-			@change:role="onChangePermission"
-		/>
+		>
+			<KebabMenuList>
```
NFriedo, line 14: Maybe it makes sense to move the <KebabMenuList> to the ActionMenu.vue, because this will make sure that the first menu item is focused when using the component. 
Otherwise it's possible that we would forget to use <KebabMenuList> and just use the KebabMenuActions 💭 
But that is just an idea :D

src/components/templates/DefaultWireframe.vue
```diff
@@ -212,7 +209,7 @@ $fab-wrapper-height: 80px;
 	align-items: center;
 	justify-content: flex-end;
 	height: $fab-wrapper-height;
-	margin-top: -#{$fab-wrapper-height};
+	margin-top: -#{$fab-wrapper-height}; // stylelint-disable-line sh-waqar/declaration-use-variable
```
MartinSchuhmacher, line 212: can we avoid that comment somehow?

MartinSchuhmacher (commented):
https://github.com/hpi-schul-cloud/nuxt-client/pull/3639/files#diff-4a235e59277475fe275fd761c2ea1f47e93eadeef8f89ebadae4594ffaaa8bcdL11

is not used anymore as it seems, can be removed as well

src/modules/feature/room/RoomMembers/MembersTable.vue
```diff
@@ -1,22 +1,26 @@
 <template>
 	<div
 		class="d-flex justify-space-between align-center ga-2 mb-2 table-title-header"
-		:class="{
-			'fixed-position': fixedPosition.enabled,
-			'flex-column': isExtraSmallDisplay,
-		}"
-		:style="{ top: `${fixedPosition.positionTop}px` }"
+		:class="{ sticky: isMobileDevice, 'flex-column': isExtraSmallDisplay }"
+		:style="{ top: `${headerBottom + 1}px` }"
```
MartinSchuhmacher, line 5: as the `headerBottom` is only used here when we hand it into the component, I think its easier to follow up in the future, if we already give the needed value here (with the `+1`) to the component in `RoomMembers.page.vue` or compute it inside of this component. It does not change the behaviour but improves readability imo.

src/modules/page/room/RoomMembers.page.vue
```diff
@@ -146,12 +144,8 @@ const memberList: Ref<RoomMember[]> = ref(roomMembers);
...
 	buildPageTitle(`${room.value?.name} - ${t("pages.rooms.members.manage")}`)
 );
-const wireframe = ref<HTMLElement | null>(null);
-const fixedHeaderOnMobile = ref({
-	enabled: false,
-	positionTop: 0,
-});
-const { y } = useElementBounding(wireframe);
+const header = ref<HTMLElement | null>(null);
+const { bottom: headerBottom } = useElementBounding(header);
```
MartinSchuhmacher, line 148: do we need to be specific here what the value will be in case `header` is `null` ? I was not able to see how Vue handles that case codewise.

src/modules/feature/room/RoomMembers/MembersTable.vue
hoeppner-dataport: ok. converted into computed style.

src/modules/page/room/RoomMembers.page.vue
hoeppner-dataport: don't know if I understood the question correctly. line 147: header: this is the right way to define what comes back... helps if you want to use a property of the HTMLElement.
line 148: headerBottom: the colon means: the property "bottom" from the response should be stored in the variable "headerBottom"
Both lines are correct, I think

src/modules/feature/room/RoomMembers/MembersTable.vue
hoeppner-dataport: Did not know of that consequence. Like the idea! Will do.

src/modules/feature/room/RoomMembers/MembersTable.vue
```diff
@@ -148,10 +153,13 @@ const emit = defineEmits<{
 }>();
 
 const { t } = useI18n();
-const { xs: isExtraSmallDisplay } = useDisplay();
+const { xs: isExtraSmallDisplay, mdAndDown: isMobileDevice } = useDisplay();
 const search = ref("");
 const memberList = toRef(props, "members");
 const membersFilterCount = ref(memberList.value?.length);
+const stickyStyle = computed(() => ({
+	top: `${props.headerBottom + 1}px`,
```
muratmerdoglu-dp, line 161: The extra 1px addition causes a gap that shows the scrolling text above the input's margin.
image

src/modules/feature/room/RoomMembers/MembersTable.vue
```diff
@@ -35,6 +35,7 @@
 	</div>
 
 	<v-divider role="presentation" />
+	{{ stickyStyle }}
```
NFriedo, line 38: I guess this is/was for testing purposes. We shouldn't forget to remove it :D

src/components/templates/DefaultWireframe.vue
hoeppner-dataport: removed it

src/modules/page/room/RoomMembers.page.vue
hoeppner-dataport: No - because it is hypothetical that it can be null. We define it in line 8. If somebody removes it the code should break.

MartinSchuhmacher (approved):

muratmerdoglu-dp (approved):

PR #3654, opened 2025-04-25
BC-9026 - room invitation link store

hoeppner-dataport:
# Short Description
Implementing the store to interact with the room invitation links endpoints.

BC-9026
https://github.com/hpi-schul-cloud/schulcloud-server/pull/5657

## Checklist before merging


src/modules/data/room/roomMembers/RoomInvitationLink.store.ts
```diff
@@ -0,0 +1,128 @@
...
+
+		const roomApi = RoomApiFactory(undefined, "/v3", $axios);
+		const api = RoomInvitationLinkApiFactory(undefined, "/v3", $axios);
+
+		const getRoomId = () => {
+			if (!roomId.value) {
+				throw new Error("RoomDetailStore is not initialized");
+			}
+			return roomId.value;
+		};
```
muratmerdoglu-dp, line 34: I think this can be controlled on `onMounted` lifecycle to make sure that if `room.value?.id` exists or not. Then, using `roomId` in line #21 should be enough.

src/modules/data/room/roomMembers/RoomInvitationLink.store.ts
```diff
@@ -0,0 +1,128 @@
...
+
+		const initStore = async () => {
+			if (roomId.value === undefined) {
+				setTimeout(() => initStore(), 100);
+			} else {
+				await fetchLinks();
+			}
+		};
```
muratmerdoglu-dp, line 42: Probably this `setTimeout` will not be necessary if we check roomId on `onMounted` described above.

src/modules/data/room/roomMembers/RoomInvitationLink.store.ts
```diff
@@ -0,0 +1,128 @@
...
+
+				roomInvitationLinks.value = roomInvitationLinks.value.map((l) =>
+					l.id === link.id ? response : l
+				);
+			} catch {
+				showFailure(t("pages.rooms.members.error.update"));
+			}
+		};
+
+		const deleteLink = async (linkId: string) => {
```
muratmerdoglu-dp, line 90: Is it possible to make the parameter a list for further bulk deletion?

src/modules/data/room/roomMembers/RoomInvitationLink.store.unit.ts
```diff
@@ -0,0 +1,294 @@
...
+					roomId: "room-id",
+					title: "Updated Link",
+					activeUntil: new Date("2023-12-31").toISOString(),
+					isOnlyForTeachers: true,
+					restrictedToCreatorSchool: true,
+					requiresConfirmation: true,
+					creatorUserId: "user-id",
+					creatorSchoolId: "school-id",
+				},
+			];
```
muratmerdoglu-dp, line 256: Is it possible to use `roomInvitationLinkFactory` here?

src/modules/feature/room/RoomMembers/Invitations.vue
```diff
@@ -2,9 +2,83 @@
...
+		Math.ceil(Math.random() * 1000 * 3600) * (Math.ceil(Math.random()) * -1);
+	roomInvitationLinkStore.updateLink({
+		id: linkId,
+		title: title,
+		activeUntil: new Date(Date.now() + maxOneOur).toISOString(),
+		isOnlyForTeachers: Math.random() > 0.5,
+		restrictedToCreatorSchool: Math.random() > 0.5,
+		requiresConfirmation: Math.random() > 0.5,
+	});
+};
```
muratmerdoglu-dp, line 83: This implementation are only testing the store in UI, isn't it?

src/modules/data/room/roomMembers/RoomInvitationLink.store.ts
hoeppner-dataport: getRoomId is only for the other functions to avoid:
`if (roomId.value === undefined) return`
because `roomId.value` is `string | undefined`

src/modules/data/room/roomMembers/RoomInvitationLink.store.ts
hoeppner-dataport: onMounted is not enough because the code is not dependend on the component being rendered but on the other store having loaded the room-details.

src/modules/feature/room/RoomMembers/Invitations.vue
hoeppner-dataport: Yes. it is used to test it visually for Stella, Anna and Uzair.

src/modules/data/room/roomMembers/RoomInvitationLink.store.ts
hoeppner-dataport: Good point. Will improve the endpoint and the store-method.

hoeppner-dataport: done

PR #3657, opened 2025-04-28
BC-9031 - FE: Error messages for RoomInvitationLinks failing

MartinSchuhmacher:
# Short Description
The user should understand when and why the roomInvitationLink does not work.

BC-9031

## Checklist before merging


src/modules/page/room/RoomInvitationLinkStatus.page.unit.ts
```diff
@@ -0,0 +1,135 @@
...
+		boardNotifierCalls = createMock<ReturnType<typeof useBoardNotifier>>();
+		boardNotifier.mockReturnValue(boardNotifierCalls);
+
+		const wrapper = mount(RoomInvitationLinkStatusPage, {
+			attachTo: document.body,
+			global: {
+				plugins: [
+					createTestingVuetify(),
+					createTestingI18n(),
+					createTestingPinia({
```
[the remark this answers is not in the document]
MartinSchuhmacher: sadly this gives us a syntax error as we can firstly access the store when we mounted the component. we can set the `createTestingPinia` before the mount but the rest needs to move after the mount. 

and without the `await nextTick()` we would need the store being present during onMounted, which is not possible due to the above reasons. 

But I am happy to talk on monday in case I missunderstood the idea or did something wrong :)

src/modules/page/room/RoomInvitationLinkStatus.page.vue
```diff
@@ -0,0 +1,126 @@
...
+		title: t("pages.rooms.invitationLinkStatus.title"),
+		disabled: true,
+	},
+];
+
+onMounted(() => {
+	useLink();
+});
+
+const determineStatus = (status: string) => {
```
hoeppner-dataport, line 94: perhaps better updateInfomessage

src/modules/page/room/RoomInvitationLinkStatus.page.vue
```diff
@@ -0,0 +1,126 @@
...
+			break;
+		default:
+			infoMessage.value = t("error.generic");
+			break;
+	}
+};
+</script>
+
+<style lang="scss" scoped>
+.text-centered {
```
hoeppner-dataport, line 123: class is not used

src/modules/page/room/RoomInvitationLinkStatus.page.vue
hoeppner-dataport: renamed

src/modules/page/room/RoomInvitationLinkStatus.page.vue
hoeppner-dataport: removed

PR #3662, opened 2025-04-29
BC-5097 - Fix spacing for headings ins cards

f-g-s:
# Short Description
- fix typo spacing for text in board cards
- no more "jumping" lines between view and edit mode

https://ticketsystem.dbildungscloud.de/browse/BC-5097

## Changes
image
image
image

## Checklist before merging


src/modules/feature/board-text-element/RichTextContentElement.vue
```diff
@@ -160,3 +167,15 @@ const onKeyUp = () => ensurePoliteNotifications();
 	}
 }
 </style>
+
+<style>
```
virgilchiriac, line 171: so this section is not scoped, need to understand why...

f-g-s: You're right, I think it's necessary to access the first parent div. Without access to this information, the first heading of a card can't be styled using CSS.

src/modules/feature/board-text-element/RichTextContentElement.vue
```diff
@@ -160,3 +167,15 @@ const onKeyUp = () => ensurePoliteNotifications();
...
+.v-card-text > div:first-child .ck-content h4:first-child,
+.v-card-text > div:first-child .ck-content h5:first-child {
```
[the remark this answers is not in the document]
f-g-s: I think the same. I'll ask Nicole if that's okay. It's a bit hacky.

src/modules/feature/board-text-element/RichTextContentElement.vue
```diff
@@ -99,6 +99,7 @@ const onKeyUp = () => ensurePoliteNotifications();
 		overflow-x: auto;
 		overflow-y: hidden;
 		padding-right: 1px;
+		margin-bottom: var(--space-xs);
```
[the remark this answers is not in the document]
f-g-s: That's true, but unfortunately it's not so easy to replace in this case, because with deep we don't style the component directly but the element inside it (text).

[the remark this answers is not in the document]
odalys-dataport: My point is, that instead of using the variables `--space-xx` we use exact values like "8px". Otherwise this is just adding more work for our future tickets about removing these variables.

odalys-dataport (changes requested):

src/modules/feature/board-text-element/RichTextContentElement.vue
```diff
@@ -108,31 +109,43 @@ const onKeyUp = () => ensurePoliteNotifications();
...
-		right: 0.5rem;
+		bottom: 8px;
+		right: 8px;
 	}
 }
 </style>
+<style>
+/* Set margin for first headings in cards */
+.v-card-text > div:first-child .ck-content h4:first-child,
+.v-card-text > div:first-child .ck-content h5:first-child {
```
NFriedo, line 148: I think we can (and probably should :D) avoid this global style here. We have the elementIndex as prop here, so we know if it's the first element or not.

So we can add in the template to `RichTextContentElementDisplay ` and  `RichTextContentElementEdit`
```
:class="{ 'first-element': isFirstElement }"
```

in the script: 
```ts
const isFirstElement = computed(() => props.elementIndex === 0);
```

and in the scoped style we could either do:
```scss
.first-element > :is(h4, h5):first-child {
	margin-top: 0;
}
```
or 
```scss
:deep(.first-element.ck-content h4:first-child),
:deep(.first-element.ck-content h5:first-child) {
	margin-top: 0;
}
```

src/modules/feature/board-text-element/RichTextContentElement.unit.ts
```diff
@@ -85,6 +91,32 @@ describe("RichTextContentElement", () => {
...
+				elementIndex: 0,
+			});
+			const displayComponent = wrapper.findComponent(
+				RichTextContentElementDisplayComponent
+			);
+
+			expect(displayComponent.classes()).toContain("first-element");
+		});
+
+		it("should not add first-element class when element is not first element", () => {
```
odalys-dataport, line 108: @NFriedo Can we please adjust these so that for every "when" in an `it`-statement, instead we have another `describe`

```
describe("and element is not first element", () => {
  it("it should not ...", () => {
  ...
  ```

odalys-dataport (changes requested):

odalys-dataport (approved):

PR #3670, opened 2025-05-07
BC-9024 - Implementing Room Invitation Links Table

muratmerdoglu-dp:
# Short Description

- https://ticketsystem.dbildungscloud.de/browse/BC-9024

## Checklist before merging


src/modules/data/room/roomMembers/types.ts
```diff
@@ -14,6 +14,16 @@ export type RoomMember = RoomMemberResponse & {
 
 export type RoomInvitationLink = RoomInvitationLinkResponse;
 
+export type RoomInvitationFormData = {
+	activeUntil: string;
+	activeUntilChecked: boolean;
+	isAlsoForStudents: boolean;
```
MartinSchuhmacher, line 20: how about something a bit more specific here? like `validForStudents`? at least the "is" in the variable somehow sounds to general to me

src/modules/feature/room/RoomMembers/InvitationTable.vue
```diff
@@ -0,0 +1,187 @@
...
+		:show-select="true"
+		:external-selected-ids="selectedIds"
+		@update:model-value="updateSelectedIds"
+	>
+		<template #[`action-menu-items`]>
+			<KebabMenuActionRemoveMember @click="onRemoveLinks(selectedIds)" />
+		</template>
+
+		<template #[`item.title`]="{ item }">
+			<span :class="{ 'text-medium-emphasis': item.isExpired }">
```
MartinSchuhmacher, line 15: is it supposed to check on every template position if the item is expired? maybe we can check that more general as a computed value below and hand it into the template? just reads weirdly when e.g. "item.title" checks "item.isExpired" in the same position

src/modules/feature/room/RoomMembers/InvitationTable.vue
```diff
@@ -0,0 +1,187 @@
...
+		"pages.rooms.members.invitationTable.multipleDelete.confirmation"
+	);
+	if (linkIds.length === 1) {
+		const invitationTitle = invitationTableData.value.find(
+			(link) => link.title === linkIds[0]
+		)?.title;
+		message = t("pages.rooms.members.invitationTable.delete.confirmation", {
+			invitation: invitationTitle,
+		});
+	}
```
MartinSchuhmacher, line 127: shall we put that lines from the function until here into a separate method for readability?

src/modules/feature/room/RoomMembers/InvitationTable.vue
```diff
@@ -0,0 +1,187 @@
+<template>
+	<DataTable
+		:items="invitationTableData"
+		:header-bottom="headerBottom"
+		:table-headers="tableHeaders"
+		:show-select="true"
+		:external-selected-ids="selectedIds"
+		@update:model-value="updateSelectedIds"
```
MartinSchuhmacher, line 8: not sure about the current way of handling it, but should this be `onUpdateSelectedIds`?

src/modules/feature/room/RoomMembers/InvitationTable.vue
```diff
@@ -0,0 +1,187 @@
...
+
+		<template #[`item.actions`]="{ item }">
+			<div
+				class="d-flex align-center"
+				:class="{ 'text-medium-emphasis': item.isExpired }"
+			>
+				<v-btn
+					variant="text"
+					:icon="mdiShareVariantOutline"
+					@click="openShareModal(item.id)"
```
MartinSchuhmacher, line 53: see above `on...`

src/modules/feature/room/RoomMembers/InvitationTable.vue
muratmerdoglu-dp: These templates are iterative for creating table rows. To use a function to get its value, we need to pass at least `item.id` as a parameter to the function and return a boolean after filtering the whole table data. I think it would increase the complexity. 
To use a computed property, since computed properties don't accept parameters, using it here might not be appropriate.

src/modules/feature/room/RoomMembers/InviteMembersDialog.vue
```diff
@@ -80,6 +80,11 @@
 								v-model="formData.activeUntil"
 								:disabled="isDatePickerDisabled"
 								:min-date="new Date().toString()"
+								:date="
```
MartinSchuhmacher, line 83: maybe we want that into a separate method?

src/modules/feature/room/RoomMembers/InviteMembersDialog.vue
```diff
@@ -223,9 +222,19 @@ const isDatePickerDisabled = computed(() => {
 });
 
 const modalTitle = computed(() => {
-	return step.value === InvitationStep.PREPARE
-		? t("pages.rooms.members.inviteMember.firstStep.title")
-		: t("pages.rooms.members.inviteMember.secondStep.title");
+	const titleMap = {
```
MartinSchuhmacher, line 225: do we want `titleMap` declared outside of this computation?

src/modules/feature/room/RoomMembers/InviteMembersDialog.vue
```diff
@@ -273,22 +297,35 @@ const { pause, unpause, deactivate } = useFocusTrap(inviteMembersContent, {
...
 	}
 );
 
 watch(
-	() => isOpen.value,
-	(newValue: boolean) => {
-		if (!newValue) {
-			deactivate();
+	() => editedLink.value,
+	(newVal) => {
```
MartinSchuhmacher, line 309: how about a separate method here for better readability? or would that destroy the watch functionality?

src/modules/ui/kebab-menu/KebabMenuActionShare.vue
```diff
@@ -3,7 +3,7 @@
 		:icon="mdiShareVariantOutline"
 		data-testid="kebab-menu-action-share"
 	>
-		{{ t("common.actions.shareCopy") }}
+		{{ text ?? t("common.actions.shareCopy") }}
```
MartinSchuhmacher, line 6: can we have a separate method for that as well?

src/modules/feature/room/RoomMembers/InviteMembersDialog.vue
muratmerdoglu-dp: is simplified :)

MartinSchuhmacher (approved):

src/modules/ui/kebab-menu/index.ts
```diff
@@ -4,6 +4,7 @@ import KebabMenuActionChangeLayout from "./KebabMenuActionChangeLayout.vue";
 import KebabMenuActionChangePermission from "./KebabMenuActionChangePermission.vue";
 import KebabMenuActionCopy from "./KebabMenuActionCopy.vue";
 import KebabMenuActionDelete from "./KebabMenuActionDelete.vue";
+import KebabMenuActionDeleteMemberInvitation from "./KebabMenuActionDeleteMemberInvitation.vue";
```
hoeppner-dataport, line 7: Max Bischof told me, that he thinks that components, that can only be used in one place of the application, should not be placed in the ui-kebab-menu-module. I agreed to that and had a first glance at the different actions we have here: i think for some of the components only the name of the component is too specific... 
We should go through this and talk about this. But you can merge the PR.

hoeppner-dataport (approved):

PR #3676, opened 2025-05-08
BC-9081 - FE: Enable adding of students to a room (incl. Connecting Student Endpoint)

MartinSchuhmacher:
# Short Description
In the dialog to add members to a room, users should be able to select the role students and then pick students from the currently selected school (at least the ones the teacher should be able to see) to add them.

BC-9081
https://github.com/hpi-schul-cloud/schulcloud-server/pull/5689

## Screenshots of UI changes
grafik

## Checklist before merging


src/assets/img/CrossedHandsSvg.vue
```diff
@@ -0,0 +1,143 @@
+<template>
+	<div>
```
NFriedo, line 2: I think we don't need the div here :)

src/assets/img/CrossedHandsSvg.vue
```diff
@@ -0,0 +1,143 @@
...
+		<svg
```
NFriedo, line 3: We should add `aria-hidden=true` because that is just decorative graphic.

src/modules/feature/room/RoomMembers/AddMembers.vue
NFriedo, line 1: Hm I don't know where it comes from but the screenreader output is strange on the VSelect: 
Navigating to Schulrolle Select reads out "Lernend" and when navigating further it reads out "student" (same for teacher, reading out Lehrkraft and then "teacher".

My first thought was that maybe we need a vuetiy update because they are improving VSelect and VAutocpmplete constantly but I  also checked a VSelect in a vuetify docs with our version and there the issue doesn't occur 🤔 

Also on main it doesn't happen. The input is still weird there (double output) but at least everything is correctly translated. So maybe we need to have a look into our implementation again.

src/modules/feature/room/RoomMembers/AddMembers.vue
```diff
@@ -17,30 +17,34 @@
...
 				/>
 			</div>
 
 			<div class="mt-4" data-testid="add-participant-role">
-				<v-autocomplete
-					ref="autoCompleteRole"
+				<v-select
+					ref="selectRole"
 					v-model="selectedSchoolRole"
 					auto-select-first="exact"
```
NFriedo, line 31: the prop `auto-select-first` can be removed because it's just an `VAutocomplete` prop :)

src/modules/feature/room/RoomMembers/AddMembers.vue
NFriedo: Another strange thing: when going in the selection it reads out:
"Lernend selected, 2 of 4"
"Lehrkraft selected, 3 0f 4"
(tested with Safari/VoiceOver)  

🤔 

Okay really really weird, but I checked this also on vuetify and it seems to be a vuetify issue on Safari with Voiceover. Chrome looks fine.
So we should probably open a vuetify issue/ have a look into the open issues. 
So nothing that can be fixed in this PR 😅

NFriedo: Seems like the first issue is also a vuetify issue: https://github.com/vuetifyjs/vuetify/issues/20656

MartinSchuhmacher: hint for your review @muratmerdoglu-dp 
we discuss several solutions currently to make it work but none of them works as expected, my suggestion is to wait for the vuetify update to not implement something totally hacky to make the SR work

muratmerdoglu-dp (approved):

src/modules/data/room/roomMembers/RoomMembers.store.ts
```diff
@@ -106,18 +107,29 @@ export const useRoomMembersStore = defineStore("roomMembersStore", () => {
...
 				.map((user) => {
 					return {
 						...user,
 						userId: user.id,
 						fullName: `${user.lastName}, ${user.firstName}`,
-						schoolRoleNames: [RoleName.Teacher],
-						displayRoomRole: roomRole[RoleName.Roomadmin],
-						displaySchoolRole: schoolRoleMap[RoleName.Teacher],
+						schoolRoleNames: [schoolRoleName],
+						displayRoomRole:
```
hoeppner-dataport, line 128: Remove displayRole from potentialRoomMembers, as it will be overwritten after addMembers by the role returned from the backend.

src/modules/feature/room/RoomMembers/AddMembers.vue
```diff
@@ -85,14 +96,25 @@
...
+import {
+	VAutocomplete,
+	VCard,
+	VSelect,
+} from "vuetify/lib/components/index.mjs";
+import { InfoAlert, WarningAlert } from "@ui-alert";
 import { storeToRefs } from "pinia";
 
+defineProps({
+	canAddAllStudents: {
```
hoeppner-dataport, line 112: This should not be a property. We use the RoomAuthorization-composable wherever we need to find out, if the user is allowed to do certain things. Please refactor this.

hoeppner-dataport (changes requested):
See two comments regarding:
- removing displayRoomRole from potentialRoomMembers
- not passing permissions as a property

src/modules/data/room/roomMembers/RoomMembers.store.ts
MartinSchuhmacher: removed

src/modules/feature/room/RoomMembers/AddMembers.vue
MartinSchuhmacher: refactored

hoeppner-dataport (approved):

PR #3678, opened 2025-05-08
N21-2453 add disable logic for external logout

GordonNicholasCap:
# Short Description
- add logic to disable external logout when session token is expired

- https://ticketsystem.dbildungscloud.de/browse/N21-2453
- https://github.com/hpi-schul-cloud/schulcloud-server/pull/5675
- https://github.com/hpi-schul-cloud/schulcloud-client/pull/3633

## Checklist before merging


src/modules/data/oauth/OAuthApi.composable.ts
```diff
@@ -0,0 +1,34 @@
...
+
+			const expirationDate = new Date(response.data.expiresAt);
+
+			return expirationDate;
+		} catch (error) {
+			if (
+				!isAxiosError(error) ||
+				error.response?.status !== HttpStatusCode.NotFound
+			) {
+				handleError(error, {
```
MarvinOehlerkingCap, line 26: Why use handleError if it is supposed to do nothing? If you don't need to process the error then just `return undefined` as a catch case

MarvinOehlerkingCap (changes requested):

src/modules/data/oauth/oauthApi.composable.ts
```diff
@@ -0,0 +1,25 @@
...
+
+	const getSessionTokenExpiration = async (): Promise<Date | undefined> => {
+		try {
+			const response: AxiosResponse<OAuthSessionTokenExpirationResponse> =
+				await oauthApi.oAuthControllerGetSessionTokenExpiration();
+
+			const expirationDate = new Date(response.data.expiresAt);
+
+			return expirationDate;
+		} catch (error) {
```
MarvinOehlerkingCap, line 19: there is a linter warning here. `'error' is defined but never used`

MarvinOehlerkingCap (approved):

PR #3722, opened 2025-06-02
BC-9711 - Room Invitation Dialog: No User Feedback on Text Character Limit Exceed + make field mandatory

MartinSchuhmacher:
# Short Description
The "Room Invitation" dialog box includes an optional text field for users to add a message when inviting others. Currently, this field has a character limit of 100 characters. However, the user interface does not provide any indication or warning when a user inputs text exceeding this limit.

The text fields should be mandatory going forward and provide good validation feedback for the users.

BC-9711

## Changes
- implementing validation for room invitation link description
-- e.g. required, non-whitespace-only, opening-tag

## Screenshots of UI changes
e.g.

grafik

## Checklist before merging

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,7 +62,7 @@
 				"cross-env": "^7.0.3",
 				"css-loader": "^6.10.0",
 				"eslint": "^9.27.0",
-				"eslint-config-prettier": "^9.1.0",
+				"eslint-config-prettier": "^10.1.5",
 				"eslint-plugin-prettier": "^5.4.0",
 				"eslint-plugin-vue": "^9.33.0",
 				"eslint-webpack-plugin": "^5.0.1",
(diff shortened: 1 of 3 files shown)


src/modules/feature/room/roomMembers/dialogs/InviteMembersDialog.vue
```diff
@@ -212,6 +219,28 @@ const defaultFormData = {
...
+			containsOpeningTag: helpers.withMessage(
+				t("common.validation.containsOpeningTag"),
+				(name: string) => !containsOpeningTagFollowedByString(name)
+			),
+			required: helpers.withMessage(t("common.validation.required2"), required),
+		},
+	},
+}));
+
+const v$ = useVuelidate(
```
NFriedo, line 238: I am just curious is there a reason why we use vuelidate and not the vuetify validation?

MartinSchuhmacher: no specific reason to be honest
1. saw that it was used in a lot of places (especially in the room create page)
2. personally think its more readable

but as said both reasons are highly subjective to me but I prefer vuelidate

NFriedo: Okay I just remember a discussion in the frontend exchange if we want to keep vuelidate or use vuetify validation. I thought we want to use vuetify's validation if possible.

There also already some places that uses vuetify validation like the LinkContentElementCreate.vue or the GeneralSettings.vue.

But seems like the only outcome of the topic was a ticket creation: https://docs.dbildungscloud.de/display/DBH/Meeting+2025-02-19 and I can't find a ticket about it. So seems like it is not decided or researched yet. Maybe it would be a good meeting topic again. 
In my opinion we should try using the vuetify validation if it's possible in that case, but that is just a personal preference and you could also leave it as it is 😅

src/modules/feature/room/roomMembers/dialogs/InviteMembersDialog.unit.ts
```diff
@@ -65,6 +66,14 @@ describe("InviteMembersDialog", () => {
...
 
+	const setDescription = async (wrapper: VueWrapper, value: string) => {
+		const descriptionField = wrapper.findComponent({
+			ref: "descriptionField",
+		});
+		await descriptionField.setValue(value);
+		await nextTick();
+	};
+
 	const setup = (
```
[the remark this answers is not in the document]
MartinSchuhmacher: will adapt that

PR #3743, opened 2025-06-11
BC-9807 - Rename Copy to Duplicate

f-g-s:
# Short Description

We're currently a bit inconsistent in the naming of the copy function. It should be called "Duplicate" everywhere (as in the room), since that's what actually happens—not "Copy."

https://ticketsystem.dbildungscloud.de/browse/BC-9807

## Changes
- text changes for de, en, es and some for uk
- longer sentences have not been adapted in Ukrainian because the translation is likely to be harder to understand

## Screenshots of UI changes
image
image
image
image

## Checklist before merging


src/locales/en.ts
```diff
@@ -6,6 +6,7 @@ export default {
 	"common.actions.confirm": "Confirm",
 	"common.actions.continue": "Continue",
 	"common.actions.copy": "Copy",
+	"common.actions.duplicate": "Duplicate",
```
odalys-dataport, line 9: Awesome, now we only need to remove the language key in line 14 and replace any usage with this one. :)

src/locales/es.ts
```diff
@@ -682,7 +683,7 @@ export default {
 	"components.molecules.copyResult.etherpadCopy.info":
 		"El contenido no se copia por razones de protección de datos y debe agregarse nuevamente.",
 	"components.molecules.copyResult.failedCopy":
-		"No se pudo completar el proceso de copia.",
+		"NNo se pudo completar la duplicación.",
```
odalys-dataport, line 686: That's one "N" too much :)

odalys-dataport (changes requested):

odalys-dataport (approved):

PR #3791, opened 2025-07-17
BC-9738 - Room editors may only create/start video conferences if permission is granted

MartinSchuhmacher:
# Short Description
Allow create/start video conference on boards for room editors only if room setting `editor_manage_videoconference` is active.
Remaining board roles should not be altered.

BC-9738

## Changes
- implementing BoardManageVideoconference permission into board permission workflow
- videoconference element in add element dialog only present according to permission
- videoconference element on board can only be started with correct permission
- UI adjustments on the room form

## Checklist before merging

### DIFF
--- a/src/modules/feature/board-video-conference-element/components/VideoConferenceContentElement.unit.ts
+++ b/src/modules/feature/board-video-conference-element/components/VideoConferenceContentElement.unit.ts
@@ -118,7 +118,7 @@ describe("VideoConferenceContentElement", () => {
 			elementIndex = 2,
 			isRunning = false,
 			error = null,
-			hasManageVideoConferencePermission,
+			hasManageVideoConferencePermission = false,
 		} = options;
 
 		const element = {


src/components/organisms/Ldap/LdapRolesSection.vue
```diff
@@ -23,12 +23,14 @@
 					"
 					value="group"
 					class="ml-n2"
+					color="primary"
```
NFriedo, line 26: Maybe it makes sense to add VRadio to the defaults here: src/plugins/vuetify.ts

```ts
defaults: {
  VRadio: {color:"primary}
},
```

So we don't need to write `color=primary` to each `VRadio`

src/modules/feature/room/RoomForm.vue
```diff
@@ -49,12 +49,20 @@
...
+					:class="{
+						'video-conference-checkbox--sm': !smAndUp,
+					}"
 					data-testid="room-video-conference-checkbox"
 					:label="t('components.roomForm.labels.videoConference.label')"
 					:hide-details="true"
 					@click="onToggleVideoConferenceFeature"
 				/>
-				<div class="text-grey-darken-1 ms-8 helper-text">
+				<div
```
NFriedo, line 60: Hmm I think this helper text should be part of the checkbox label.
We could use the label slot for the v-checkbox for it.

src/components/organisms/Ldap/LdapRolesSection.vue
MartinSchuhmacher: actually we need `VRadioGroup` it seems as we use the VRadio within the VRadioGroup everytime, so maybe we later need to add it for the `VRadio` separate once we need it. Interesting enough that we dont have the same behavior for `VCheckbox` there

NFriedo (approved):

PR #3832, opened 2025-08-18
BC-7402 - remove old breakpoints

odalys-dataport:
# Short Description

To have a more consistent design setup and remove some overhead we want to remove the breakpoints defined in our project and replace them with vuetifys defined breakpoints.

BC-7402

## Checklist before merging


src/components/legacy/NavigationBar.vue
```diff
@@ -100,26 +100,26 @@ const setActive = (idx: number) => {
 }
 
 .nav-container {
-	@include breakpoint(tablet) {
+	@media (min-width: 750px) {
```
dyedwiper, line 103: In other files you use the vuetify breakpoints with sass:map. Why not here?

src/pages/administration/StudentOverview.page.vue
```diff
@@ -739,22 +737,6 @@ export default {
 	justify-content: center;
 }
 
-.info-box {
```
dyedwiper, line 742: The two classes above are also unused AFAIS and you could remove them, too.

src/modules/feature/news-form/FormActions.vue
```diff
@@ -12,14 +12,15 @@
 </template>
 
 <style lang="scss" scoped>
-@use "@/styles/mixins" as *;
+@use "sass:map";
+@use "@/styles/settings.scss" as *;
```
dyedwiper, line 16: Why is this import needed?

src/components/lern-store/LernStoreGrid.vue
```diff
@@ -20,18 +20,19 @@ const col = computed(() => {
...
+@use "sass:map";
+@use "@/styles/settings.scss" as *;
 
 .grid {
 	display: grid;
 	grid-gap: 16px;
 	width: 100%;
 
-	@include breakpoint(tablet) {
+	@media #{map.get($display-breakpoints, 'sm')} {
```
dyedwiper, line 31: In `_typography.scss` you have a similar case of two possibly conflicting media queries and there you use `@media #{map.get($display-breakpoints, 'sm-and-up')}`. It leads to the same result because the latter rule takes precedence anyway. I'm just wondering if we want to be consistent in such cases.

src/components/legacy/NavigationBar.vue
odalys-dataport: This component seems to rely on the old breakpoints quite heavily. I started looking into refactoring it all, but eventually decided it was gonna be too much effort considering we might need to apply any changes to the legacy client as well.

src/components/lern-store/LernStoreGrid.vue
odalys-dataport: I'm not sure we need to "force" consistency, but for the sake of clarity I changed the line in `_typography.scss`

dyedwiper (approved):

src/modules/feature/auto-logout/AutoLogoutWarning.vue
```diff
@@ -119,7 +120,7 @@ watch(
 		width: 100%;
 		font-size: var(--text-lg);
 
-		@include breakpoint(tablet) {
+		@media #{map.get($display-breakpoints, 'sm-and-up')} {
```
dyedwiper, line 123:
```suggestion
		@media #{map.get($display-breakpoints, 'md-and-up')} {
```
Otherwise it gets too narrow and doesn't look good.

PR #3839, opened 2025-08-20
BC-7404 Layers cleanup

wolfganggreschus:
# Short Description

BC-704
## Checklist before merging

### DIFF
--- a/src/styles/css-variables/_layers.scss
+++ b/src/styles/css-variables/_layers.scss
@@ -5,7 +5,6 @@
 	--layer-dropdown: 30;
 	--layer-popover: 40;
 	--layer-tooltip: 50;
-	--layer-top-menu: 60;
 	--layer-fab: 100;
 	--layer-modal: 1000;
 }


src/components/organisms/StepProgress.vue
```diff
@@ -109,20 +109,15 @@ ul.progressbar {
...
+		left: -30%;
+		width: 60%;
 		height: calc(3 * var(--border-width));
 		content: "";
 		background: map.get($grey, base);
 	}
 }
 
 .progressbar li.active {
-	&::after {
```
odalys-dataport, line 122: Why was this deleted?

src/components/molecules/AlertContainer.vue
```diff
@@ -56,7 +56,7 @@ const onRemoveNotification = (notification: AlertPayload) => {
 	right: 0;
 	bottom: 5vh;
 	left: 0;
-	z-index: var(--layer-tooltip);
```
odalys-dataport, line 59: For newer components please double check if there is a way to manage these via Vuetify and if they align with Vuetify's layering.

src/components/templates/DefaultWireframe.vue
```diff
@@ -194,7 +194,7 @@ const showDivider = computed(() => {
 .sticky {
 	position: sticky;
 	top: var(--topbar-height);
-	z-index: var(--layer-sticky-header);
```
odalys-dataport, line 197: While this is semi a semi old component, this is still widely used and applied to new features. Maybe there is a vuetify class for stickyness by now that can be applied instead? In some cases it can also still make sense to keep variables, especially for components that are new or commonly used, so that we don't have random layering. For legacy code like Lernstore or AdminTables I personally wouldn't put in the effort, but here it could make sense.

wolfganggreschus: Franz and I have looked at it. The idea is to make the header no longer sticky -> another ticket (UX is thinking about it)
As a result, the whole sticky class could then be removed, which leads to other current problems.

Keeping the global variables now works against the current ticket, doesn't it? I would leave it as it is for now, but Franz will take the “wirefame” issue to the UX round.

PR #3847, opened 2025-08-25
WIP: BC-10453 Minor a11y fixes

sebmue-dataport:
# Short Description

 - Fix: Pressing escape instead of clicking the cancel button in the Room Copy Info Dialog caused the dialog to close but kept it from ever being opened again in the same session. Fixed by using after-leave on VDialog
 - Refactor: Several legacy uses of $t for using i18n replaced with t
 - Refactor: Simplify the Share Dialog and remove dependency towards Custom Dialog using VDialog instead

https://ticketsystem.dbildungscloud.de/browse/BC-10453

## Checklist before merging


hoeppner-dataport (changes requested):

NFriedo:
Will close this PR, because it is really old and the changed Dialogs are either not exist anymore or repalced with SVSDialog

PR #4156, opened 2026-04-09
BC-11513 - Migration VueX VideoConferenceModule to composable

MartinSchuhmacher:
# Short Description
Vuex-Module 'VideoConferenceModule' -> composable

BC-11513

## Changes
- replacing VueX VC module (course scope) with already existing composable (used in board scope)
- refactoring composable with new async-task wrapper for easier handling
- moving rewritten VC composable to new general file placement `src/modules/data/data-access`
- refactoring course scope and board scope component usages for new composable handling

## Checklist before merging


src/components/course-rooms/tools/RoomVideoConferenceSection.vue
```diff
@@ -62,91 +61,78 @@ const props = defineProps({
...
-});
+const onCloseErrorDialog = () => {
+	errorDismissed.value = true;
+};
 
-const onRefresh = async () => {
-	if (isRefreshing.value) {
-		return;
-	}
+const isConfigurationDialogOpen: Ref<boolean> = ref(false);
```
luejoh: infer type?

src/modules/data/data-access/VideoConference.composable.ts
```diff
@@ -0,0 +1,86 @@
+import { useSafeAxiosTask } from "@/composables/async-tasks.composable";
+import { useI18nGlobal } from "@/plugins/i18n";
+import { VideoConferenceInfo, VideoConferenceOptions, VideoConferenceState } from "@/store/types/video-conference";
+import { $axios } from "@/utils/api";
+import { VideoConferenceApiFactory, VideoConferenceScope, VideoConferenceStateResponse } from "@api-server";
+import { computed, onMounted, ref } from "vue";
+
+const videoConferenceStateMapping: Partial<Record<VideoConferenceStateResponse, VideoConferenceState>> = {
```
luejoh: seems to be noisy code. All relevant types are already available from endpoint. Can we remove mapping them?

src/modules/data/data-access/VideoConference.composable.ts
```diff
@@ -0,0 +1,86 @@
...
+				options: result.data.options,
+			};
+			return result;
+		}
+	};
+
+	const startVideoConference = async (options: VideoConferenceOptions) => {
+		const { success } = await execStart(
+			() => videoConferenceApi.videoConferenceControllerStart(scope, scopeId, { ...options }),
+			t("common.notification.error")
```
luejoh: can we have better error messages?

luejoh: side note:
common.notifications.errors.xXx

is covering some default error messages with just a word being different, allowing to add an item name, what kind of thing was failing.

Could work out here too with "{item} konnte nicht gestartet werden"
You could have a look at room.store for examples.

src/components/course-rooms/tools/RoomVideoConferenceSection.vue
MartinSchuhmacher: sometimes I feel like having "more control" but you're right, unnecessary here

src/modules/data/data-access/VideoConference.composable.ts
MartinSchuhmacher: well thats some more changes due to the nature of it being like a "safety net" as a mapping between client and server. Also this weird "UNKNOWN" fallback from the past needs to be figured out than as well (which I never understood why its there). But I will try that.

src/modules/data/data-access/VideoConference.composable.ts
MartinSchuhmacher: will check, thx

MartinSchuhmacher: thought the whole "item" behavior was a bit overengineered in this case, so went with a more direct approach, see if you like that as well

src/modules/data/data-access/VideoConference.composable.ts
MartinSchuhmacher: UNKNOWN was not used anyway...so easier than expected. Removed the translation layer

src/modules/data/data-access/VideoConference.composable.unit.ts
```diff
@@ -9,21 +8,39 @@ import {
...
 		videoConferenceApi = mockApi<serverApi.VideoConferenceApiInterface>();
-
 		vi.spyOn(serverApi, "VideoConferenceApiFactory").mockReturnValue(videoConferenceApi);
 	});
 
 	afterEach(() => {
 		vi.clearAllMocks();
 	});
 
+	const withComponentSetup = <T>(composableFn: () => T): T => {
```
luejoh, line 30: i dont get the benefit of this.
Do we need that for anything?

src/modules/data/data-access/VideoConference.composable.unit.ts
```diff
@@ -336,4 +273,63 @@ describe("VideoConferenceComposable", () => {
 			expect(isWaitingRoomActive.value).toBe(false);
 		});
 	});
+
+	describe("fetchImmediate", () => {
+		it("should call fetchVideoConferenceInfo on mount when fetchImmediate is true (default)", async () => {
```
luejoh, line 278: i really struggle in likeing such tests. Its testing that code in an if block is executed, which is really safe.
I see why we would still wanna do that, i just dont like seeing so much code added for so little value.

What are your feelings?

src/modules/feature/board-video-conference-element/components/VideoConferenceContentElement.vue
```diff
@@ -185,17 +192,23 @@ const onDelete = async () => {
 	if (await askDeletionForType("components.cardElement.videoConferenceElement"))
 		emit("delete:element", computedElement.value.id);
 };
+
 const onStartVideoConference = async () => {
 	const windowReference = window.open();
```
luejoh, line 197: The copilot comment below has some valid arguments

src/modules/ui/video-conference-configuration-dialog/VideoConferenceConfigurationDialog.vue
```diff
@@ -62,7 +62,7 @@ const { t } = useI18n();
 
 defineEmits(["close", "start-video-conference"]);
 
-const localOptions: ComputedRef<VideoConferenceOptions> = computed(() => props.options);
+const localOptions: ComputedRef<VideoConferenceOptionsResponse> = computed(() => props.options);
```
luejoh: remove typing ? Infer it?

luejoh (approved):

src/modules/data/data-access/VideoConference.composable.unit.ts
MartinSchuhmacher: because the test needs a basic component being present to "simulate" the immediate fetching being done via the composable

src/modules/data/data-access/VideoConference.composable.unit.ts
MartinSchuhmacher: I feel you, but I also see the benefit in extensive testing.

src/modules/feature/board-video-conference-element/components/VideoConferenceContentElement.vue
MartinSchuhmacher: you (and copilot) are right, but the issue lies within the way BBB handles these responses. We had that same discussion in the past - with different browsers as topics - and this way was the only one at the time to make it work. And as long as we dont need to I dont want to rumble that box again :D

src/modules/ui/video-conference-configuration-dialog/VideoConferenceConfigurationDialog.vue
MartinSchuhmacher: 👍

src/modules/data/data-access/VideoConference.composable.unit.ts
luejoh: ah ye true, due to onMounted

src/modules/feature/board-video-conference-element/components/VideoConferenceContentElement.vue
luejoh: Ok, enough information for me.

PR #4173, opened 2026-04-20
BC-11401 - Editing link elements

MartinSchuhmacher:
# Short Description

BC-11401
https://github.com/hpi-schul-cloud/e2e-system-tests/pull/525

## Changes
- edit existing link elements
- save/validate when clicking outside the card OR using the btn
- providing positive validation result as well
- refactoring URL validation

## Screenshot
grafik

## Checklist before merging


src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
```diff
@@ -15,83 +18,99 @@
...
-			onKeydown,
-			onSubmit,
-			onKeydownSubmit,
-			form,
-			url,
-			rules,
-			mdiCheck,
-		};
-	},
+const form = ref<VuetifyFormApi | null>(null);
```
luejoh: can you try
const form: TemplateRef<VForm> ? Possibly also without using null as a value

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
```diff
@@ -15,83 +18,99 @@
...
-	components: {},
-	emits: ["create:url"],
-	setup(props, { emit }) {
-		const { t } = useI18n();
-		const url = ref<string>("");
-		const form = ref<VuetifyFormApi | null>(null);
+const props = defineProps({
+	existingUrl: {
+		type: String,
+		default: "",
```
luejoh, line 46: Why defaulting to an empty string?
Empty strings are also behaving differently from undefined/null if it comes to falsyness checks.

Is there a good reason to have an empty string here?

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
```diff
@@ -15,83 +18,99 @@
...
-		};
-	},
+const form = ref<VuetifyFormApi | null>(null);
+const rules = [isRequired(t("common.validation.required2")), isValidUrl()];
+
+const onSubmit = async () => {
+	if (form?.value) {
+		const { valid } = await form.value.validate();
+		if (valid) {
+			isUrlValidated.value = true;
```
luejoh, line 69: i think using the form templateref, you should be able to access the isValid property of the given VForm.

The property updates lazy on form validation.

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
```diff
@@ -15,83 +18,99 @@
...
+			emit("create:url", url.value);
+		}
+	}
+};
+
+const onKeydown = (e: KeyboardEvent) => {
+	if (e.key === "enter" || e.key === "Enter") {
```
luejoh: Isnt it either always "enter" or always "Enter"?

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
```diff
@@ -15,83 +18,99 @@
...
+		onSubmit();
+	} else if (form?.value) {
+		isUrlValidated.value = false;
+		form.value.resetValidation();
+	}
+};
+
+onBeforeUnmount(() => {
+	if (isUrlValidated.value) return;
+	const isValid = rules.every((rule) => rule(url.value) === true);
```
luejoh, line 83: Use the VForm validation functions here?

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
```diff
@@ -15,83 +18,99 @@
...
 	margin-right: -6px;
 }
+
+/* simulating success on validation, similar to error case */
+:deep(.input-success .v-field__outline) {
+	--v-field-border-opacity: 1;
+	color: rgb(var(--v-theme-success));
+}
+
+:deep(.input-success .v-label) {
```
luejoh: using "," to not write the same style twice?

src/modules/feature/board-video-conference-element/components/VideoConferenceContentElement.vue
```diff
@@ -92,8 +92,8 @@ const props = defineProps({
 		required: true,
 	},
 	isEditMode: { type: Boolean, required: true },
-	isNotFirstElement: { type: Boolean, requried: false },
```
luejoh, line 95: Since the property was never written correctly, it gives prove, that the required property with value : false doesant add anything.
The default is, that required is false. Could we just remove it then in all the occurences youve seen?

src/modules/feature/board-link-element/components/LinkContentElement.vue
luejoh, line 25: Could we have this test id on the root element given by the parent. I dont see, why the component should know, "where it is", as it actually does not care. The parent does.

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
MartinSchuhmacher: gives me some typescript errors, especially with the unavailability of `validate()`.
I could use `const form: TemplateRef<VForm> = ref(null)` but not sure where the advantage is.

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
MartinSchuhmacher: unnecessary type widening if we use something "not pragmatic" as from the parent (where the only usage of that `Create` component is) we always get "" :

`const sanitizedUrl = computed(() => (props.element.content.url ? sanitizeUrl(props.element.content.url) : ""));`

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
MartinSchuhmacher: as we already use the property on css stuff (necessary) we would need to have two sources that way (no improvement) and also imo we wont get the correct status there if the user expicitly confirmed the URL via the submit action. But as in the beginning, I dont see an advantage

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
MartinSchuhmacher: nope, we need the synchronous behavior here as we need to catch the emit before the promise resolves

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
MartinSchuhmacher: need to investigate on that, maybe need some refactor

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
MartinSchuhmacher: you're right, we can improve here

src/modules/feature/board-video-conference-element/components/VideoConferenceContentElement.vue
MartinSchuhmacher: you mean you want to remove e.g. isNotFirstElement or simply the required part of the property?
because even if `required: false` does not give us something function wise if its the default, it still gives the reader a clear understanding about (un)necessary properties here. I would leave it in for that reason.

src/modules/feature/board-link-element/components/LinkContentElement.vue
MartinSchuhmacher: that is the button data-testid, which is necessry at least for the cy tests, not sure what you mean in this context by the parent.

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
luejoh: Well VuetifyFormApi is written by ourselves, if vuetify decides to rename "validate" to "validateForm", then we would not recognize. And we add code, duplicate to interface of vuetify.

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
luejoh: Hm? If the user would submit the form, the property will be updated. Were not able to follow your argumentation :/ . What are u referring to in css?

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
luejoh: await validate() ?

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
luejoh: Hm i still dont like it, but will not insist in changing it.

src/modules/feature/board-video-conference-element/components/VideoConferenceContentElement.vue
luejoh: I was speaking about the required part of the property.
Id argue, that when the property is not present, its even easier to recognize ;) . Anything which is not "required:true" is not required ^^.

I wont insist tho. up 2 u

src/modules/feature/board-link-element/components/LinkContentElement.vue
luejoh: partly. The parent renders a "LinkContentElement" and has an id, by the various index properties. The parent knows that its a list and has an index. The child should not care about the rendering of its parents. Can you follow?
Cy could select by "linkelement-id .the-btn-to-click-on"

MartinSchuhmacher: I know what you mean, but as with the above topic of saying "if the required prop is not present it means XYZ" there is some implicity in that argument as well and I dont like code being too implicit and prefer it to being explicit. I guess for the button it was simply the solution to make it "easily solvable" but regarding "knowledge of the position" you are right. I have no strong opinion on that, but I'm no big fan of changing it in general on all the positions. So only if you insist ;)

src/modules/feature/board-video-conference-element/components/VideoConferenceContentElement.vue
MartinSchuhmacher: Ok then I will stick with the current state :)

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
MartinSchuhmacher: did not work, the component is unmounted too fast

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
MartinSchuhmacher: the "success" css is adjusted by the `isUrlValidated` and I like that we can follow the state of that explicitly during the component flow.  Hence I would like to keep it like that. That was my argument :)

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
MartinSchuhmacher: I see, I will try to adjust

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
luejoh: ah the class binding. I had the feeling, that isUrlValidated is not necessary as it is a proxy to the VForm's valid property. I know tho, that the property is not necessarily a reactive / is lazy. So it may be not "easy" or breaks reactivity. I had the feeling to at least try getting rid of the proxy.

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
luejoh: This kinda gives me a bad feeling. It feels like that the component should not unmount. But ye, its not part of the ticket to solve that. That was done already before your ticket. Lets leave it be.

src/modules/feature/board-link-element/components/LinkContentElement.vue
luejoh: I can feel the argument, that it was done for so many other elements and it would need to change for all of them. It bothers me, that it was done that way, but I see the efforts, involved with changing it. Lets leave it be 😂.

src/modules/feature/board-link-element/components/LinkContentElementCreate.vue
MartinSchuhmacher: adjusted here, also wrote that down for the following VC ticket

luejoh (approved):

PR #4176, opened 2026-04-20
BC-11519 Replace systems store by composable

dyedwiper:
# Short Description

BC-11519

## Checklist before merging


src/modules/data/system/system.composable.ts
```diff
@@ -0,0 +1,38 @@
...
+	const { execute } = useSafeAxiosTask();
+
+	const system = ref();
+	const systemName = ref();
+
+	const fetchSystem = async () => {
+		const { result } = await execute(() => systemApi.systemControllerGetSystem(systemId));
+
+		// TODO: Should this if be necessary?
+		if (result) {
```
luejoh: The check is necessary, otherwise the result is possibly undefined. 
Usually the pattern uses the "success" tho.
Furthermore you can consider directly returning the response of "execute", if you are not having state in your composable. State in your composable makes always sense, if you would like to introduce computed values to enhance your data, which everybody else would benefit from.

src/modules/data/system/system.composable.ts
```diff
@@ -0,0 +1,38 @@
...
+export const useSystem = (systemId: string, fetchImmediate = true) => {
+	const systemApi = SystemsApiFactory(undefined, "/v3", $axios);
+
+	const { execute } = useSafeAxiosTask();
+
+	const system = ref();
+	const systemName = ref();
+
+	const fetchSystem = async () => {
+		const { result } = await execute(() => systemApi.systemControllerGetSystem(systemId));
```
luejoh: U should consider adding a error translation key, which automatically raises alerts in the UI, what kind of thing failed. In that case fetching the system data

src/modules/data/system/system.composable.ts
```diff
@@ -0,0 +1,38 @@
+import { useSafeAxiosTask } from "@/composables/async-tasks.composable";
+import { $axios } from "@/utils/api";
+import { SystemsApiFactory } from "@api-server";
+import { onMounted, ref } from "vue";
+
+export const useSystem = (systemId: string, fetchImmediate = true) => {
+	const systemApi = SystemsApiFactory(undefined, "/v3", $axios);
+
+	const { execute } = useSafeAxiosTask();
```
luejoh: id recommend destructering status, error and isLoading as well and returning it.
That allows any user of the composable to write behavior on these meta information

src/modules/data/system/system.composable.ts
```diff
@@ -0,0 +1,38 @@
...
+	const systemName = ref();
+
+	const fetchSystem = async () => {
+		const { result } = await execute(() => systemApi.systemControllerGetSystem(systemId));
+
+		// TODO: Should this if be necessary?
+		if (result) {
+			system.value = {
+				id: result.data.id,
+				displayName: result.data.displayName ?? "",
```
luejoh: why empty string? Its always harder to check on falsyness of emptystrings, instead of e.g. undefined

src/modules/data/system/system.composable.ts
```diff
@@ -0,0 +1,38 @@
...
+				hasEndSessionEndpoint: !!result.data.oauthConfig?.endSessionEndpoint,
+			};
+
+			systemName.value = result.data.displayName ?? "";
```
luejoh: why having the same information twice?

src/modules/data/system/system.composable.unit.ts
```diff
@@ -0,0 +1,73 @@
...
+		oauthConfig: { endSessionEndpoint: "some-endpoint" },
+	},
+};
+
+describe("useSystem", () => {
+	let systemApi: { systemControllerGetSystem: ReturnType<typeof vi.fn> };
+	let execute: ReturnType<typeof vi.fn>;
+
+	beforeEach(() => {
+		systemApi = { systemControllerGetSystem: vi.fn() };
```
luejoh: can you check uwe's "mockComposable"?

src/modules/page/class-members/ClassMembersInfoBox.vue
```diff
@@ -26,56 +26,24 @@
...
-	props: {
-		systemId: {
-			type: String,
-			default: "",
-		},
-	},
-	setup(props) {
-		const { t } = useI18n();
-		const { getSystem } = useSystemApi();
+const { systemId = "" } = defineProps<{ systemId?: string }>();
```
luejoh: why empty string?

src/modules/page/class-members/ClassMembersInfoBox.vue
```diff
@@ -26,56 +26,24 @@
 		</div>
 	</div>
 </template>
-<script lang="ts">
-import { System, useSystemApi } from "@data-system";
-import { computed, ComputedRef, defineComponent, onMounted, Ref, ref, watch } from "vue";
+<script setup lang="ts">
+import { useSystem } from "@data-system";
```
luejoh: consider placing it in modules/data/data-access. Martin introduced it to provide a container, in which we can put any migrated data composables. That frees us from defining a new name for every data composable. I like that approach, u2?

src/modules/page/class-members/ClassMembersInfoBox.vue
```diff
@@ -26,56 +26,24 @@
...
-	setup(props) {
-		const { t } = useI18n();
-		const { getSystem } = useSystemApi();
+const { systemId = "" } = defineProps<{ systemId?: string }>();
 
-		const system: Ref<System | undefined> = ref();
+const { t } = useI18n();
+// TODO: How to handle this?
+// eslint-disable-next-line vue/no-setup-props-reactivity-loss
+const { system, systemName } = useSystem(systemId);
```
luejoh: systemId is reactive, you can treat it as a reactive.
Question is, if the systemId may change during runtime. If that can happen,
you can teach the composable to watch the systemid and refetch on every change. The input variable is fetchSystem = (systemId: Ref<string>)=>
and can be easily created using
toRef(props, "systemId")

src/modules/page/class-members/ClassMembersInfoBox.vue
```diff
@@ -26,56 +26,24 @@
...
-
-		return {
-			hasSystem,
-			systemName,
-			infoBoxTextListItems,
-			t,
-		};
-	},
-});
+const hasSystem: ComputedRef<boolean> = computed(() => !!system.value);
```
luejoh: remove typeing?

luejoh: By the way. Are you a fan of !! operating ? I am not :) . You may consider using isNil from lodash or !== undefined ?

src/modules/ui/layout/topbar/UserMenu.vue
```diff
@@ -60,24 +60,21 @@ const props = defineProps({
 const { systemId } = useAppStoreRefs();
 
 const { t } = useI18n();
-const { getSystem } = useSystemApi();
+// TODO: How to handle undefined?
+const { system, systemName } = useSystem(systemId.value!);
```
luejoh: if input is a Ref<string>, you can just pass the ref towards the composable.

src/pages/user-login-migration/UserLoginMigrationConsent.page.vue
```diff
@@ -52,56 +52,35 @@
...
 
-		const migrationDescription: ComputedRef<string> = computed(() =>
-			userLoginMigration.value?.mandatorySince
-				? "pages.userMigration.description.firstParagraph.fromSourceMandatory"
-				: "pages.userMigration.description.firstParagraph.fromSource"
-		);
+const canSkipMigration: ComputedRef<boolean> = computed(() => !userLoginMigration.value?.mandatorySince);
 
-		const canSkipMigration: ComputedRef<boolean> = computed(() => !userLoginMigration.value?.mandatorySince);
+const isLoading: Ref<boolean> = ref(true);
```
luejoh: Remove typing?

src/pages/user-login-migration/UserLoginMigrationConsent.page.vue
```diff
@@ -52,56 +52,35 @@
...
+		? "pages.userMigration.description.firstParagraph.fromSourceMandatory"
+		: "pages.userMigration.description.firstParagraph.fromSource"
+);
 
-		const migrationDescription: ComputedRef<string> = computed(() =>
-			userLoginMigration.value?.mandatorySince
-				? "pages.userMigration.description.firstParagraph.fromSourceMandatory"
-				: "pages.userMigration.description.firstParagraph.fromSource"
-		);
+const canSkipMigration: ComputedRef<boolean> = computed(() => !userLoginMigration.value?.mandatorySince);
```
luejoh: remove typing?

src/pages/user-login-migration/UserLoginMigrationConsent.page.vue
```diff
@@ -52,56 +52,35 @@
...
-		const pageTitle = buildPageTitle(t("pages.userMigration.title"));
-		useTitle(pageTitle);
+const pageTitle = buildPageTitle(t("pages.userMigration.title"));
+useTitle(pageTitle);
 
-		const getSystemName = (): string =>
-			systemsModule?.getSystems.find(
-				(system: System): boolean => system.id === userLoginMigration.value?.targetSystemId
-			)?.name ?? "";
+const migrationDescription: ComputedRef<string> = computed(() =>
```
luejoh: remove typing?

src/pages/user-login-migration/UserLoginMigrationSuccess.page.vue
luejoh: You could consider using SvsSuspense here.

src/modules/page/class-members/ClassMembersInfoBox.vue
dyedwiper: Not a fan, but I don't have anything against it. Why do you?

luejoh: I think its less readable. In this particular case, its not, but combined with more conditions i often feel like its less well "telling", whats going on

src/modules/page/class-members/ClassMembersInfoBox.vue
dyedwiper: data + data-access seems one layer too much to me, dont you think?

src/modules/data/system/system.composable.ts
```diff
@@ -0,0 +1,40 @@
...
+import { $axios } from "@/utils/api";
+import { SystemsApiFactory } from "@api-server";
+import { computed, Ref, ref, watch } from "vue";
+
+export const useSystem = (systemId: Ref<string | undefined>) => {
+	const systemApi = SystemsApiFactory(undefined, "/v3", $axios);
+
+	const { execute } = useSafeAxiosTask();
+
+	const system = ref();
```
luejoh: type is properly inferred here?

src/modules/data/system/system.composable.unit.ts
```diff
@@ -0,0 +1,63 @@
...
+	});
+
+	it("fetches and sets system data when systemId is provided", async () => {
+		const systemId = ref("abc");
+		const mockSystem = publicSystemResponseFactory.build();
+		systemApiMock.systemControllerGetSystem.mockResolvedValueOnce(mockApiResponse({ data: mockSystem }));
+
+		const { system, systemName } = useSystem(systemId);
+		// TODO: Is it okay to wait for 4 ticks? Seems a little random
+		await nextTick();
```
luejoh: use await flushPromises() ?

src/modules/data/data-access/system.composable.unit.ts
```diff
@@ -0,0 +1,63 @@
...
+		// TODO: Is it okay to wait for 4 ticks? Seems a little random
+		await nextTick();
+		await nextTick();
+		await nextTick();
+		await nextTick();
+
+		expect(system.value).toEqual({
+			id: mockSystem.id,
+			displayName: mockSystem.displayName,
+			hasEndSessionEndpoint: !!mockSystem.oauthConfig?.endSessionEndpoint,
```
luejoh: dont like to much, that the code of finding out if hasEndSessionEndpoint has the proper value executes the same code as in the original code. Can we just test against "true" or "false"?

src/modules/page/class-members/ClassMembersInfoBox.unit.ts
```diff
@@ -27,16 +27,19 @@ describe("ClassMembersInfoBox", () => {
 		};
 	};
 
-	beforeEach(() => {
-		useSystemApiMock = mockComposable(useSystemApi);
-
-		vi.mocked(useSystemApi).mockReturnValue(useSystemApiMock);
+	const mockSystem = {
```
luejoh: use publicSystemResponseFactory.build(); ?

src/modules/ui/layout/topbar/UserMenu.vue
```diff
@@ -87,9 +83,6 @@ const isSessionTokenExpired = computed(
 );
 
 onMounted(async () => {
-	if (systemId.value) {
```
luejoh, line 90: That was actualyl the example we spoke about. If the systemId.value is undefined at given time, the system will never be refetched, once the value becomes available. Thats different now and is not relying on any order or time

src/pages/user-login-migration/UserLoginMigrationError.page.vue
```diff
@@ -44,78 +44,45 @@
...
+	}
+	return subject;
+};
 
-		onMounted(async () => {
-			await systemsModule?.fetchSystems();
-			await fetchLatestUserLoginMigrationForSchool();
-			isLoading.value = false;
-		});
+const supportLink: ComputedRef<string> = computed(() =>
```
luejoh: remove typings and infer them?

src/pages/user-login-migration/UserLoginMigrationSuccess.page.vue
```diff
@@ -18,55 +18,29 @@
...
-	props: {
-		targetSystem: {
-			type: String,
-			required: true,
-		},
-	},
-	setup() {
-		const systemsModule: SystemsModule = injectStrict(SYSTEMS_MODULE_KEY);
-		const { t } = useI18n();
+const props = defineProps<{ targetSystem: string }>();
```
luejoh: include "id" in the prop name?

luejoh (approved):

src/modules/data/data-access/system.composable.unit.ts
dyedwiper: Hm... I'm also thinking if we should include this logic here anyway or if it would be better to return the oauth config and let the component do the math. What do you think?

luejoh (approved):

PR #4177, opened 2026-04-20
BC-11512 - Migrate Vuex copyModule to composable

uidp:
# Short Description

We want to get rid of old Vuex modules in favor of much simpler composables. The `copyModule` is used widely in the project for duplicating and importing share links. So we expect the code to get much cleaner. 

BC-11512

## Checklist before merging


src/composables/copy-content.composable.ts
```diff
@@ -0,0 +1,144 @@
...
+	text: string;
+}
+
+export interface CopyContentTemplate {
+	text: string;
+	info: string;
+	warnings: CopyWarning[];
+}
+
+export const useCopyContent = (copyItemType: MaybeRef<ContentItemTypeEnum>) => {
```
luejoh: its not reacting on reactive data, so shall we only support primitive? Otherwise for me using the composable is supposed to do stuff reactively

src/composables/copy-content.composable.ts
```diff
@@ -0,0 +1,144 @@
...
+	const { t } = useI18n();
+
+	const copyContent = computed<CopyContentTemplate>(() => {
```
luejoh: Is it inferred correctly without typing?

src/composables/copy-content.composable.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+export interface CopyContentTemplate {
+	text: string;
+	info: string;
+	warnings: CopyWarning[];
+}
+
+export const useCopyContent = (copyItemType: MaybeRef<ContentItemTypeEnum>) => {
+	const { t } = useI18n();
```
luejoh: do we potentially should useI18nGlobal() ? Since from just looking at the file, i would not be sure, if we r in a setup context, in which the useI18n only works in.

src/composables/copy-content.composable.ts
```diff
@@ -0,0 +1,144 @@
...
+
+	const copyContent = computed<CopyContentTemplate>(() => {
+		const templates: Record<ContentItemTypeEnum, CopyContentTemplate> = {
```
luejoh: this table is actually a constant from which the useCopyContent just need to choose from.
Right now, its redeclared on every composable call.
Can we instead put it in a constant on top of the file?
Would also make it more clear, that the composable just do a pick from a template

src/composables/copy-content.composable.ts
```diff
@@ -0,0 +1,144 @@
...
+				],
+			},
+			[ContentItemTypeEnum.Unknown]: {
+				text: "",
+				info: "",
+				warnings: [],
+			},
+		};
+
+		return templates[unref(copyItemType) || ContentItemTypeEnum.Unknown];
```
luejoh: Ok its actually reactive, right? Using unref way to less to be certain XD

luejoh: Tell me about your use cases

src/modules/data/application/loading.store.ts
```diff
@@ -14,9 +14,25 @@ export const useLoadingStore = defineStore("loadingStore", () => {
 		isLoading.value = value;
 	}
 
+	const withLoadingState = async <T>(
+		fn: () => Promise<T>,
+		loadingMessage: string,
+		minDisplayTime = 300
+	): Promise<T> => {
+		setLoadingState(true, loadingMessage);
+		try {
```
luejoh: useTryCatch util?

src/modules/data/application/loading.store.ts
```diff
@@ -14,9 +14,25 @@ export const useLoadingStore = defineStore("loadingStore", () => {
...
+			const result = await fn();
+			return result;
+		} finally {
+			await new Promise((resolve) => setTimeout(resolve, minDisplayTime));
```
luejoh: Hm is a minimum waiting time necessary? I like it more to only display loading times, once they exceed 200ms. (e.g. in svssuspense)

src/modules/feature/board/board/Board.vue
```diff
@@ -399,18 +387,17 @@ const boardColumnClass = computed(() => {
 	return classes;
 });
 
-const onCopyResultModalClosed = () => {
-	copyModule.reset();
-};
+const copyFlow = useCopyFlow();
```
luejoh: desruct right here?

src/modules/feature/board/board/Board.vue
```diff
@@ -399,18 +387,17 @@ const boardColumnClass = computed(() => {
...
+const { isDialogOpen: isCopyDialogOpen, copyItemType, onConfirm: onConfirmCopy, onCancel: onCancelCopy } = copyFlow;
```
luejoh: I wonder that useCopyFlow does not ask for the copyItemType, but indirectly decides, which one to copy on confirm action. (+defaulting to "Course" on execution)

luejoh: That actually would also allow the composable to decide already, which "executor" is needed and handles internally, which one to offer? But ye from a typing perspective, which executor needs potentially different data, maybe a functional approach would still add value?

e.g.
useBoardCopyFlow();

const useBoardCopyFlow = ()=> {
const {executeBoardCopy, ...} = useCopyFlow(ContentItemTypeEnum.Board);

return {execute: executeBoardCopy, ...}
}

src/modules/feature/copy/copy-flow.composable.ts
```diff
@@ -0,0 +1,142 @@
+import { useSafeAxiosTask } from "@/composables/async-tasks.composable";
+import { useAwaitableAction } from "@/composables/awaitable-action.composable";
+import { ContentItemTypeEnum } from "@/types/enum/content-item-type.enum";
+import { $axios } from "@/utils/api";
+import { BoardApiFactory, CourseRoomsApiFactory, RoomApiFactory, TaskApiFactory } from "@api-server";
+import { notifySuccess, useLoadingStore } from "@data-app";
+import { ref } from "vue";
+import { useI18n } from "vue-i18n";
+
+export const useCopyFlow = () => {
```
luejoh, line 10: Ive was starting some comments on the board.vue regarding how to use the composable.

I think "coupling together" the different user stories into own functions, which internally use the "useCopyFlow", does the trick in making it more beautiful.

const useCopyFlow = () // all the identical things, loadingstate, dialogstate, awaiting action props
const useBoardCopy = (copyItemType) uses useCopyFlow and "enhances" it with the differences (different executor, different api)

src/modules/feature/copy/copy-flow.composable.unit.ts
```diff
@@ -0,0 +1,206 @@
...
+		roomApi = mockApi<serverApi.RoomApiInterface>();
+
+		vi.spyOn(serverApi, "CourseRoomsApiFactory").mockReturnValue(courseRoomsApi);
+		vi.spyOn(serverApi, "TaskApiFactory").mockReturnValue(taskApi);
+		vi.spyOn(serverApi, "BoardApiFactory").mockReturnValue(boardApi);
+		vi.spyOn(serverApi, "RoomApiFactory").mockReturnValue(roomApi);
+
+		const loadingStore = mockedPiniaStoreTyping(useLoadingStore);
+		mockWithLoadingState = vi.fn().mockImplementation(async (fn) => fn());
+		vi.spyOn(loadingStore, "withLoadingState").mockImplementation(mockWithLoadingState);
```
luejoh: is it necessary to mock this composable?

src/modules/feature/copy/copy-flow.composable.unit.ts
```diff
@@ -0,0 +1,206 @@
...
+import { Mocked } from "vitest";
+
+let courseRoomsApi: Mocked<serverApi.CourseRoomsApiInterface>;
+let taskApi: Mocked<serverApi.TaskApiInterface>;
+let boardApi: Mocked<serverApi.BoardApiInterface>;
+let roomApi: Mocked<serverApi.RoomApiInterface>;
+
+let mockWithLoadingState: ReturnType<typeof vi.fn>;
+
+const mountCopyFlowComposable = () =>
```
luejoh: file is wip

src/modules/feature/copy/CopyDialog.unit.ts
```diff
@@ -0,0 +1,37 @@
...
+
+		const wrapper = mount(CopyDialog, {
+			props: {
+				isOpen: true,
+				copyItemType: ContentItemTypeEnum.Room,
+			},
+			global: {
+				plugins: [createTestingVuetify(), createTestingI18n()],
+			},
+			attachTo: document.body,
```
luejoh, line 24: attachment is necessary? Just seen to often, dass its done, when its not needed

src/modules/feature/copy/CopyDialog.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+	<SvsDialog
+		:model-value="isOpen"
+		:is-open-state-managed-externally="true"
```
luejoh: u can also just write the attribute name, which is identical to that.

e.g. no-cancel, also just works without a value

src/modules/feature/copy/CopyDialog.vue
```diff
@@ -0,0 +1,63 @@
...
+import { InfoAlert, WarningAlert } from "@ui-alert";
+import { SvsDialog } from "@ui-dialog";
+import { computed, toRef } from "vue";
+import { useI18n } from "vue-i18n";
+
+const { t } = useI18n();
+
+const props = withDefaults(
+	defineProps<{
+		isOpen?: boolean;
```
luejoh: define as defineModel and remove the prop?

src/modules/feature/copy/CopyDialog.vue
luejoh: Is it necessary to manage that externally? onConfirm and onCancel is still being send, but the dialogs state is managed by the svsdialog. Maybe its necessary, to have the below mentioned v-model. It is likely, that dialog state setting to false in oncancel of useCopyFlow is not necessary anymore, but could still be done.

It would just mean, that two parties are allowed to change the same ref, which sounds ok to me. To you too?

Thats at least sth. ive seen a couple of times on the existiting dialogs, that cancel is sometimes then just not bound on

src/modules/feature/import/import-flow.composable.ts
```diff
@@ -0,0 +1,97 @@
...
+export const useImportFlow = () => {
+	const shareApi = ShareTokenApiFactory(undefined, "/v3", $axios);
+	const { execute } = useSafeAxiosTask();
+	const importAction = useAwaitableAction<{ newName: string; destination?: ImportDestination }>();
+	const { withLoadingState } = useLoadingStore();
+	const { t } = useI18n();
+
+	const shareTokenInfo = ref<ShareTokenInfoResponse>();
+
+	const isGenericImportDialogOpen = computed(
```
luejoh: maybe split it for better readability?

`
const isImportActive = computed(
    () => importAction.isActive.value && !!shareTokenInfo.value
);

const isCardImport = computed(
    () => shareTokenInfo.value?.parentType === ShareTokenInfoResponseParentType.CARD
);

const isGenericImportDialogOpen = computed(() => isImportActive.value && !isCardImport.value);
const isCardImportDialogOpen    = computed(() => isImportActive.value &&  isCardImport.value);
`

src/modules/feature/import/import-flow.composable.ts
```diff
@@ -0,0 +1,97 @@
...
+
+	const executeImport = async (token: string) => {
+		const { result: validationResult } = await validateShareToken(token);
+
+		if (!validationResult) {
+			return { success: false, error: new Error("Validation failed") };
+		}
+		shareTokenInfo.value = validationResult;
+
+		const { completed, data } = await importAction.start();
```
luejoh: I struggled a bit in understanding what happens here and i think just renaming importAction.start does the trick in making it easier.

The importAction is used to reflect the async action of asking the user for confirming the import action.

Maybe sth. like await "askUserToImport" or similar?

luejoh: await tokenvalidation
await userConfirmation
await importShareToken

src/modules/feature/import/import-flow.composable.ts
```diff
@@ -0,0 +1,97 @@
...
+		if (!completed) return { success: false, error: new Error("Import cancelled") };
+
+		const { result, success, error } = await withLoadingState(
+			() => importShareToken(validationResult, { newName: data.newName, destinationId: data.destination?.id }),
+			t("components.molecules.import.options.loadingMessage")
+		);
+
+		if (success) {
```
luejoh, line 64: I liked that each of the async tasks were raising an own error message to be returned on failure.

That is a very good example in which multiple things can fail during one proccess and for each "error case", we can return an appropriate error type (in this case message is sufficient). Shall we have a third?
On top we could also "enhance" the error returned from async functions with the cause property ? This allows to retain the original error and tell more about it. What do you think?

src/modules/feature/import/ImportDialog.vue
```diff
@@ -0,0 +1,206 @@
+<template>
+	<SvsDialog
+		v-model="isDialogOpen"
+		:is-open-state-managed-externally="true"
```
luejoh: similar question as in copy dialog. Furthermore importcarddialog and this one uses defineModel. Do we really need externally managed?

src/modules/feature/import/ImportDialog.vue
```diff
@@ -0,0 +1,206 @@
...
+	activeStep.value = hasSelectStep.value ? "select" : "rename";
+	selectedDestinationId.value = undefined;
+	nameInput.value = undefined;
+};
+
+onMounted(() => {
+	resetDialog();
+});
+
+watch(isDialogOpen, (isOpen) => {
```
luejoh: hm can we instead reset the dialog on "after-leave" from svsdialog?
I guess we neither need to watch and onMounted then?

src/modules/feature/import/ImportDialog.vue
```diff
@@ -0,0 +1,206 @@
...
+	if (isOpen) {
+		resetDialog();
+	}
+});
+
+const rules = reactive({
+	required: (value: string) => !!value || t("common.validation.required"),
```
luejoh: use isRequired validation util?

src/modules/feature/import/ImportDialog.vue
```diff
@@ -0,0 +1,206 @@
...
+	validateOnOpeningTag: (value: string) => validateOnOpeningTag(value),
```
luejoh: use useOpeningTagValidator validation util?

src/modules/feature/import/ImportDialog.vue
```diff
@@ -0,0 +1,206 @@
...
+
+const hasSelectStep = computed(
+	() =>
+		props.shareTokenInfo.parentType === ShareTokenInfoResponseParentType.LESSONS ||
+		props.shareTokenInfo.parentType === ShareTokenInfoResponseParentType.TASKS ||
+		props.shareTokenInfo.parentType === ShareTokenInfoResponseParentType.COLUMN_BOARD
+);
+
+const confirmBtnLangKey = computed(() => {
+	if (activeStep.value === "select") {
```
luejoh, line 171: ternary?

src/modules/page/room/Rooms.page.vue
```diff
@@ -65,48 +75,47 @@ const fabAction = computed(() => {
...
+const executeImportFlow = async (token: string) => {
+	const { result: importResult } = await executeImport(token);
+
+	if (!importResult) {
+		router.push({ name: "rooms" });
+		return;
+	}
+
+	if (importResult.destination && importResult.destination.type === "room") {
+		router.replace({ name: "room-details", params: { id: importResult.destination.id } });
```
luejoh, line 97: hm router.push instead?

src/modules/page/room/Rooms.page.vue
```diff
@@ -65,48 +75,47 @@ const fabAction = computed(() => {
...
+	} else {
+		router.replace({ name: "rooms" });
+		fetchRooms();
+	}
+};
 
 watch(
 	() => route.query.import,
 	() => {
 		if (route.query.import) {
```
luejoh: use the newValue passed to the callback instead of the reactive prop?
Thats at least the common practice when using watch

src/modules/page/room/Rooms.page.vue
```diff
@@ -65,48 +75,47 @@ const fabAction = computed(() => {
...
+	} else if (importResult.destination && importResult.destination.type === "column") {
+		router.replace({ name: "boards-id", params: { id: importResult.destination.boardId } });
+	} else {
+		router.replace({ name: "rooms" });
+		fetchRooms();
+	}
+};
 
 watch(
 	() => route.query.import,
```
luejoh, line 107: could you check if vue-use "useUrlSearchParams" is a better choice here?

src/pages/course-rooms/CourseRoomOverview.page.vue
```diff
@@ -99,12 +99,14 @@
...
 		@drag-from-group="dragFromGroup"
 	/>
-	<ImportFlow
-		:is-active="isImportMode"
-		:token="importToken"
-		:destinations="courses.filter((course) => !course.isLocked)"
-		:destination-type="BoardExternalReferenceType.COURSE"
-		@success="onImportSuccess"
+	<ImportDialog
+		v-if="isGenericImportDialogOpen"
```
luejoh: the v-if is necessary? Maybe its not anymore, if resetting state is done on "after-leave" (referring to another comment)

src/pages/course-rooms/CourseRoomOverview.page.vue
```diff
@@ -185,21 +188,46 @@ const rooms = computed(() =>
...
+	} else {
+		router.replace({ name: "course-room-overview" });
+		fetchCourses();
+	}
+};
+
+watch(
+	() => route.query.import,
+	() => {
+		if (route.query.import) {
```
luejoh: similar comments as in courseRoomDetails.page

src/utils/content-item.utils.ts
```diff
@@ -0,0 +1,38 @@
+import { ContentItemTypeEnum } from "@/types/enum/content-item-type.enum";
+import { ShareTokenInfoResponseParentType } from "@api-server";
+
+export const mapShareTokenParentTypeToContentItemType = (shareTokenParentType: ShareTokenInfoResponseParentType) => {
```
luejoh, line 4: Hm if we already have an enum from the server, do we really need an own additional enum?

src/composables/copy-content.composable.ts
uidp: It is. Removed.

src/composables/copy-content.composable.ts
uidp: Good point. Changed.

src/modules/data/application/loading.store.ts
uidp: tried but doesn't make the code easier to read. maybe because of the `try...finally`.

src/modules/feature/board/board/Board.vue
uidp: Right. Changed.

src/modules/feature/copy/copy-flow.composable.unit.ts
uidp: No, not the entire composable. I changed it to just the mock for `withLoadingState` to directly call and return.

src/modules/feature/copy/CopyDialog.unit.ts
uidp: I found no other way to access dialog contents. Any ideas?

src/modules/feature/copy/CopyDialog.vue
uidp: Good point. Changed.

src/modules/feature/copy/CopyDialog.vue
uidp: Done.

src/modules/page/room/Rooms.page.vue
uidp: but that would mean that the user can go back to the import url? i would prefer "losing" the import url fom the history after successful import. what do you think?

src/modules/feature/copy/CopyDialog.vue
uidp: Having the opening state externally managed leaves the decision when the dialog should be actually closed to the flow composable. In the `ImportDialog` e.g. we have potentially more than one step and the user advances the steps by hitting "confirm". This leads to the dialog being closed too early. This is not always the case but I think having the flow manage the dialog state is generally a good choice.

src/modules/feature/import/import-flow.composable.ts
uidp: Good idea. Changed.

src/modules/feature/import/import-flow.composable.ts
uidp: I'm not sure what "Shall we have a third?" means. Do you mean adding a third property to `AwaitableResult` for a possible error?

uidp: Good idea setting the cause of the token validaion error. I added that.

src/modules/feature/import/ImportDialog.vue
uidp: I would vote for generally leave the state managed externally (see other comment).

src/modules/feature/import/ImportDialog.vue
uidp: Fixed.

src/modules/feature/import/ImportDialog.vue
uidp: Done.

src/modules/feature/import/ImportDialog.vue
uidp: Done.

src/modules/feature/import/ImportDialog.vue
uidp: Done.

src/modules/page/room/Rooms.page.vue
uidp: Fixed. Also in `CourseRoomOverview.page.vue`.

src/modules/page/room/Rooms.page.vue
uidp: That would still require the watcher right? I cannot see any advantage when reading `params.token` instead of `route.query.import`.

src/pages/course-rooms/CourseRoomOverview.page.vue
uidp: ~Actually the reset (or re-initialization) in `after-leave` did not work because we need to initialize the current step based on the import type when the dialog opens. As the dialog can be opened by mounting or setting `isOpen = true` I think we need both options: `onMounted` and `watch` on `isOpen`.
The `v-if` might be left out though.~

src/pages/course-rooms/CourseRoomOverview.page.vue
uidp: Fixed.

src/utils/content-item.utils.ts
uidp: `ContentItemTypeEnum` was an attempt to have something more general which can be used for duplication, sharing and importing. `ShareTokenInfoResponseParentType` is just the share token type from the share api response. Maybe we should have provided something more general in the API.

src/modules/feature/import/import-flow.composable.ts
uidp: I changed the name to `importDialogAction`. hope that's okay. I really would like to keep the `Action` suffix to make clear that it's a pending action.

luejoh (approved):

PR #4182, opened 2026-04-23
BC-11632 Use newest version of hpi-schul-cloud/ckeditor

dyedwiper:
# Short Description
Includes some unrelated changes:
- A problem with JSON formatting came to light that stemmed from putting the prettier-config inside the eslint-config. Thus the prettier-config is put back in `.prettierrc` which is also recommended for eslint-plugin-prettier: https://github.com/prettier/eslint-plugin-prettier#options
- Some of the values in the prettier-config equaled the defaults and are removed. See: https://prettier.io/docs/options
- Also the specification of the typescript workspace version in VSCode settings was deprecated and seems unnecessary, see: https://code.visualstudio.com/docs/typescript/typescript-transpiling#_using-the-workspace-version-of-typescript

BC-11632

## Checklist before merging


.vscode/settings.default.json
```diff
@@ -1,20 +1,21 @@
...
+	},
+	"[typescript]": {
+		"editor.defaultFormatter": "dbaeumer.vscode-eslint"
+	},
+	"[json]": {
+		"editor.defaultFormatter": "esbenp.prettier-vscode",
+		"prettier.useTabs": true
+	},
+	"editor.formatOnSave": true,
+	"eslint.format.enable": true
```
virgilchiriac, line 18: as an observation since I use a diferent editor, to me makes little sense to see this config in the repo
this means I could save with diferent settings

instead, each editor should use defaults / as configured in the project root

package-lock.json
```diff
@@ -702,7 +702,7 @@
 			"version": "2.6.2",
 			"resolved": "https://registry.npmjs.org/@bufbuild/protobuf/-/protobuf-2.6.2.tgz",
 			"integrity": "sha512-vLu7SRY84CV/Dd+NUdgtidn2hS5hSMUC1vDBY0VcviTdgRYkU43vIz3vIFbmx14cX1r+mM7WjzE5Fl1fGEM0RQ==",
-			"dev": true,
+			"devOptional": true,
```
virgilchiriac, line 705: why has this change? I supposed you did run it on npm 11, but why did it change?

dyedwiper: Yes, I ran it on npm 11.11.0. I still haven't understood why this happens...

.vscode/settings.default.json
dyedwiper: Do you mean "eslint.format.enable" or "prettier.useTabs"?

virgilchiriac: prettier.useTabs 
if I use prettier in my editor, it might have other setting
not sure if eslint.format.enable is editor setting or should be eslint configuration

virgilchiriac (approved):

.vscode/settings.default.json
dyedwiper: Yes, this shouldn't rely on the editor. I pulled out the prettier-config to .prettierrc again.

dyedwiper: In eslint-config the eslint-prettier-plugin is enabled, but for it to work in VSCode `eslint.format.enable` needs to be enabled too. I don't know about other editors.

eslint.config.js
```diff
@@ -110,22 +110,6 @@ export default defineConfigWithVueTs([
 			"no-useless-escape": "error",
 			"no-var": "error",
 			"prefer-const": "error",
-			"prettier/prettier": [
-				"error",
```
virgilchiriac, line 114: ok, cool, so we only specify what is not already as default in prettier.
But unsure how this works now with eslint.  Where / how is now specify this now? 
KI explains: The "error" is the severity level for the prettier/prettier rule. It means ESLint will report Prettier formatting violations as errors (exit code 1), causing the linting process to fail.

package.json
virgilchiriac, line 71: ~do we actually need these 2~ ?
nevermind, we do

package-lock.json
virgilchiriac: for the records, happens when running npm install

virgilchiriac (approved):

package.json
dyedwiper: Yes, because eslint should still format vue, ts and js files. That was Johannes' ides some time ago and I didn't want to change that. Now the plugin just reads the prettier-config from prettierrc instead of eslint-config, which is also recommended. See the docs: https://github.com/prettier/eslint-plugin-prettier#options

eslint.config.js
dyedwiper: It still does. It just reads the prettier-config from prettierrc now. See my other answer also.

package.json
dyedwiper: ...did not see your edit

PR #4184, opened 2026-04-23
BC-11358 - duplicate board column

virgilchiriac:
# Short Description

BC-11358
https://github.com/hpi-schul-cloud/schulcloud-server/pull/6255

## Screenshots of UI changes
Bildschirmfoto 2026-04-23 um 15 32 13

## Checklist before merging


src/modules/data/board/Board.store.ts
```diff
@@ -148,6 +149,49 @@ export const useBoardStore = defineStore("boardStore", () => {
...
+		);
+
+	const duplicateColumnSuccess = (payload: DuplicateColumnSuccessPayload) => {
+		if (!board.value) return;
+
+		const { columnId, duplicatedColumn } = payload;
+		const columnIndex = getColumnIndex(columnId);
+		if (columnIndex < 0) return;
+
+		if (!duplicatedColumn.id) return;
```
luejoh: can that ever happen?

src/modules/data/board/Board.store.ts
```diff
@@ -148,6 +149,49 @@ export const useBoardStore = defineStore("boardStore", () => {
...
+
+		const duplicatedColumnSkeleton: ColumnResponse = {
+			...duplicatedColumn,
+			cards: duplicatedColumn.cards.map(
```
luejoh, line 175: ^why is mapping necessary here?

src/modules/feature/board/board/BoardColumnHeader.vue
```diff
@@ -44,13 +49,15 @@
 </template>
 
 <script setup lang="ts">
+import { useSafeTaskRunner } from "../../../../composables/async-tasks.composable";
```
luejoh: shorthand ? form @utils`?

src/modules/feature/board/board/BoardColumnHeader.vue
```diff
@@ -27,6 +27,11 @@
 			<div class="mt-2 mr-3">
 				<BoardMenu v-if="canDeleteColumn" :scope="BoardMenuScope.COLUMN" :data-testid="`column-menu-btn-${index}`">
 					<KebabMenuActionRename v-if="!isEditMode" @click="onStartEditMode" />
+					<KebabMenuActionDuplicate
+						v-if="allowedOperations?.copyColumn"
+						data-testid="kebab-menu-action-duplicate-column"
+						@click="duplicateColumn"
```
luejoh, line 33: hm all the other chainges emit to parent. Why you did not?

src/modules/feature/board/board/BoardColumnHeader.vue
```diff
@@ -139,6 +150,10 @@ const onMoveColumnUp = () => emit("move:column-up");
 
 const onUpdateTitle = (newTitle: string) => (updatedTitle.value = newTitle);
 
+const { run: duplicateColumn } = useSafeTaskRunner(async () => {
```
luejoh, line 153: we dont wanna show any "transmitting state"? Duplicate card e.g. was showing a skeleton while duplicating.

src/modules/data/board/Board.store.ts
virgilchiriac: It is a bit sketchy but yes
duplicateColumn is of type ColumnFullResponse but to add it to board we need type ColumnResponse, which is what do we here...

src/modules/data/board/Board.store.ts
virgilchiriac: actually, not. removed

src/modules/feature/board/board/BoardColumnHeader.vue
virgilchiriac: it's how it is done on card duplicate too

src/modules/data/board/Board.store.ts
```diff
@@ -148,6 +149,47 @@ export const useBoardStore = defineStore("boardStore", () => {
...
+					cardId: card.id,
+					height: card.height,
+				})
+			),
+		};
+
+		board.value.columns?.splice(columnIndex + 1, 0, duplicatedColumnSkeleton);
+
+		if (payload.isOwnAction === true) {
+			if (hasRelevantContentForDuplicationWarning(duplicatedColumn)) {
```
dyedwiper, line 187: If I undestand correctly, this check is the only reason for the new ColumnFullResponse. I think that is a lot of overhead. Can't we just always display the info? 

I see there is a similar check when duplicating cards, but in the ticket it just says that the info should be displayed without mentioning a condition. To me it seems like okay UX to just always display it.

virgilchiriac: no, we show success message otherwise
yes, there is overhead, including in BE...

src/locales/de.ts
```diff
@@ -489,6 +489,9 @@ export default {
 	"components.board.notifications.errors.notUpdated": "Die Änderungen konnten nicht gespeichert werden.",
 	"components.board.notifications.info.cardDuplicated":
 		"Inhalte aus Etherpads und Whiteboards sowie geschützte Einstellungen externer Tools werden nicht übernommen.",
+	"components.board.notifications.info.columnDuplicated":
```
dyedwiper: I'm thinking if a new key is good here or if "components.board.notifications.info.cardDuplicated" should be used (maybe with another name). I tend to say the latter because it refers to exactly the same scenario. When using two different keys, it seems like the values are only the same by accident and could differ in the future.

virgilchiriac: I considered it, but I think for translation is better to have separated keys. 
 
it should be a componded warning or info message, depending on type of content.
so for example if we disable Whiteboard, this does not show up in the message. With the refactoring on dialogs, something like this is supported.

src/modules/data/board/Board.store.ts
dyedwiper: True, I've overseen that it is an if-else because I've expected the success message to be always shown. Seems wrong to me that it is only conditionally shown because the content-warning is only an info and does not impact the success of the copy. I would just always show both notifications on success - or better one notification with both texts. Have you talked about it with PO/UX?

src/locales/de.ts
dyedwiper: But if we disable Whiteboards, you surely have to change both translations in the same way. That's why I mean there should only be one.

dyedwiper: I think the decisive question is: can you think of any scenario where the translations have to change independently?

virgilchiriac:
see also followup https://github.com/hpi-schul-cloud/nuxt-client/pull/4198

src/locales/de.ts
virgilchiriac: was updated - makes sense to use same text
Note:  Was thinking to rename cardDuplicated to cardOrColumnDuplicated  but to avoid too many file changes, I have not change the label

src/modules/data/board/Board.store.ts
virgilchiriac: was done

src/modules/feature/board/board/BoardColumnHeader.vue
virgilchiriac: for the record, was decided to skip it for now - part of  #4198

luejoh (approved):

PR #4198, opened 2026-04-30
BC-11665 - show loading state on column duplication

virgilchiriac:
# Short Description

BC-11665

## Checklist before merging


src/modules/data/board/Board.store.ts
```diff
@@ -101,6 +102,17 @@ export const useBoardStore = defineStore("boardStore", () => {
 		isLoading.value = loading;
 	};
 
+	const setDuplicatingColumnId = (columnId: string, isDuplicating = true) => {
+		if (columnId === undefined) return;
```
uidp: Is this even possible? At least at compile time not right? Is there any specific requirement for the runtime check?

virgilchiriac: ah, good catch. that's a leftover forgotten to remove

src/modules/data/board/Board.store.ts
```diff
@@ -101,6 +102,17 @@ export const useBoardStore = defineStore("boardStore", () => {
...
+		if (isDuplicating) {
+			duplicatingColumnIds.value.add(columnId);
+		} else {
+			duplicatingColumnIds.value.delete(columnId);
+		}
+	};
```
uidp, line 111: I'm wondering if it wouldn't be better to have a central `loadingStore` like (example):
```ts
import { defineStore } from 'pinia';
import { ref, computed } from 'vue';

export const useLoadingStore = defineStore('loading', () => {
  // Store active "Task-Keys" or IDs
  const activeLoadings = ref(new Set());

  // Getter: Anything running at all?
  const isLoading = computed(() => activeLoadings.value.size > 0);

  // Getter: Is a specific task running?
  const isLoadingTask = (taskKey) => computed(() => activeLoadings.value.has(taskKey));

  function start(taskKey = 'global') {
    activeLoadings.value.add(taskKey);
  }

  function stop(taskKey = 'global') {
    activeLoadings.value.delete(taskKey);
  }

  return { isLoading, isLoadingTask, start, stop };
});
```

virgilchiriac: I would consider this out of scope, better for a followup / refactor

MartinSchuhmacher:
decided against the change for now -> closed

PR #4199, opened 2026-04-30
BC-11490 - refactor auto-logout

hoeppner-dataport:
# Short Description
Refactoring of autologout 

BC-11490
https://github.com/hpi-schul-cloud/schulcloud-client/pull/3814
https://github.com/hpi-schul-cloud/dof_app_deploy/pull/1399

## Checklist before merging


src/modules/data/application/application.store.ts
```diff
@@ -1,10 +1,11 @@
+import { notifySuccess } from "./notification-store";
```
muratmerdoglu-dp, line 1: Some unit tests are needed for this refactoring.

muratmerdoglu-dp (approved):

src/modules/data/application/application.store.ts
hoeppner-dataport: fixed

src/modules/data/application/application.store.ts
```diff
@@ -55,26 +67,50 @@ export const useAppStore = defineStore("applicationStore", () => {
...
+			const result = await $axios.post(JWT_TIMER_ENDPOINT);
+			setTimerSeconds(result.data.ttl);
+			notifySuccess("feature-autoLogout.message.extending-session-success");
+			postTimerSeconds(result.data.ttl);
+		} catch {
+			logoutWithoutBroadcast("/logout?auto-logout=true");
+		}
+	};
+
+	// TODO: remove after testing
```
sebmue-dataport: 🦄

src/modules/data/application/application.store.ts
```diff
@@ -16,13 +17,24 @@ const setCookie = (cname: string, cvalue: string, exdays: number) => {
 	document.cookie = `${cname}=${cvalue}; Expires=${expires}; Path=/; Secure; SameSite=None`;
 };
 
+const DEFAULT_TIMEOUT_SECONDS = 2 * 60 * 60; // 2 hours
```
sebmue-dataport, line 20: In a way I wonder whether such values should be defined in a more global scope, although DOF seems like overkill here. And it would allow for instances to have different settings leading to more confusion when trying to reproduce bugs. So maybe it still is best to leave them here?

src/modules/data/application/application.store.ts
hoeppner-dataport: Testing is not over yet :)

src/modules/data/application/application.store.ts
hoeppner-dataport: It's only the default value. The actual value is taken from dof-configuration - so from env-config.
But fun fact aside: the static text string says: "Achtung: Du wirst in {0} automatisch ausgeloggt. Verlängere jetzt deine Sitzung auf zwei Stunden." :)

src/modules/data/application/application.store.ts
hoeppner-dataport: removed

hoeppner-dataport: removed now

PR #4200, opened 2026-05-04
BC-11663 - update release preferences after changing them

luejoh:
# Short Description

BC-11663

## Checklist before merging


src/modules/data/application/application.store.ts
```diff
@@ -75,6 +84,19 @@ export const useAppStore = defineStore("applicationStore", () => {
...
 
+	const updateUserPreferences = async (options: UpdatePreferencesBodyParams) => {
+		const { execute: executePreferences } = useSafeAxiosTask();
+
+		const { success, result } = await executePreferences(async () => {
+			await meApi.meControllerUpdateMePreferences(options);
+			return meApi.meControllerMe();
+		}, "common.notification.error.preferences");
+
+		if (success && meResponse.value?.preferences) {
```
dyedwiper: Why only set them when there are existing preferences?

luejoh: meReponse is actually always given as it is the first thing being requested, but written into a ref, which is undefined at boot time. assigning a value to a potentially "undefined" path is complained by typescript.

meResponse.value?.preferences = // sth 
Is raising:

TS2779: The left-hand side of an assignment expression may not be an optional property access.

dyedwiper: But shouldn't it be `if (success && meResponse.value)`, so that new preferences are assigned even if they were undefined before?

luejoh: true, fixed

src/locales/en.ts
```diff
@@ -154,6 +154,7 @@ export default {
 	"common.notifications.errors.notFinished": "{type} could not be completed.",
 	"common.notifications.errors.notReverted": "{type} could not be reset",
 	"common.notifications.errors.notRestored": "{type} could not be restored.",
+	"common.notification.error.preferences": "User preferences could not be retrieved.",
```
dyedwiper: The call where this is displayed is an update. Shouldn't it then be "User preferences could not be updated."?

src/modules/data/application/application.store.ts
```diff
@@ -75,6 +84,19 @@ export const useAppStore = defineStore("applicationStore", () => {
 
 	const externalLogout = () => logout("/logout/external");
 
+	const updateUserPreferences = async (options: UpdatePreferencesBodyParams) => {
```
MartinSchuhmacher, line 117: is it still an "update" even if the preferences where undefined before?

src/locales/de.ts
```diff
@@ -146,6 +146,7 @@ export default {
 	"common.medium.information.teacher": "Bitte an Schuladministrator:in wenden.",
 	"common.nodata": "Keine Daten vorhanden",
 	"common.notification.error": "Es ist ein Fehler aufgetreten.",
+	"common.notification.error.preferences": "Nutzereinstellungen konnten nicht abgerufen werden.",
```
MartinSchuhmacher: I agree to @dyedwiper  comment, if we say the method is "updating" something we should change the message. Or determine that we generalize the method naming to `set{sth}`, but even then we might name the message to saying "Nutzereinstellungen konnten nicht gesetzt werden" or similar.

src/modules/data/application/application.store.ts
luejoh: yes, especially if they were undefined. He did not had any and now he will have

MartinSchuhmacher (approved):

PR #4211, opened 2026-05-08
BC-11521 - Migrate Vuex ShareModule to composable

uidp:
# Short Description
We want to get rid of old Vuex modules in favor of much simpler composables. The shareModule is used widely in the project for sharing content items. So we expect the code to get much cleaner. We also want to make use of new concepts like async task and awaitable action - similar to the `copyFlow` and `importFlow` composables.

- BC-11521
- BC-11512

## Checklist before merging


src/components/course-rooms/CourseRoomDashboard.vue
```diff
@@ -274,31 +269,25 @@ const boardLayoutAriaLabel = (itemLayout: BoardLayout) => {
...
 };
 
-const getSharedBoard = (boardId: string) => {
-	if (useEnvConfig().value.FEATURE_COLUMN_BOARD_SHARE) {
-		shareModule.startShareFlow({
-			id: boardId,
-			type: ShareTokenBodyParamsParentType.COLUMN_BOARD,
-		});
-	}
+const shareBoard = (boardId: string) => {
```
luejoh: maybe one function, which receives the target type is somewhat shorter?

src/composables/async-tasks.composable.ts
```diff
@@ -55,9 +56,14 @@ export const useSafeTask = () => {
 export const useSafeAxiosTask = () => {
 	const { execute: safeExec, isRunning, reset, status, error } = useSafeTask();
 	const { t } = useI18nGlobal();
+	const isLoading = ref(false);
+	const isBlocked = computed(() => isRunning.value || isLoading.value);
```
luejoh, line 60: can we add a comment, so it is shown, what the isBlocked is reflecting. It would show up when its hovered by devs i think

src/composables/confirmation-dialog.composable.ts
```diff
@@ -19,33 +20,21 @@ export interface ConfirmationOptions {
  * This composable is only meant to be used in App.vue where the dialog is rendered.
  */
 export const useInternalConfirmationDialog = createSharedComposable(() => {
-	let resolvePromise: ((value: boolean) => void) | undefined = undefined;
+	const action = useAwaitableAction<boolean>();
```
luejoh: can we give that action a name? So its reflecting, whats acutally happening while the action is executed?

src/composables/copy-content.composable.ts
```diff
@@ -13,10 +14,19 @@ export interface CopyContentTemplate {
 	warnings: CopyWarning[];
 }
 
-export const useCopyContent = (copyItemType: Ref<ContentItemTypeEnum>) => {
+const apiTypeToEntityType: Record<string, ContentItemTypeEnum | undefined> = {
```
luejoh: isnt it

Record<ShareTokenInfoResponseParentType, ContentItemTypeEnum>
?

src/composables/copy-content.composable.ts
```diff
@@ -120,22 +130,57 @@ export const useCopyContent = (copyItemType: Ref<ContentItemTypeEnum>) => {
...
 		text,
 		info,
 		warnings,
+		itemNameKey,
+	};
+};
+
+export const useImportContent = (shareTokenParentType: Ref<ShareTokenInfoResponseParentType | undefined>) => {
+	const contentItemType = computed(() =>
+		shareTokenParentType.value ? apiTypeToEntityType[shareTokenParentType.value] : undefined
```
luejoh: I still struggle a bit with the alternative name for the same thing.
We just have it, because the alternative name is easier to memorize?

src/composables/copy-content.composable.ts
```diff
@@ -120,22 +130,57 @@ export const useCopyContent = (copyItemType: Ref<ContentItemTypeEnum>) => {
...
+	);
+
+	return useCopyContent(contentItemType);
+};
+
+export const useShareContent = (shareTokenParentType: Ref<ShareTokenBodyParamsParentType | undefined>) => {
+	const { t } = useI18nGlobal();
+
+	const contentItemType = computed(() =>
+		shareTokenParentType.value ? apiTypeToEntityType[shareTokenParentType.value] : undefined
```
luejoh: Just do the lookup? should not need the ternary.

if the property doesant match, its also undefined

src/composables/copy-content.composable.ts
```diff
@@ -120,22 +130,57 @@ export const useCopyContent = (copyItemType: Ref<ContentItemTypeEnum>) => {
...
+	);
+
+	return {
+		...useCopyContent(contentItemType),
+		text: computed(() => t(`components.molecules.share.${shareTokenParentType.value}.options.infoText`)),
```
luejoh: hm didnt u disliked dynamic keys? Shall we do differently?

src/modules/data/application/loading.store.ts
```diff
@@ -20,9 +20,9 @@ export const useLoadingStore = defineStore("loadingStore", () => {
 	const withLoadingState = <T>(
 		fn: AsyncFunction<T>,
 		loadingMessage: string,
-		params: WithLoadingStateOptions = {}
+		options: LoadingStateOptions = {}
```
luejoh: hm should we even allow different delay+mindisplaytime options? 
I try to think of, when we ever would want to use different times. I can only think of changing such times for all, not for some.

src/modules/data/data-access/system.composable.ts
```diff
@@ -9,7 +9,7 @@ export const useSystem = (systemId: Ref<string | undefined>) => {
 
 	const { t } = useI18nGlobal();
 
-	const { execute, status, error, isRunning: isLoading } = useSafeAxiosTask();
+	const { execute, status, error, isLoading: isLoading } = useSafeAxiosTask();
```
luejoh: remove :isLoading ? 
Just less characters for the same

src/modules/feature/board/board/Board.vue
```diff
@@ -443,22 +439,36 @@ const onCopyBoard = async () => {
 	}
 };
 
-const shareModule = injectStrict(SHARE_MODULE_KEY);
+const {
+	isDialogOpen: isShareDialogOpen,
```
luejoh: some returned props contain share, some do not. Shall we either have them all include "share" or none? Dont have a clear preference tho.

luejoh: What do you think?

src/modules/feature/dashboard/DashboardTasks.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<SvsSuspense :loading="isLoadingTasks">
+	<SvsLoadingWrapper :is-loading="isLoadingTasks">
```
luejoh: hm was there an important reason to change the component name?
I liked that it refers to the same name, which the "<suspense" from vue has.

src/modules/feature/import/import-flow.composable.ts
```diff
@@ -72,11 +72,7 @@ export const useImportFlow = () => {
 		};
 	};
 
-	const getMessageKeyForImportFailure = (parentType: ShareTokenInfoResponse["parentType"]) => {
-		const key = getTranslationKeyForContentItem(mapShareTokenParentTypeToContentItemType(parentType));
-
-		return key ?? "common.labels.link";
-	};
+	const { itemNameKey } = useImportContent(computed(() => shareTokenInfo.value?.parentType));
```
luejoh: could  we put it before its been used? Its working, since both arent executed immediately, but from reading order, it makes more sense, no?

src/modules/feature/share/share-flow.composable.ts
```diff
@@ -0,0 +1,100 @@
...
+		const sharePath = buildSharePath(type, destinationType);
+		shareUrl.value = `${window.location.origin}/${sharePath}?import=${result?.token}`;
+
+		await resultAction.start();
+		closeDialog();
+
+		return { success, result };
+	};
+
+	const { itemNameKey } = useShareContent(computed(() => shareItemType.value));
```
luejoh: move topwards? Reading order?

src/modules/feature/share/ShareDialog.vue
```diff
@@ -0,0 +1,142 @@
...
+}>();
+
+const isDialogOpen = defineModel("is-open", {
+	type: Boolean,
+	default: false,
+});
+
+type StepType = "askOptions" | "showResult";
+const steps: StepType[] = ["askOptions", "showResult"];
+const stepIndex = ref(0);
```
luejoh, line 85: do we even need an index? There are only two steps and its one or another. I dont really see, why we need an array.

src/modules/feature/share/ShareDialog.vue
```diff
@@ -0,0 +1,142 @@
...
+const activeStep = computed(() => steps[stepIndex.value]);
+const isLastStep = computed(() => stepIndex.value === steps.length - 1);
+
+const shareOptions = reactive<ShareOptions>({ isSchoolInternal: true, hasExpiryDate: true });
+
+const currentStepTitle = computed((): string => {
+	if (activeStep.value === "askOptions") {
```
luejoh: ternary?

src/modules/feature/share/ShareDialog.vue
```diff
@@ -0,0 +1,142 @@
...
+
+type StepType = "askOptions" | "showResult";
+const steps: StepType[] = ["askOptions", "showResult"];
+const stepIndex = ref(0);
+const activeStep = computed(() => steps[stepIndex.value]);
+const isLastStep = computed(() => stepIndex.value === steps.length - 1);
+
+const shareOptions = reactive<ShareOptions>({ isSchoolInternal: true, hasExpiryDate: true });
+
+const currentStepTitle = computed((): string => {
```
luejoh: remove type and infer?

src/modules/feature/share/ShareDialog.vue
```diff
@@ -0,0 +1,142 @@
...
+const resetDialog = () => {
+	stepIndex.value = 0;
+	shareOptions.isSchoolInternal = true;
+	shareOptions.hasExpiryDate = true;
+};
+
+watch(
+	isDialogOpen,
+	(isOpen) => {
+		if (isOpen) {
```
luejoh, line 135: hm i think, we had a comment on the last pr too about such.

Is that keeping track if whether the dialog was reopened?
@after-leave wasnt doing the trick?

src/modules/ui/containers/SvsLoading.vue
```diff
@@ -0,0 +1,21 @@
+<template>
+	<div v-if="isLoading" class="d-flex my-10 justify-center align-center">
```
luejoh, line 2: Considering that svssuspense was removed in favor of svsloadingwrapper, could we remove the "wrapper" part from the name?
I feel like such arent adding any real value to a components name.

Can we also add a comment, that points to our helper managing the loading state timings?

Just looking at the component, you wont find out, that the loading state times are managed by using our async utils. If a loading state is managed by sth. else, the timings would not apply.

Alternatively we could also consider "recovering" svssuspense, doing these timings explicitely and having a SvsLoading in parallel without such.

You could choose between "delayed" animations and non delayed ones.

Somewhat bad comment, since we were already discussing a lot about this topic and where to manage those times and have them scale.

I just struggle with the component not being able to do it.
Maybe have somebody like martin adding some feedback?

src/pages/course-rooms/CourseRoomDetails.page.vue
```diff
@@ -429,6 +426,34 @@ const onCopyRequested = async ({ id, type }: { id: string; type: ContentItemType
...
+
+const onShareCourse = () =>
+	onShareRequested({
+		id: courseId.value,
+		type: ShareTokenBodyParamsParentType.COURSES,
+	});
+
+const onShareRequested = (params: ShareParams) => {
+	const featureFlagByType: Partial<Record<ShareTokenBodyParamsParentType, boolean>> = {
+		[ShareTokenBodyParamsParentType.COURSES]: useEnvConfig().value.FEATURE_COURSE_SHARE,
```
luejoh: hm this map from type to feature flag is actually a computed no?

src/pages/course-rooms/CourseRoomDetails.page.vue
```diff
@@ -429,6 +426,34 @@ const onCopyRequested = async ({ id, type }: { id: string; type: ContentItemType
...
+		[ShareTokenBodyParamsParentType.COLUMN_BOARD]: useEnvConfig().value.FEATURE_COLUMN_BOARD_SHARE,
+		[ShareTokenBodyParamsParentType.LESSONS]: useEnvConfig().value.FEATURE_LESSON_SHARE,
+		[ShareTokenBodyParamsParentType.TASKS]: useEnvConfig().value.FEATURE_TASK_SHARE,
+	};
+	if (featureFlagByType[params.type]) {
```
luejoh: could we have a name like "is..."

luejoh (approved):

src/components/course-rooms/CourseRoomDashboard.vue
uidp: Done.

src/composables/async-tasks.composable.ts
uidp: Added comment.

src/composables/confirmation-dialog.composable.ts
uidp: renamed to `confirmationAction`.

src/composables/copy-content.composable.ts
uidp: right. fixed.

src/composables/copy-content.composable.ts
uidp: We could maybe get rid of `useImportContent` when:
- `ContentItemTypeEnum` would had the same values like `ShareTokenBodyParamsParentType` wich would be doable
- we added a cast to `ContentItemTypeEnum` on the calling side which would not be nice IMHO

The composable `useShareContent` additionally overwrites the `text` value. So I think that cannot be removed easily.

src/composables/copy-content.composable.ts
uidp: requires a non-null assertion on 'shareTokenParentType.value' but works. updated.

src/composables/copy-content.composable.ts
uidp: good point :-) I fixed that by adding a static lookup map.

src/modules/data/application/loading.store.ts
uidp: well, the easier the better. I agree. updated.

src/modules/data/data-access/system.composable.ts
uidp: Fixed

src/modules/feature/board/board/Board.vue
uidp: I think that adding the `share` to `isDialogOpen` makes sense because if we removed the `share` from the other revealed props we would probably have to rename them to something specific on the calling side anyway. better to provide the specific name. I also changed `isDialogOpen` in `copy-flow.composable.ts` to `isCopyDialogOpen` for consistency.

src/modules/feature/dashboard/DashboardTasks.vue
uidp: well, the component doesn't actually work like the `<Suspense>` from vue because it doesn't handle async components (like with `async setup` or `await defineComponent`). it's just a component that conditionally renders a loading spinner.

src/modules/feature/import/import-flow.composable.ts
uidp: fixed

src/modules/feature/share/share-flow.composable.ts
uidp: fixed

src/modules/feature/share/ShareDialog.vue
uidp: I agree but this is just an attempt for a general pattern. easier to understand IMHO and could be extended easier.

src/modules/feature/share/ShareDialog.vue
uidp: fixed

src/modules/feature/share/ShareDialog.vue
uidp: fixed

src/modules/feature/share/ShareDialog.vue
uidp: In the other PR (in `ImportDialog.vue`) it was to conditionally initialize the first step based on the type property. here it would likely work without the watcher because there is no difference in the first step. but I'd rather keep the clean initialization. works like the other dialog. same pattern.

src/modules/ui/containers/SvsLoading.vue
uidp: well I understand the wish to implement the debounced loading display entirely in one component. but I also think that it's not bad to just have components for displaying the state and functions (composables) that provide it. "one tool for the job" :-)

src/pages/course-rooms/CourseRoomDetails.page.vue
uidp: technically correct, yes. but one could argue that feature flags don't change at runtime and we could practically rely on them being initialized at this point. anyway, I updated to `computed`.

src/pages/course-rooms/CourseRoomDetails.page.vue
uidp: don't understand this one, sorry.

src/modules/ui/containers/SvsLoading.vue
uidp: I renamed the component to `SvsLoading`. much shorter.

src/pages/course-rooms/CourseRoomDetails.page.vue
MartinSchuhmacher: with the runtime config that might actually change, so I also like the change to computed

src/pages/course-rooms/CourseRoomDetails.page.vue
MartinSchuhmacher: maybe to determine like "isLesson" or something?! but not too sure

PR #4214, opened 2026-05-11
BC-11691 - Links in News

odalys-dataport:
# Short Description

BC-11691
https://github.com/hpi-schul-cloud/schulcloud-server/pull/6284

## Checklist before merging


src/modules/data/data-access/news.composable.ts
```diff
@@ -0,0 +1,128 @@
...
+		const isConfirmed = await askConfirmation({
+			title: "components.organisms.FormNews.remove.confirm.message",
+			confirmBtnKey: "components.organisms.FormNews.remove.confirm.confirm",
+		});
+		if (isConfirmed) {
+			return await execute(
+				() => newsApi.newsControllerDelete(newsId),
+				t("common.notifications.errors.notDeleted", { type: t("common.words.news") })
+			);
+		}
```
uidp: I think there's a final `return` missing. Otherwise we're returning `undefined` here which cannot be destructured in the calling code.

src/modules/page/news/index.ts
```diff
@@ -0,0 +1,5 @@
+import NewsCreatePage from "./NewsCreate.page.vue";
+import NewsDetailsPage from "./NewsDetails.page.vue";
+import NewsEditPage from "./NewsEdit.page.vue";
+
+export { NewsCreatePage, NewsDetailsPage, NewsEditPage };
```
uidp, line 5: just as a hint, we could also write:
```ts
export { NewsCreatePage } from "./NewsCreate.page.vue";
export { NewsDetailsPage } from "./NewsDetails.page.vue";
export { NewsEditPage } from "./NewsEdit.page.vue";
```

src/modules/page/news/NewsDetails.page.unit.ts
```diff
@@ -0,0 +1,104 @@
...
+			routes: [
+				{ path: "/news", name: "news-list", component: { template: "<div />" } },
+				{ path: "/news/:id", name: "news-details", component: { template: "<div />" } },
+			],
+		});
+		injectRouterMock(router);
+		await router.push({ name: "news-details", params: { id: news.id } });
+
+		const wrapper = mount(NewsDetailsPage, {
+			attachTo: document.body,
```
uidp: Is this necessary?

src/modules/page/news/NewsDetails.page.vue
```diff
@@ -0,0 +1,93 @@
...
+	});
+});
+
+const onEdit = () => {
+	router.push({ path: `/news/${newsInstance.value?.id}/edit` });
+};
+
+const onDelete = async () => {
+	await deleteNews(newsInstance.value?.id);
+	notifySuccess(t("components.organisms.FormNews.success.remove"));
```
uidp: Check for success before notify + redirect?

tsconfig.json
```diff
@@ -1,142 +1,124 @@
...
+			"@api-h5p": ["src/generated/h5pEditorApi/v3"],
+			"@api-common-cartridge": ["src/generated/commonCartridgeApi/v3"],
+			"@/*": ["src/*"],
+			"@@/*": ["*"]
+		},
+		"lib": ["esnext", "dom", "dom.iterable", "scripthost"],
+		"outDir": "dist"
+	},
+	"include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.vue", "tests/**/*.ts", "tests/**/*.tsx", "types/**/*.d.ts"],
+	"exclude": ["node_modules", "dist", "./src/components/base/components.js"]
```
uidp, line 124: Apparently we had spaces instead of tabs on `main` again?

uidp (changes requested):

src/modules/page/news/NewsDetails.page.vue
odalys-dataport: @luejoh I agree and am confused by this, too. Apparently at some point the function returned a `success`, can you explain?

tsconfig.json
odalys-dataport: Looks so, at least I hope my settings are correct :D

src/modules/data/data-access/news.composable.ts
```diff
@@ -0,0 +1,130 @@
...
+			confirmBtnKey: "components.organisms.FormNews.remove.confirm.confirm",
+		});
+		if (isConfirmed) {
+			return await execute(
+				() => newsApi.newsControllerDelete(newsId),
+				t("common.notifications.errors.notDeleted", { type: t("common.words.news") })
+			);
+		} else {
+			return { success: false };
+		}
```
uidp: As a hint I would recommend someting like:
```ts
if (!confirmed) return { success: false, error: new Error("Deletion cancelled") };

return await execute(
	() => newsApi.newsControllerDelete(newsId),
	t("common.notifications.errors.notDeleted", { type: t("common.words.news") })
);
```
Would read much easier IMHO and the return value would match the `TaskResult` type. We have someting similar in `src/modules/feature/copy/copy-flow.composable.ts` and `src/modules/feature/import/import-flow.composable.ts`.

uidp (approved):

src/modules/page/news/NewsDetails.page.unit.ts
```diff
@@ -0,0 +1,103 @@
...
+	beforeEach(() => {
+		setActivePinia(createTestingPinia({ stubActions: false }));
+		newsApi = mockApi<NewsApiInterface>();
+		axiosMock = mockAxiosInstance();
+
+		vi.spyOn(serverApi, "NewsApiFactory").mockReturnValue(newsApi);
+		initializeAxios(axiosMock);
+	});
+
+	afterEach(() => {
```
luejoh: This is rarely needed, but any test suite having those makes a reader feel, that this is crucial.
If state is depending on pinia, mocks are automatically reset upon beforeEach

src/modules/page/news/NewsDetails.page.unit.ts
```diff
@@ -0,0 +1,103 @@
...
+				{ path: "/news", name: "news-list", component: { template: "<div />" } },
+				{ path: "/news/:id", name: "news-details", component: { template: "<div />" } },
+			],
+		});
+		injectRouterMock(router);
+		await router.push({ name: "news-details", params: { id: news.id } });
+
+		const wrapper = mount(NewsDetailsPage, {
+			global: {
+				plugins: [createTestingVuetify(), createTestingI18n(), createTestingPinia()],
```
luejoh: testingpinia is already in beforeEach. Having it twice may change behavior

luejoh (approved):

src/pages/Dashboard.page.vue
```diff
@@ -24,7 +24,7 @@
 
 				<!-- Dashboard news -->
 				<EmptyState
-					v-if="latestNews.length === 0"
+					v-if="latestNews?.length === 0"
```
uidp: I'm not really liking that `latestNews` defaults to `undefined` and not to empty array. That should probably be fixed in `news.composable.ts` either by assigning `data` to a ref like:
```ts
const news = ref<NewsResponse[]>([]);
news.value = newsData
```
or using a `computed` property like:
```ts
const news = computed(() => newsData.value ?? []);
```

PR #4227, opened 2026-05-19
BC-11770 - Centralized Dialog Management

uidp:
# Short Description
Introduce a centralized dialog system for the UI that allows dialogs to be opened programmatically through a shared manager instead of each feature owning its own dialog state.

**The implementation should:**
- provide a single dialog host responsible for rendering dialogs
- use a shared dialog manager/service to control dialog lifecycle and results
- support promise-based dialog handling for confirm/prompt-style flows
- ensure dialog open state is managed centrally rather than inside each dialog
- integrate existing shared dialog shell/components for consistent styling and behavior
- allow only one active dialog at a time, with additional dialog requests queued
- correctly handle cancel/close/escape/outside-click scenarios
- guarantee cleanup so no dialog requests remain unresolved if the host is destroyed

**Outcome:** Dialogs across the application can be launched in a consistent, typed, and centrally managed way, with predictable behavior and no overlapping modal state management in feature components.

BC-11770

## Checklist before merging


src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+	item.resolve(item.pendingSettlement);
+
+	activateNextDialog();
+};
+
+export const openDialog = <K extends DialogKey>(
+	type: K,
+	props: DialogProps<K>
+): Promise<AwaitableResult<DialogResult<K>>> =>
+	new Promise((resolve) => {
```
luejoh: Hm I like that you can directly receive/await the result. I am missing tho, that i could identify/cancel the dialog from where I raise it.

In the notification store the cancel function is returned and allows the one raising the dialog to also cancel it whenever necessary. Controlling the dialog from where its been used should still be possible.
We could achieve that by returning functions, which would be able to do the necessary stuff.

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+		item.resolved = true;
+		item.resolve({
+			completed: false,
+			data: undefined,
+		});
+	}
+};
+
+export const useDialogManager = () => ({
+	activeDialog,
```
luejoh: not seeing the benefit of not putting the declaration inside the function scope. Can we put it right into it?

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+		if (item.resolved) continue;
+		item.resolved = true;
+		item.resolve({
+			completed: false,
+			data: undefined,
+		});
+	}
+};
+
+export const useDialogManager = () => ({
```
luejoh: shouldnt that be wrapped with createGlobalState() ?
Or become a store like the notifcation store? Makes tracking of states easier with pinia and testing too and its anyway supposed to be global state, no?
Then we also dont need "Manager", which we rarely have by now

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+
+	if (active && !active.resolved) {
+		active.resolved = true;
+		active.resolve({
+			completed: false,
+			data: undefined,
+		});
+	}
+
+	for (const item of queued) {
```
luejoh: isnt that just supposed to execute "cancelDialog" for all items?

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+	}
+};
+
+export const onDialogAfterLeave = (id: number) => {
+	finalizeSettlement(id);
+};
+
+export const cancelAllDialogsImmediately = () => {
+	const active = activeDialog.value;
+	const queued = [...queue.value];
```
luejoh: its the same after doing that, no? its just a copy

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+
+	activeDialog.value = undefined;
+	queue.value.splice(0, queue.value.length);
```
luejoh: I guess it was cloned to clear the real one. Can we not just iterate through all queued ones and cancel them and clear the array at the end?

luejoh: or at least do sth. like:

const toCancel = [activeDialog.value, ...queue.value]

toCancel.forEach(cancelDialog)

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+
+	item.modelValue = value;
+
+	if (value === false && !item.pendingSettlement) {
+		item.pendingSettlement = markRaw({ completed: false, data: undefined });
+	}
+};
+
+export const onDialogAfterLeave = (id: number) => {
+	finalizeSettlement(id);
```
luejoh: do we need an alias for the dialog leave action?
Settlement is somewhat a new term. Do we really need that term?

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+export const setDialogModelValue = (id: number, value: boolean) => {
+	const item = activeDialog.value;
+	if (!item) return;
+	if (item.id !== id) return;
+	if (item.resolved) return;
+
+	item.modelValue = value;
+
+	if (value === false && !item.pendingSettlement) {
+		item.pendingSettlement = markRaw({ completed: false, data: undefined });
```
luejoh: can you explain the markRaw usage here?

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+
+		if (activeDialog.value === undefined) {
+			activeDialog.value = item as AnyDialogRequest;
+		} else {
+			queue.value.push(item as AnyDialogRequest);
+		}
+	});
+
+export const completeDialog = <K extends DialogKey>(id: number, data: DialogResult<K>) => {
+	beginSettlement(id, { completed: true, data });
```
luejoh: This actually feels like we arent beginning, but finishing the dialog?

luejoh: somewhat also feels like that by having "complete" --> pointing to "begin"

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+	if (!item) return;
+	if (item.id !== id) return;
+	if (item.resolved) return;
+	if (item.pendingSettlement) return;
+
+	item.pendingSettlement = result;
+	item.modelValue = false;
+};
+
+const finalizeSettlement = (id: number) => {
```
luejoh: i feel like id want to work with the DialogRequest any time, instead of doing the id check. I also feel like those if checks are likely not necessary then.

If these "error cases" can occur, can we achieve, that they cannot happen? We are controlling the entire state. It should be possible to be sure, that neither of these conditions should occur.

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+	[K in DialogKey]: DialogRequest<K>;
+}[DialogKey];
+
+const activeDialog = ref<AnyDialogRequest | undefined>(undefined);
+const queue = ref<AnyDialogRequest[]>([]);
+
+let nextId = 1;
+
+const activateNextDialog = () => {
+	if (activeDialog.value || queue.value.length === 0) return;
```
luejoh: could that just become part of finilizeSettlement/afterleave?

By that a check against activeDialog should not be necessary, because its secured, that one lifecycle has ended and the next can begin.

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
+import { dialogRegistry } from "./dialog-registry";
+import type { AwaitableResult, DialogKey, DialogProps, DialogResult } from "./dialog-types";
+import { markRaw, ref } from "vue";
+
+type DialogRequest<K extends DialogKey = DialogKey> = {
```
luejoh: is "= DialogKey" needed?

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+	activateNextDialog();
+};
+
+export const openDialog = <K extends DialogKey>(
+	type: K,
+	props: DialogProps<K>
+): Promise<AwaitableResult<DialogResult<K>>> =>
+	new Promise((resolve) => {
+		const item: DialogRequest<K> = {
+			id: nextId++,
```
luejoh: lets use lodash's "uniqueId()" ?

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+	resolved: boolean;
+	pendingSettlement: AwaitableResult<unknown> | undefined;
+	resolve: (value: AwaitableResult<unknown>) => void;
+};
+
+type AnyDialogRequest = {
+	[K in DialogKey]: DialogRequest<K>;
+}[DialogKey];
+
+const activeDialog = ref<AnyDialogRequest | undefined>(undefined);
```
luejoh: you can remove both undefined and it will result in the same type

src/modules/feature/dialog/dialog-manager.ts
```diff
@@ -0,0 +1,137 @@
...
+	id: number;
+	type: K;
+	props: DialogProps<K>;
+	modelValue: boolean;
+	resolved: boolean;
+	pendingSettlement: AwaitableResult<unknown> | undefined;
+	resolve: (value: AwaitableResult<unknown>) => void;
+};
+
+type AnyDialogRequest = {
```
luejoh: replace by

`type AnyDialogRequest = DialogRequest<DialogKey>;
`

?

src/modules/feature/dialog/dialog-contracts.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ShareTokenBodyParamsParentType, ShareTokenInfoResponse } from "@api-server";
+import { ImportDestination, ImportDestinationItem, ImportDestinationType } from "@feature-import";
+
+export type ConfirmationOptions = {
+	title: string;
+	message?: string;
+	messageType?: "warning" | "info";
+	confirmBtnKey?: string;
+};
+export type ConfirmationDialogProps = { options: ConfirmationOptions };
```
luejoh: id prefer not doing options object within prop objects.
Makes understanding the data structure more difficult

src/modules/feature/dialog/dialog-types.ts
```diff
@@ -0,0 +1,24 @@
+import { dialogRegistry } from "./dialog-registry";
+import type { Component } from "vue";
+
+export type AwaitableResult<T> = { completed: true; data: T } | { completed: false; data: undefined };
```
luejoh: this type already exists no? in awaitable-action.composable

src/modules/feature/dialog/dialog-types.ts
```diff
@@ -0,0 +1,24 @@
...
+
+declare const propsTypeSymbol: unique symbol;
+declare const resultTypeSymbol: unique symbol;
+
+export type DialogDefinition<Props, Result> = {
+	component: Component;
+	readonly [propsTypeSymbol]?: Props;
```
luejoh: why do we need a symbol? I guess there is no need to use it?

src/modules/feature/dialog/dialog-registry.ts
```diff
@@ -0,0 +1,27 @@
...
+} from "./dialog-contracts";
+import { defineDialog } from "./dialog-types";
+import ConfirmationDialog from "./dialogs/ConfirmationDialog.vue";
+import ConfirmDialog from "./dialogs/ConfirmDialog.vue";
+import PromptDialog from "./dialogs/PromptDialog.vue";
+import { CopyDialog } from "@feature-copy";
+import { ImportCardDialog, ImportDialog } from "@feature-import";
+import { ShareDialog } from "@feature-share";
+
+export const dialogRegistry = {
```
luejoh, line 30: can we put together the registry and the types?
I think we dont need them much further elsewhere

src/modules/feature/dialog/dialog-types.ts
```diff
@@ -0,0 +1,24 @@
...
+	readonly [resultTypeSymbol]?: Result;
+};
+
+export const defineDialog = <Props, Result>(component: Component): DialogDefinition<Props, Result> => ({ component });
+
+export type DialogRegistry = typeof dialogRegistry;
+export type DialogKey = keyof DialogRegistry;
+
+export type PropsOf<T> = T extends DialogDefinition<infer P, unknown> ? P : never;
```
luejoh: do we need the PropsOf types?

Declare the dialog props and dialogresult straight?

`export type DialogProps<K extends DialogKey> =
	DialogRegistry[K] extends DialogDefinition<infer P, unknown> ? P : never;

export type DialogResult<K extends DialogKey> =
	DialogRegistry[K] extends DialogDefinition<unknown, infer R> ? R : never;`

src/modules/feature/dialog/dialogs/PromptDialog.vue
```diff
@@ -0,0 +1,39 @@
...
+import { computed } from "vue";
+import { VTextField } from "vuetify/components";
+
+const props = defineProps<{
+	title: string;
+	placeholder?: string;
+	initialValue?: string;
+}>();
+
+const emit = defineEmits<{
```
luejoh: i guess, since all of them are using svsdialog, all the bindings to emits n stuff is identical between the dialog variations we are having. Maybe we can inherit them somehow to ensure, that the same behavior is secured and never forgotten

src/modules/feature/dialog/dialog-manager.ts
uidp: ~Well, I would rather like to have the dialog state in the module scope than in the function scope. Or do you suggest that we can define more than one dialog state per module?~

src/modules/feature/dialog/dialog-manager.ts
uidp: ~I think this is related to the former question about function scope vs module scope.~

src/modules/feature/dialog/dialog-manager.ts
luejoh: Ye related to the comment about the global state.
By having it currently on module scope, the state is automatically global, but not tracked as it would be with pinia. 
I would not suggest more than one dialog state. There is one single source of truth --> the module, the global composable or the pinia store. id tend to store tho, since we did similar for global state in most cases.

What do you think?

src/modules/feature/dialog/dialog-manager.ts
luejoh: yes

uidp: I decided to use a pinia store for better testability, devTool support etc. Updated the code accordingly.

src/modules/feature/dialog/dialog-manager.ts
uidp: Well, `cancelDialog` is calling `beginSettlement` which only ever operates on `activeDialog`:
```ts 
const item = activeDialog.value;  // only looks at activeDialog
if (item.id !== id) return;       // silently no-ops if id doesn't match
```
The loop in `cancelAllDialogsImmediately` intentionally bypasses that indirection. The intention of both functions is different:
- `cancelDialog`: cancel the active dialog (`id` as a safeguard to protect "cross-cancellation")
- `cancelAllDialogsImmediately`: cancel all dialogs - active and queued

src/modules/feature/dialog/dialog-manager.ts
uidp: We're creating a shallow copy of `queued.value` as a "snapshot". Otherwise we wouldn't be able to resolve all queued items in the for loop below after clearing the queue:
```ts
queue.value.splice(0, queue.value.length);
```
So the copy is really only a way to hold on to the root item references.

src/modules/feature/dialog/dialog-manager.ts
uidp: We cannot use `cancelDialog` in that way (see comment above).

src/modules/feature/dialog/dialog-manager.ts
uidp: Also `cancelDialog` is ultimately activating the next dialog from the queue. Which is most likely not what we want during cleanup.

src/modules/feature/dialog/dialog-manager.ts
uidp: On the alias:
- `onDialogAfterLeave ` is a public event hook. The dialog component calls it as the @after-leave event handler
- `finalizeSettlement` is an internal action name
So the alias basically decouples the external and internal API.

On the term "settlement":
Currently closing a dialog is a  two-step async process:
- Step 1: (`beginSettlement`): "We know the result, start the close animation" => sets `modelValue = false`
- Step 2: (`finalizeSettlement`): "The animation finished, the dialog is gone" => resolve the promise + show the next dialog
The term "settlement" describes a state of a transaction when all parties receive what was agreed. As an alternative we could maybe use `beginResolve`/`finalizeResolve`. What do you think?

src/modules/feature/dialog/dialog-manager.ts
uidp: That was an attempt to make the assignment non-reactive because we need only a plain object. However this inconsistent with teh assignment in `beginSettlement` and also not really needed. So I removed the `markRaw`.

src/modules/feature/dialog/dialog-manager.ts
uidp: This is basically a view from two different perspectives:
- `completeDialog`: from the callers perspective - the dialog's user interaction is complete
- `beginSettlement`: from the store's perspective - the promise resolution process has started
I renamed `completeDialog` and `cancelDialog` to `onCompleteDialog` and `onCancelDialog` in order to more reflect that these are event handlers. Also more consistent with `onDialogAfterLeave`.

src/modules/feature/dialog/dialog-manager.ts
uidp: Right, we can remove the `id` lookup and rely on referential equality. Updated the code.

src/modules/feature/dialog/dialog-manager.ts
uidp: Right, can be part of `finalizeSettlement`. Updated the code.

src/modules/feature/dialog/dialog-manager.ts
uidp: No, not needed. Removed.

src/modules/feature/dialog/dialog-manager.ts
uidp: Done.

src/modules/feature/dialog/dialog-manager.ts
uidp: Fixed.

src/modules/feature/dialog/dialog-manager.ts
uidp: This is a mapped type for building a discriminated union that keeps the `type` and `props`correlated, e.g. if if `type` is `"confirmation"`, then `props` must match `DialogProps<"confirmation">`and so forth.

src/modules/feature/dialog/dialog-contracts.ts
uidp: Right, removed the unneccessary nesting level. Also using `ConfirmationDialogProps` in `src/utils/confirmation-dialog.utils.ts` now.

src/modules/feature/dialog/dialog-types.ts
uidp: I removed `awaitable-action.composable` because it's no longer needed. We can re-add it when we see use.

src/modules/feature/dialog/dialog-types.ts
uidp: The symbols act as phantom type fields - they don't exist at runtime, but they're essential to make `PropsOf` and `ResultOf` work.
```ts
export type PropsOf<T> = T extends DialogDefinition<infer P, unknown> ? P : never;
```
TypeScript can only `infer P` here because the symbol field `[propsTypeSymbol]?: Props` gives it something structural to match against. Without the symbols the type would return `never`.

src/modules/feature/dialog/dialog-registry.ts
uidp: Done. Also makes `defineDialog` module-private and avoids circular imports.

src/modules/feature/dialog/dialog-types.ts
uidp: Done.

src/modules/feature/dialog/dialogs/PromptDialog.vue
uidp: A general parent dialog like `SvsManagedDialog` that handles all event bindings would be technically possible but would hit a wall at `@complete`. For example:
- ConfirmationDialog: `emit('complete', true)`
- ImportDialog: `emit('complete', { newName, destination })` (from form fields)

Instead we could have someting like:
```ts
export type ManagedDialogEmits<Result> = {
	complete: [result: Result];
	cancel: [];
	"after-leave": [];
};
const emit = defineEmits<ManagedDialogEmits<boolean>>();
```
But I don't like that very much. What do you think?

src/modules/feature/dialog/dialog-manager.ts
uidp: Now also returning a `cancel` function that can be used to cancel the dialog.

src/modules/feature/dialog/dialog-manager.ts
luejoh: hm lets put names reflecting whats happening then instead?

animateOut
afterLeave
dialogLeft

Settlement is a bit more abstract. I did not immediately had an idea, what this was about.
Its part of the dialogs lifecycle. Id tend to think of either putting a term and its state alltogether, like openDialog, gatherDialogResults, startAnimateOut, onAnimateOut. It feels a bit like named steps.

In the end, i think its all about naming, what I try to improve even further.

src/modules/feature/dialog/dialogs/PromptDialog.vue
luejoh: Ye not to much as it is also not used multiple times, which I was hoping for. Fine to revert it then. Sorry for this bothering 👏

luejoh (approved):

luejoh (approved):

PR #4268, opened 2026-06-10
BC-11730 - fix logsteps being undefined

hoeppner-dataport:
# Short Description
Ensure that logSteps property of board-report-error is never undefined

BC-11730

## Checklist before merging


src/modules/data/board/socket/socket-error-handler.ts
```diff
@@ -133,13 +133,13 @@ export const useConnectionErrorHandling = (socket: Socket) => {
 		}
 	});
 
-	useEventListener(window, "beforeunload", async () => {
+	useEventListener(globalThis, "beforeunload", async () => {
 		await reportLogs("page_unload");
 	});
 
 	const reportLogs = async (cause: string) => {
```
sebmue-dataport: Since we are not awaiting apiCall anymore I think this does not need to be async anymore. Same for line 130, 132, 136 and 137.

sebmue-dataport (changes requested):

src/modules/data/board/socket/socket-error-handler.ts
hoeppner-dataport: fixed

sebmue-dataport (approved):

PR #4296, opened 2026-07-02
BC-11362-double-click-cursor-card

luejoh:
# Short Description

Ensure double click puts focus on element being clicked at.

BC-11362

## Checklist before merging


src/modules/data/board/BoardFocusHandler.composable.ts
uidp, line 1: BTW (not part of this PR) the `id?.valueOf()` seems not right. I think `toValue(id)` would be the right way to do that.

src/modules/data/board/BoardFocusHandler.composable.ts
```diff
@@ -121,8 +126,14 @@ export function useBoardFocusHandler(
...
 	 */
+	const interactionHandled = inject<Ref<boolean>>(InlineEditInteractionHandled, shallowRef(false));
 	useInlineEditInteractionHandler(async () => {
-		await forceFocusOnMount();
+		await nextTick();
+		if (onFocusReceived !== undefined) {
+			onFocusReceived();
+			interactionHandled.value = true;
+		}
+		setFocus(id);
```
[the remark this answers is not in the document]
uidp: 👍

src/modules/feature/board-link-element/components/LinkContentElement.vue
```diff
@@ -189,4 +194,13 @@ const onClick = (event: MouseEvent | KeyboardEvent) => {
 		focusNodeFromHash();
 	}
 };
+
+watch(
+	() => props.isEditMode,
+	(newVal) => {
+		if (!newVal) {
```
uidp, line 201: `newVal == false`?

uidp (approved):

src/modules/data/board/BoardFocusHandler.composable.ts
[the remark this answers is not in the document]
MartinSchuhmacher: intentional, as we need the distinction between the element being a content element (received focus) or not. So that for the empty card (no content elemnet, only title) we can have that the interaction was not yet handled

src/modules/feature/board-link-element/components/LinkContentElement.vue
MartinSchuhmacher: no realy advantage over `!newVal`, rather `newVal === false` would be more strict, but I want to keep it with `!newVal` for the current functionality

PR #4297, opened 2026-07-03
BC-11940 - Wrong Behavior After Room Deletion

uidp:
# Short Description

If Person A (via the administration interface) deletes a room while another person (Person B) currently has it open, Person B is redirected to an empty course.

BC-11940

## Checklist before merging

### DIFF
--- a/src/modules/data/folder/Folder.state.ts
+++ b/src/modules/data/folder/Folder.state.ts
@@ -70,10 +70,8 @@ export const useFolderState = () => {
 
 			fileFolderElement.value = castToFileFolderElement(response.data.element);
 			parentNodeInfos.value = response.data.parentHierarchy;
-			return true;
 		} catch (error) {
 			handleApplicationError(error);
-			return false;
 		}
 	};
 
--- a/src/modules/data/folder/Folder.state.ts
+++ b/src/modules/data/folder/Folder.state.ts
@@ -5,12 +5,14 @@ import {
 	ParentNodeInfo,
 	ParentNodeType,
 } from "@/types/board/ContentElement";
-import { $axios, mapAxiosErrorToResponseError } from "@/utils/api";
+import { HttpStatusCode } from "@/types/enum/http-status-code.enum";
+import { $axios } from "@/utils/api";
 import { createApplicationError } from "@/utils/create-application-error.factory";
 import { buildPageTitle } from "@/utils/pageTitle";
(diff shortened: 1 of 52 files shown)


src/modules/data/folder/Folder.state.ts
```diff
@@ -68,8 +70,10 @@ export const useFolderState = () => {
 
 			fileFolderElement.value = castToFileFolderElement(response.data.element);
 			parentNodeInfos.value = response.data.parentHierarchy;
+			return true;
 		} catch (error) {
-			throwApplicationError(error);
+			handleApplicationError(error);
+			return false;
```
MartinSchuhmacher: not quite sure why we need the return here because the `handleApplicationError` should handle the throw, right?

I kind of understand the return in Line 73, but not this one.

Is that only because of the different handling in `Folder.vue` during the `onMounted` ? I see the implication and with that the need for adjustment. But implementing that with the boolean return value somehow feels too "constructed" if you know what I mean.

src/modules/page/room/RoomMembers.page.vue
```diff
@@ -266,6 +266,9 @@ onMounted(async () => {
 	if (room.value === undefined) {
 		await fetchRoom(roomId);
 	}
+	if (room.value === undefined) {
```
MartinSchuhmacher, line 269: why do we have a duplicate condition here, cant that be handled after the `fetchRoom` from above?

uidp: This is actual two sequential guards — fetch if needed, then bail if still undefined. We could write that in a nested form like:
```ts
	if (room.value === undefined) {
		await fetchRoom(roomId);
		if (room.value === undefined) {
			return;
		}
	}
```
but I think that the sequential version is more clear.

src/modules/data/folder/Folder.state.ts
uidp: Yes, I a gree on the boolean return value. Replaced with a check for existence of  loaded `fileFolderElement`.

src/modules/page/room/RoomMembers.page.vue
MartinSchuhmacher: I just remember that when we have similar situations we do 
```ts
....
const result = await fetchRoom(...)
if (!result) ...
```

But if your approach fits better here I'm fine with it.

MartinSchuhmacher (approved):

odalys-dataport (approved):

PR #4302, opened 2026-07-08
BC-11989 - Paginate room list

sebmue-dataport:
# Short Description

If skip/limit are not provided the application will try to fetch all data at once, leading to issues in an aggregation with many rooms. Therefore loading those in batches of 500 each is implemented.

Since the frontend does not use actual server-pagination but instead fetches all the data at once this behavior is kept, but all data is fetched in batches.

BC-11989
https://github.com/hpi-schul-cloud/schulcloud-server/pull/6378

## Checklist before merging


muratmerdoglu-dp (approved):

src/modules/data/room/manageRoom/AdministrationRoom.store.ts
```diff
@@ -6,6 +6,8 @@ import { notifyError, useSchoolStoreRefs } from "@data-app";
 import { defineStore } from "pinia";
 import { computed, ref } from "vue";
 
+const DEFAULT_ROOM_STATS_LIMIT = 50;
```
Metauriel: remove this

src/modules/data/room/manageRoom/AdministrationRoom.store.ts
```diff
@@ -33,21 +35,42 @@ export const useAdministrationRoomStore = defineStore("administrationRoomStore",
 			});
 	};
 
+	// this is a workaround. Proper (server-based) pagination should be implemented across all our tables in future
```
Metauriel: I think this comment can be removed as well

src/modules/data/room/manageRoom/AdministrationRoom.store.ts
```diff
@@ -33,21 +35,42 @@ export const useAdministrationRoomStore = defineStore("administrationRoomStore",
...
+		const firstBatch = (await roomApi.roomControllerGetRoomStats(0, batchSize)).data;
+		if (firstBatch.total > batchSize) {
+			rooms.push(...firstBatch.data);
+
+			for (let skip = batchSize; skip < firstBatch.total; skip += batchSize) {
+				const nextBatch = (await roomApi.roomControllerGetRoomStats(skip, batchSize)).data;
+				rooms.push(...nextBatch.data);
+			}
+		} else {
+			rooms.push(...firstBatch.data);
```
Metauriel: If these are pushed into the array anyway, you can pull that before the if and get rid of the else statement.
If you prefer, you can even move the initial declaration of the rooms array after the firstBatch, and initialize it with the first 500 rooms

Metauriel (changes requested):

Metauriel (approved):

PR #4318, opened 2026-07-23
BC-12044 - Prevent Infinite Loop

muratmerdoglu-dp:
# Short Description

- BC-12044

## Checklist before merging


sebmue-dataport (approved):

NFriedo:
Unfortunately this still does not work with e.g. https://bc-12044-infinite-loop.dbc.dbildungscloud.dev/boards/1

(I am still looking into this to understand the whole issue but don't expect too much 😅 )

sebmue-dataport (changes requested):
What Nicole said. It works for the static routes such as /boards but not for dynamic ones with an invalid ID it seems

NFriedo:
Ideally we would have a single source of routing but to change that would be a huge change.


# The Problem 
Example: `/boards` (as I understand it):
Browser: `/boards`-> [ingress](https://github.com/hpi-schul-cloud/dof_app_deploy/blob/main/ansible/group_vars/all/x_ingress.yml) sent it to nuxt-client -> nuxt client checks vue routes, is not in there -> must be legacy route ([code](src/router/guards/legacy-route-compatibility.guard.ts)), `window.assign` redirect to ingress again -> infinite loop
(legacy client is not involved in the redirect loop!)

That means EVERY invalid route that gets routed to the vue client is an infinite redirect (e.g. `/rooms/1` because it's partially part of the valid route `/rooms`)

# My proposal would be the following:
1. Define a fallback for routes that don't exist in the vue router. 
```ts
// routes.ts
	{
		path: "/:pathMatch(.*)*",
		name: "not-found",
		beforeEnter: () => {
			useAppStore().handleApplicationError(HttpStatusCode.NotFound);
			return true;
		},
		component: () => import("@/pages/Error.page.vue"),
		meta: {
			isPublic: true,
		},
	},
```

I guess the reason we didn't have this before is that we didn't know, for non existing routes, if they might be valid legacy routes. 

Example:  `:to="/teams"` (we don't have this, but we still need to catch such cases)

For this reason we have the `legacy-route-compatibility.guard.ts`

2. Adjust the `legacy-route-compatibility.guard.ts`. 
```ts
export const legacyCompatibilityGuard: NavigationGuard = (to: RouteLocationNormalized) => {
	if (isLegacyClient(to.path)) {
		if (window.location.pathname === to.path) {
			return true; // already here via server routing, don't loop
		}
		window.location.assign(to.path);
		return false;
	}
	return true;
};
```

(Actually `isLegacyClient` is not the correct name because it's only checks if the path has a valid entry.)

We need to differentiate if we already went through the ingress check or not.
If the window location is the same as our target, ingress already made its decision and we use the vue router fallback.
If it isn't, let ingress decide first.

Disclaimer: This is pretty complicated and I am not 100% sure if it's exactly how it works but I tested this locally (with adjustments in the config/vite/dev-server-proxy-plugin.ts for testing) and it seems to work.

muratmerdoglu-dp:
The catch-all fallback route is added to invalidRoutes. 
Catch-all is defined under Vue-owned prefixes, so invalid IDs (e.g. `/boards/1`) stay on Vue and show `NotFound` instead of being proxied to the legacy client.

NFriedo (changes requested):
> The catch-all fallback route is added to invalidRoutes. Catch-all is defined under Vue-owned prefixes, so invalid IDs (e.g. `/boards/1`) stay on Vue and show `NotFound` instead of being proxied to the legacy client.

The issue is still that "EVERY invalid route that gets routed to the vue client is an infinite redirect (e.g. /rooms/1 because it's partially part of the valid route /rooms)".
So I think we shouldn't hard code some routes in there. 
And the legacy client is not involved in the redirect loop.

Did you have a look at my proposal? (Note: this is a complete alternative to your solution and wouldn't need changes to vue-client-route.ts or to list invalid routes in routes.ts)

EDIT: I created a test branch(`BC-12044-alternative-solution-test`) to test my proposal: https://github.com/hpi-schul-cloud/nuxt-client/commit/4008c23aad1c7678988f793939d8411974cdf10c

src/router/guards/legacy-route-compatibility.guard.unit.ts
```diff
@@ -0,0 +1,70 @@
...
+			mockedIsLegacyClient.mockReturnValue(true);
+			vi.stubGlobal("location", { pathname: "/dashboard", assign: vi.fn() });
+
+			const result = legacyCompatibilityGuard(homeworkLocation, dashboardLocation, vi.fn());
+
+			expect(result).toBe(false);
+			expect(window.location.assign).toHaveBeenCalledWith("/homework");
+		});
+
+		it("should forward query parameters when leaving the Vue client", () => {
```
hoeppner-dataport:
```suggestion
		describe('when forwarding to legacy client', () => {
				it("should forward query parameters, too", () => {
```

src/router/guards/legacy-route-compatibility.guard.unit.ts
```diff
@@ -0,0 +1,70 @@
...
+			mockedIsLegacyClient.mockReturnValue(true);
+			vi.stubGlobal("location", { pathname: "/dashboard", assign: vi.fn() });
+
+			legacyCompatibilityGuard(buildLocation("/homework", "/homework?tab=open"), dashboardLocation, vi.fn());
+
+			expect(window.location.assign).toHaveBeenCalledWith("/homework?tab=open");
+		});
+
+		it("should not redirect when the browser is already on the legacy path", () => {
```
hoeppner-dataport:
```suggestion
		describe('when the browser is already on a legacy path', () => {
				it("should not redirect", () => {
```

NFriedo (approved):

src/router/routes.ts
```diff
@@ -425,4 +425,13 @@ export const routes: Readonly<RouteRecordRaw>[] = [
 		component: () => import("@/pages/Home.page.vue"),
 		name: "home",
 	},
+	{
+		path: "/:pathMatch(.*)*",
```
hoeppner-dataport, line 429: This looks like a catch-all in the end. So: if no route fits, it goes to the not-found errorpage.
But if there was a legacy-client route - that exists... but is not provided to the isLegacyClient-function, then this route will not work any longer.
That looks dangerous and needs to be discussed - at least with QA and/or PO.

hoeppner-dataport (changes requested):

src/router/routes.ts
muratmerdoglu-dp: I think it is not dangerous for normal legacy routes served by ingress to legacy. And `isLegacyClient` function is not a kind of whitelist of legacy routes. Everything not in `vue-client-route.ts` is treated as legacy. So if there is a legacy route, then ingress redirects it to the legacy client and doesn't hit the catch-all in the end.

hoeppner-dataport (approved):

sebmue-dataport (approved):

PR #4324, opened 2026-07-27
BC-12074 Fix FileContentElement state display and user warning

luejoh:
# Short Description

After introducing EmptyElement for created, but not filled with content elements, the FileContentElement did not function as before anymore. See related ticket.

BC-12074

## Checklist before merging

### DIFF
--- a/src/modules/feature/board-file-element/FileContentElement.vue
+++ b/src/modules/feature/board-file-element/FileContentElement.vue
@@ -80,7 +80,7 @@ import { BoardMenu, BoardMenuScope, EmptyElement } from "@ui-board";
 import { KebabMenuActionDelete, KebabMenuActionMoveDown, KebabMenuActionMoveUp } from "@ui-kebab-menu";
 import { LightBoxContentType, LightBoxOptions, useLightBox } from "@ui-light-box";
 import { useDebounceFn } from "@vueuse/core";
-import { computed, onMounted, ref, toRef, watch } from "vue";
+import { computed, onBeforeUnmount, onMounted, ref, toRef, watch } from "vue";
 import { useI18n } from "vue-i18n";
 import { useRouter } from "vue-router";
 
@@ -120,8 +120,9 @@ const fileRecord = computed(() => getFileRecordsByParentId(element.value.id)[0])
 
 const { alerts, addAlert } = useFileAlerts(fileRecord);
 
-const isUploading = computed(() => fileRecord.value?.isUploading);
+const fileWasPicked = ref(false);
 const uploadProgress = ref(0);
+const isUploading = computed(() => fileRecord.value?.isUploading || fileWasPicked.value);
 
 const fileProperties = computed(() => {
 	if (fileRecord.value === undefined) {
@@ -162,9 +163,23 @@ watch(element.value, async () => {
 	isLoadingFileRecord.value = false;
 });
 
+const handleBeforeUnload = (event: BeforeUnloadEvent) => {
+	if (fileWasPicked.value) {
+		// Opens confirmation dialog in firefox
+		event.preventDefault();
+		// Opens confirmation dialog in chrome
+		event.returnValue = "";
+	}
+};
+
 onMounted(async () => {
+	window.addEventListener("beforeunload", handleBeforeUnload);
 	await tryFetchFiles(element.value.id, FileRecordParentType.BOARDNODES);
 	isLoadingFileRecord.value = false;
+});
+
+onBeforeUnmount(() => {
+	window.removeEventListener("beforeunload", handleBeforeUnload);
 });
 
 const onKeydownArrow = (event: KeyboardEvent) => {
@@ -174,8 +189,9 @@ const onKeydownArrow = (event: KeyboardEvent) => {
 	}
 };
 
-const onUploadFile = async (file: File): Promise<void> => {
-	uploadProgress.value = 0;
+const onUploadFile = async (file: File) => {
+	fileWasPicked.value = true;
+  uploadProgress.value = 0;
 	try {
 		await upload(file, element.value.id, FileRecordParentType.BOARDNODES, (progress) => {
 			uploadProgress.value = progress;
@@ -184,7 +200,8 @@ const onUploadFile = async (file: File): Promise<void> => {
 	} catch {
 		emit("delete:element", element.value.id);
 	} finally {
-		uploadProgress.value = 0;
+		fileWasPicked.value = false;
+    uploadProgress.value = 0;
 	}
 };
 
--- a/src/modules/feature/board-file-element/FileContentElement.vue
+++ b/src/modules/feature/board-file-element/FileContentElement.vue
@@ -191,7 +191,7 @@ const onKeydownArrow = (event: KeyboardEvent) => {
 
 const onUploadFile = async (file: File) => {
 	fileWasPicked.value = true;
-  uploadProgress.value = 0;
+	uploadProgress.value = 0;
 	try {
 		await upload(file, element.value.id, FileRecordParentType.BOARDNODES, (progress) => {
 			uploadProgress.value = progress;
@@ -201,7 +201,7 @@ const onUploadFile = async (file: File) => {
 		emit("delete:element", element.value.id);
 	} finally {
 		fileWasPicked.value = false;
-    uploadProgress.value = 0;
+		uploadProgress.value = 0;
 	}
 };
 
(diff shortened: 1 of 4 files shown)


src/modules/feature/board-file-element/FileContentElement.vue
```diff
@@ -120,8 +120,9 @@ const fileRecord = computed(() => getFileRecordsByParentId(element.value.id)[0])
 
 const { alerts, addAlert } = useFileAlerts(fileRecord);
 
-const isUploading = computed(() => fileRecord.value?.isUploading);
+const fileWasPicked = ref(false);
 const uploadProgress = ref(0);
+const isUploading = computed(() => fileRecord.value?.isUploading || fileWasPicked.value);
```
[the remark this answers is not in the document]
MartinSchuhmacher: Eventhough I think Copilot is right on the finding, it still would work without it in the currently implemented logic. So up to you to change it.

src/modules/feature/board-file-element/FileContentElement.vue
```diff
@@ -162,19 +163,34 @@ watch(element.value, async () => {
 	isLoadingFileRecord.value = false;
 });
 
+const handleBeforeUnload = (event: BeforeUnloadEvent) => {
+	if (fileWasPicked.value) {
+		// Opens confirmation dialog in firefox
+		event.preventDefault();
+		// Opens confirmation dialog in chrome
+		event.returnValue = "";
```
MartinSchuhmacher: if we have different solutions for different browsers, should we return early if one of them worked? Otherwise I'm little confused why `event.preventDefault()` opens some dialog for firefox browsers.

src/modules/feature/board-file-element/FileContentElement.vue
```diff
@@ -162,19 +163,34 @@ watch(element.value, async () => {
...
+	}
+};
+
 onMounted(async () => {
+	window.addEventListener("beforeunload", handleBeforeUnload);
```
MartinSchuhmacher: according to the Copilot comment from below, why not puting the listener only when needed, so for `fileWasPicked` and we put a `watcher` to check and only add/remove the listener then

dyedwiper: Can you register the event handler in `onUploadFile` and also remove it there when the upload is finished?

src/modules/feature/board-file-element/FileContentElement.vue
dyedwiper: See: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event

"The last two mechanisms are legacy features; best practice is to trigger the dialog by invoking preventDefault() on the event object, while also setting returnValue to support legacy cases."

Though it says there to set `returnValue` to a non-empty string. Strange that we set it to an empty string here.

src/modules/feature/board-file-element/FileContentElement.vue
luejoh: Ye noticed the same, but was hesitating in changing code, which is unrelated with the bug. Nevertheless. Preferring the solution without registering the event everywhere.

luejoh: adjusted

src/modules/feature/board-file-element/FileContentElement.vue
luejoh: Changed value to "true" to return a truthy value.

src/modules/feature/board-file-element/FileContentElement.vue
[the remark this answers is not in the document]
luejoh: I dont think copilot is right, since typing is inferreed to be boolean only. So wont change sth. here

src/modules/feature/board-file-element/FileContentElement.vue
dyedwiper: I think this actually preserves the behavior from before because the listener was registered in the FileUpload component and this was only mounted while uploading.

src/modules/feature/board-file-element/upload/FileUpload.unit.ts
```diff
@@ -112,52 +98,6 @@ describe("FileUpload", () => {
 
 				expect(wrapper.emitted("upload:file")).toHaveLength(1);
 			});
-
-			it("should render v-progress-linear component", async () => {
```
dyedwiper, line 116: This test seems unrelated to the change. Is it unnecessary for other reasons?

Same for the one below.

src/modules/feature/board-file-element/FileContentElement.vue
```diff
@@ -162,6 +163,15 @@ watch(element.value, async () => {
 	isLoadingFileRecord.value = false;
 });
 
+const handleBeforeUnload = (event: BeforeUnloadEvent) => {
+	if (fileWasPicked.value) {
```
dyedwiper, line 167: The condition in FileUpload before was `if (fileWasPicked.value || props.isUploading)`. So to preserve the behavior it should be `if (isUploading)` now, shouldn't it?

src/modules/feature/board-file-element/FileContentElement.vue
luejoh: i thought the same, but it was acutally registering it multiple times. Tested it during runtime

src/modules/feature/board-file-element/FileContentElement.vue
luejoh: No actually not. There are the following situations:

- User did chose a file and its currently uploading --> fileWasPicked = true --> warning on leave
- User did chose a file and its currently uploading, but he opened the same page in another tab --> fileWasPicked = false, isUploading=true --> no warning on leave
- User did chose a file and its currently uploading, but the state is not transferred before upload finishs --> the same state with a local ref was used in FileUpload.vue before to keep track if the user did upload, without syncing the current state with the server

src/modules/feature/board-file-element/upload/FileUpload.unit.ts
luejoh: Well both of them were testing against a user interaction in the FileUpload.vue, which was mutating internal state and there switching state, which caused switching what is displayed.
Since i dont wanna test FileContentElement.vue, emitting is the only thing, what happens and v-if state on boolean values (prop isUploading) is not worth testing id say.

Agree?

src/modules/feature/board-file-element/FileContentElement.vue
dyedwiper: I don't really understand the third situatiion. What do you mean by "the state is not transferred"? From where to where?

And with your change you did change the behavior, didn't you? You just corrected the previous condition because it would have shown a warning in the second situation, right?

src/modules/feature/board-file-element/upload/FileUpload.unit.ts
dyedwiper: Hm, don't know. So you mean that testing if a component is rendered is generally unnecessary if the `v-if` just gets a boolean prop?

src/modules/feature/board-file-element/FileContentElement.vue
dyedwiper: Okay, weird...

src/modules/feature/board-file-element/upload/FileUpload.unit.ts
luejoh: yep. The easier the condition is, the less its worth it, testing if the dom tree is as expected.
For me boolean values, which are derived from very simple conditions are the best example of such.

src/modules/feature/board-file-element/FileContentElement.vue
luejoh: The third one was there previously too and I just kept it.
Once you await "upload()", the server has the "isUploading" state in place, but the uploading client does not. A separate tab, while its uploading receives it. The one uploading, only after the upload finished (so consequently, its not uploading anymore). So:
Client/Tab A (Uploading) --> not in sync with servers "isUploading" --> manual switch value with the help of a ref
Client/Tab B (not uploading) --> in sync with the servers "isUploading" --> manual switch value is false

> And with your change you did change the behavior, didn't you?

Actually not. It was the same previously, the ref just did live in FileUpload.vue and not FileContentElement.vue

dyedwiper: But the condition before was `if (fileWasPicked.value || props.isUploading)`, so it checked the state from the server, did it not?

luejoh: ah ye, true. Sorry i was sure, i didnt changed it, but this caused case 2 to fail. It would warn, even tho its not uploading anything

dyedwiper: Okay, yeah, it makes sense to me to change it

src/modules/feature/board-file-element/upload/FileUpload.unit.ts
dyedwiper: Okay, will think about that. Fine for me for now.

dyedwiper (approved):

PR #4333, opened 2026-08-05
BC-11811 - Handle gateway timeouts when duplicating large rooms/boards

sebmue-dataport:
# Short Description

When duplicating a room or board the call hapen sychronously, with the FE waiting for the result (and the ID of the new content) from the BE. If a room is rather big it will take some time to copy and the connection runs into a timeout. This led to the default erorr message, making the user believe nothing was copied - only to find the copy later on.

Instead of modifying the BE behaviour (which would be a big change) we differentiate between types of errors. In case of a gateway timeout we assume the copy process is still ongoing and show a notification accordingly. 

BC-11811

## Checklist before merging

### DIFF
--- a/src/composables/async-tasks.composable.ts
+++ b/src/composables/async-tasks.composable.ts
@@ -59,10 +59,14 @@ export const useSafeAxiosTask = () => {
 
 	const { loadingState, withLoadingState } = useDebouncedLoading();
 
-	const execute = async <T>(fn: AsyncFunction<T>, onErrorNotifyMessage?: string): Promise<TaskResult<T>> => {
+	const execute = async <T>(
+		fn: AsyncFunction<T>,
+		onErrorNotifyMessage?: string,
+		options?: { skipErrorNotification?: (error: Error) => boolean }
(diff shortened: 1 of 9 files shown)


src/composables/async-tasks.composable.ts
```diff
@@ -59,10 +59,14 @@ export const useSafeAxiosTask = () => {
...
+		onErrorNotifyMessage?: string,
+		options?: { skipErrorNotification?: (error: Error) => boolean }
+	): Promise<TaskResult<T>> => {
 		const { result, success, error } = await withLoadingState(() => safeExec<T>(fn));
 
-		if (error && onErrorNotifyMessage) {
+		if (error && onErrorNotifyMessage && !options?.skipErrorNotification?.(error)) {
 			const apiError = mapAxiosErrorToResponseError(error);
 
 			if (apiError.code) {
```
virgilchiriac, line 72: I think we should add unit test for this new case alone in async-tasks.composable.unit.ts

virgilchiriac (approved):

PR #4340, opened 2026-08-14
BC-3187- news page

virgilchiriac:
# Short Description

BC-3187
https://github.com/hpi-schul-cloud/schulcloud-server/pull/6432

## Checklist before merging


src/modules/data/data-access/news.composable.ts
```diff
@@ -20,6 +20,12 @@ export const useNewsActions = () => {
 			t("common.notifications.errors.notLoaded", { count: 2, type: t("common.words.news", 2) })
 		);
 
+	const fetchNewsPage = async ({ limit, skip, unpublished }: { limit: number; skip: number; unpublished: boolean }) =>
```
muratmerdoglu-dp: It's ok, but maybe you consider extend `fetchNewsList` method with some optional parameters, since both are almost the same.

src/modules/page/news/NewsOverview.page.vue
```diff
@@ -0,0 +1,116 @@
...
+import { useI18n } from "vue-i18n";
+
+const NEWS_PER_PAGE = 10;
+
+const { t } = useI18n();
+const appStore = useAppStore();
+const { fetchNewsPage, loadingState } = useNewsActions();
+const { fetchNewsPage: fetchNewsCountPage } = useNewsActions();
+
+const news = ref<NewsResponse[]>([]);
```
muratmerdoglu-dp: I think these states should be in the composable file. And simply get exported.

src/modules/page/news/NewsOverview.page.vue
```diff
@@ -0,0 +1,116 @@
...
+			label: t("pages.news.index.new"),
+			to: "/news/new",
+			dataTestId: "create-news-btn",
+		},
+	];
+});
+
+useTitle(buildPageTitle(t("pages.news.title")));
+
+const loadNews = async () => {
```
muratmerdoglu-dp: If the states get moved into the composables, then no need to assign the results in here. So that the page/component files will be simpler.

src/modules/feature/news/NewsCard.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+	<VCard
+		class="d-flex flex-column news-card"
+		:href="`/news/${newsItem.id}`"
```
muratmerdoglu-dp: I think we should use `:to` prop to avoid the full page reloading when clicking

src/modules/feature/news/NewsCard.vue
```diff
@@ -0,0 +1,70 @@
...
+		data-testid="news-card-item"
+		@dragstart.prevent
```
muratmerdoglu-dp: Is `@dragstart.prevent` needed?

src/modules/feature/news/NewsCard.vue
```diff
@@ -0,0 +1,70 @@
...
+	>
+		<VCardTitle class="news-card-title bg-primary-lighten text-wrap" :data-testid="dataTestId">
+			<div class="d-flex align-center">
+				<VIcon size="14" class="mr-1" :icon="mdiNewspaperVariantOutline" />
+				<span class="text-sm font-weight-regular">{{ fromNowUtc(newsItem.displayAt) }}</span>
```
muratmerdoglu-dp: I prefer `fromNowUtc` should be in the script section. Even the `displayAt` property should be mapped into `fromNowUtc` format in the composable while fetching data.

src/modules/feature/news/NewsCard.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+	<VCard
+		class="d-flex flex-column news-card"
+		:href="`/news/${newsItem.id}`"
+		data-testid="news-card-item"
```
muratmerdoglu-dp: This component is used in a loop and every component gets the same `data-testid`. 
I think `data-testid` should be dynamic and unique. Maybe assigning it in the parent component while using it in the `v-for` loop.

src/modules/feature/news/NewsGrid.vue
```diff
@@ -0,0 +1,28 @@
...
+</template>
+
+<script setup lang="ts">
+import type { NewsCardItem } from "./news-card.types";
+import NewsCard from "./NewsCard.vue";
+
+withDefaults(
+	defineProps<{
+		newsItems: readonly NewsCardItem[];
+		dataTestid?: string;
```
muratmerdoglu-dp: Dashboard and NewsOverview never pass the `data-testid, so it is unused. If someone did pass it, every card would get the same test id, and it would land on NewsCard’s title rather than the grid. The grid already has a hardcoded data-testid="news-section". 
I think we should refactor all testId props drilling.

src/modules/feature/render-html/RenderHTML.vue
muratmerdoglu-dp, line 1: I think it's time to refactor this file into `script setup` approach. :)

src/modules/page/news/NewsOverview.page.vue
```diff
@@ -0,0 +1,116 @@
...
+import { useTitle } from "@vueuse/core";
+import { computed, ref, watch } from "vue";
+import { useI18n } from "vue-i18n";
+
+const NEWS_PER_PAGE = 10;
+
+const { t } = useI18n();
+const appStore = useAppStore();
+const { fetchNewsPage, loadingState } = useNewsActions();
+const { fetchNewsPage: fetchNewsCountPage } = useNewsActions();
```
muratmerdoglu-dp: Do we need to use `useNewsActions` twice?
This results in two composable instances being created.

src/modules/page/news/NewsOverview.page.vue
```diff
@@ -0,0 +1,116 @@
...
+	currentPage.value = page;
+	await loadNews();
+};
+
+watch(activeTab, async () => {
+	currentPage.value = 1;
+	await loadNews();
+});
+loadNews();
+loadUnpublishedTotal();
```
muratmerdoglu-dp: It seems `unpublished` count is fetched once and never refreshed.

muratmerdoglu-dp (changes requested):

src/modules/page/news/NewsOverview.page.vue
virgilchiriac: a simple hack which prevents the posibility that the results overwrite each-other
has to do with the the execute method

src/modules/feature/news/NewsCard.vue
virgilchiriac: this is how was done on Dashboard page, here just moved it and reused it for NewsOverview
To clean it really, I guess more refactoring would be necesary. The data comes from parent component, where for Dashboard we have a function in news.composable while for NewsOverview is in the component itself....

src/modules/data/data-access/news.composable.ts
virgilchiriac: done

src/modules/feature/news/NewsCard.vue
virgilchiriac: done, also in the navigation and on Dashboard

src/modules/feature/news/NewsCard.vue
virgilchiriac: fixed

src/modules/feature/news/NewsCard.vue
virgilchiriac: I also don't see the point why was there. Removed

src/modules/feature/news/NewsGrid.vue
virgilchiriac: was fixed

src/modules/page/news/NewsOverview.page.vue
virgilchiriac: fixed

src/modules/feature/render-html/RenderHTML.vue
muratmerdoglu-dp: done

src/modules/data/data-access/news.composable.ts
```diff
@@ -124,3 +124,70 @@ export const useNews = (newsId: Ref<string | undefined>) => {
...
+
+	const onPageChange = async (page: number) => {
+		currentPage.value = page;
+		await loadNews();
+	};
+
+	watch(activeTab, async () => {
+		currentPage.value = 1;
+		await loadNews();
+	});
```
virgilchiriac, line 178: somehow to me seems a bit odd to see watch in composable, but I see is is common practice

muratmerdoglu-dp (approved):

PR #4342, opened 2026-08-26
BC-12144 - robust file deletion of big files

hoeppner-dataport:
# Short Description
A request-timeout for delete-operations resulted into incomplete delete-operations with an incorrect frontend-state.

- new config value for delete operations in the file-storage: INCOMING_REQUEST_TIMEOUT_DELETE_API_MS
- this new config value defaults to 600000 ms = 600s = 10 mins - like the similar config-value for copying
- request no longer runs into a timeout - not even for very large files
- by configuring the new value to:
-- 1ms => the error is easily reproducible
-- 5ms => the error is reproducible for large files
-- 600000ms => the error is not reproducible even for big files
- on the frontend-side a small delay and a different way of updating the frontend results in a proper rollback of the deleted file, if a timeout occurs

hint: configure config-maps of filestorage for testing and restart file-storage pod afterwards to apply the newly configured value

- https://ticketsystem.dbildungscloud.de/browse/BC-12144
- https://github.com/hpi-schul-cloud/file-storage/pull/289

## Checklist before merging


src/modules/data/file/FileStorageApi.composable.ts
```diff
@@ -191,14 +191,34 @@ export const useFileStorageApi = () => {
...
+		const uniqueParents = new Map<string, { parentId: string; parentType: FileRecordParent }>();
+
+		for (const fileRecord of fileRecords) {
+			const key = `${fileRecord.parentType}:${fileRecord.parentId}`;
+			if (!uniqueParents.has(key)) {
+				uniqueParents.set(key, { parentId: fileRecord.parentId, parentType: fileRecord.parentType });
+			}
+		}
+
+		await new Promise((resolve) => setTimeout(resolve, 500));
```
sebmue-dataport, line 204: Why is the waiting required here? I would assume the for loop is blocking and the .set() changes the value immediately not requiring any arbitrary waiting period afterwards?

src/modules/data/file/FileStorageApi.composable.ts
```diff
@@ -191,14 +191,34 @@ export const useFileStorageApi = () => {
...
+		await Promise.all(
+			Array.from(uniqueParents.values()).map(({ parentId, parentType }) => fetchFiles(parentId, parentType))
```
sebmue-dataport: Since fechFiles mutates data elsewhere I would prefer forEach instead of map. When I see map() used I assume it won't mutate any data elsewhere but just give a result such as `const multipliedValues = numbers.map(multiplyFn)` instead of `numbers.forEach(multiplyInDatabaseFn)`

src/modules/data/file/FileStorageApi.composable.ts
hoeppner-dataport: I assumed that, too. But without it the behaviour mentioned in the comments of the ticket was still happening:
"in case of 504, the UI will insert the file again in the list (the position where it was is not respected, as we had the list sorted by size, the file appear again at the bottom)"

src/modules/data/file/FileStorageApi.composable.ts
hoeppner-dataport: It needs to be .map, as an Array of promises is generated and that is awaited.
But I agree, that it is a little much for one line. Will fix it.

sebmue-dataport (approved):

